PR #32, opened 2022-11-17
Improve setting prices

0xcodercrane:


.husky/pre-commit
```diff
@@ -1,4 +1,2 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
-
-yarn build:serverless
```
pavlovcik, line 4: This should be restored because unlike a normal repository with normal source code, this source code must also include the compiled files for the GitHub Action to be useful/up-to-date with the source code. 

This enforces that the compiled code is synced to the latest source code.

.gitignore
```diff
@@ -5,4 +5,5 @@ npm-debug.log
 .env
 coverage
 lib
-.DS_Store
\ No newline at end of file
+.DS_Store
+dist
```
pavlovcik, line 9: The code must be compiled with `@vercel/ncc` and hosted on the repo for the GitHub Action to run it.

.github/workflows/main.yml
```diff
@@ -1,57 +1,25 @@
...
-      # Use the output from the `hello` step
-      # - name: Get the output time
-      # run: echo "The time was ${{ steps.hello.outputs.time }}"
+          node-version: "18"
+
+      - name: Install
+        run: yarn install
+
+      - name: Build
+        run: yarn build:local && yarn build:serverless
```
pavlovcik, line 25: This CI is getting very meta for me. Are you sure that this works or must we fall back to manually compiling and including the `dist` directory in the repository?

0xcodercrane: I am sure

0xcodercrane:
Anyway this is still a draft pr. I will request for review once I think its ready @pavlovcik

src/types/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+  PRICE_200_2000 = "Price: 200-2000 USDC",
+  PRICE_300_3000 = "Price: 300-3000 USDC",
+  PRICE_400_4000 = "Price: 400-4000 USDC",
+  PRICE_500_5000 = "Price: 500-5000 USDC",
+
+  PRICE_100_400 = "Price: 100-400 USDC",
+  PRICE_500_2000 = "Price: 500-2000 USDC",
+  PRICE_1000_4000 = "Price: 1000-4000 USDC",
+  PRICE_2000_8000 = "Price: 2000-8000 USDC",
+}
```
pavlovcik, line 26: We should allow our customers to configure these. Something to bear in mind soon! From what I understand, GitHub Apps (bots) are normally configured with a .yml file in customer repositories. Perhaps you know the best way to proceed here.

src/types/payload.ts
```diff
@@ -0,0 +1,212 @@
...
+  LOCKED = "locked",
+  OPENED = "opened",
+  PINNED = "pinned",
+  REOPENED = "reopened",
+  TRANSFERRED = "transferred",
+  UNASSIGNED = "unassigned",
+  UNLABELED = "unlabeled",
+  UNLOCKED = "unlocked",
+  UNPINNED = "unpinned",
+}
```
pavlovcik, line 31: Curious what the purpose of the enum is. Why not use the strings directly? Aren't they simplier for typechecking via something like: 

```typescript
export type Action = "requested" | "review_requested" | "review_request_removed" ...
```

pavlovcik (commented):
Looks like really good progress on the refactor. I think the code appears to be of higher quality with better type definitions which is huge! Great progress.

src/handlers/price/calculate.ts
```diff
@@ -0,0 +1,8 @@
+const DEFAULT_BASE_VALUE = 1000;
```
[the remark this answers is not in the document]
0xcodercrane: yeah. we will definitely do that later. I just moved previous constant value.

src/types/payload.ts
```diff
@@ -0,0 +1,212 @@
...
+  action: Type.Enum(Action),
+  issue: IssueSchema,
+  label: LabelSchema,
+  sender: UserSchema,
+  repository: RepositorySchema,
+  organization: OrganizationSchema,
+  installation: InstallationSchema,
+});
+
+export type Payload = Static<typeof PayloadSchema>;
```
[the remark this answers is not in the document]
0xcodercrane: Yeah it took most of my time actually.

src/types/payload.ts
0xcodercrane: we can define in both ways but if its type should be used in other schemas, `Enum` was sometimes better with sinclair Type.Enum like 
```Typescript
received:events_url: TURL,
type: Type.Enum(UserType)
```

src/types/constants.ts
0xcodercrane: Yeah that's a good catch and what I was thinking during the research. The blocker is that we don't have any standard names/interfaces to make it configurable. We still have some unclear things as the followings.

Their time labels would be different from ours, 
Their price labels would be different from ours, 
Their price calculation logic would be different from ours.
...

From above unclear things, I am just focusing on ubiquity-specific bot on this pull request. 
we will make it configurable step by step.

src/bindings/event.ts
```diff
@@ -0,0 +1,52 @@
...
+  for (const mainAction of action) {
+    await mainAction();
+  }
+
+  // Run post-handlers
+  log.info(`Running post handlers: ${post.map((fn) => fn.name)}, action: ${payload.action}`);
+  for (const postAction of post) {
+    await postAction();
+  }
+};
```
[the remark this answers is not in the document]
0xcodercrane: The reason why I thought about pre/post handlers is to process non event-specific actions for issues like unassign bounty hunters. 

PreActionHandler - it will be processed before main handler to do some. We can put a few of small functions into a prehandler list. In terms of completion time, it needs to be shorter one as much as possible. because we don't want for preActionhandlers to take much time. Most of time should definitely be spent by main handlers. we can just leave it empty if no need to use.  In terms of speed, it should be fast.
 
 
ActionHandler - it is the main callback function to do event-based things.
 
PostActionHandler - it would be processed after the main handlers.  We can put the functions into a posthandler list which will take a bit longer. If no needed, we can leave it empty. In terms of speed, it can be slow.

.husky/pre-commit
0xcodercrane: fixed in latest commits

.gitignore
0xcodercrane: fixed in latest commits.

src/bindings/event.ts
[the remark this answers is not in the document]
pavlovcik: > In terms of completion time, it needs to be shorter one as much as possible.

Isn't this negligible? I mean, like less than one second to do these actions? From what I understand (I could be wrong, my understanding of GitHub Actions Runners isn't master level) once it hits the JavaScript runtime, it runs at JavaScript speeds. Its just the Runner having to start its vps, load its OS, download the source code etc which takes most of the time.

src/types/constants.ts
pavlovcik: For whatever its worth, I somewhat expect early partners to just copy our lead in order to onboard more easily. We spent a fair amount of time prototyping and using these values and they seem to be mostly good. I just think that the base rate is a simple but major variable. Perhaps just making base rate configurable and providing a few examples of how the entire pricing chart will look if modifying the base rate will be sufficient for getting our early partners to start with their own price ranges that make sense to them?

src/types/payload.ts
pavlovcik: I don't have a strong opinion either way. I ask out of curiosity.

src/types/constants.ts
0xcodercrane: Hmm, I can make it configurable as of now.

src/bindings/event.ts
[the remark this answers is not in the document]
0xcodercrane: I think i didn't give you detailed meaning here. I just tried to emphasize the delay starting from gh event until main handler get started. Your understanding about Github Actions Runner is correct.

src/types/constants.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/32/commits/d244ae24e2e800519fa972b538acae1bed610a60

.gitignore
```diff
@@ -6,4 +6,4 @@ npm-debug.log
 coverage
 lib
 .DS_Store
-dist
+
```
pavlovcik, line 9: You added a ton of Probot boilerplate images and code by removing this. Are you sure?

pavlovcik, line 9:
```suggestion
dist
```

pavlovcik (changes requested):

.gitignore
0xcodercrane: oh my bad. `dist` should be there for instant running. Did i forgot to revert it???

pavlovcik (approved):

pavlovcik:
Be sure to delete the `dist` folder before merging @0xcodercrane

pavlovcik (commented):
Be sure to handle the `dist` comment I left

0xcodercrane:
> Be sure to handle the `dist` comment I left

What we want is just a runnable file `index.js`. so I'd like to move `index.js` file to bin directory add `dist` to gitignore.

pavlovcik:
> > Be sure to handle the `dist` comment I left
> 
> What we want is just a runnable file `index.js`. so I'd like to move `index.js` file to bin directory add `dist` to gitignore.

Sounds good to me!

pavlovcik:
Broke the Action

https://github.com/ubiquity/bounty-bot/actions/runs/3554610485

PR #34, opened 2022-11-20
Adds payout module for safe interaction

3commascapital:
* not hooked up yet, only module
* mnemonic used for now - can easily switch to privkey


pavlovcik:
How is accounting handled? Also we should do private key instead of mnemonic.

.env.example
```diff
@@ -7,3 +7,7 @@ LOG_LEVEL=debug
 
 # Go to https://smee.io/new set this to the URL that you are redirected to.
 WEBHOOK_PROXY_URL=
+
```
zgorizzo69, line 10: I think we miss the `MNEMONIC` env

src/utils/payout.ts
```diff
@@ -0,0 +1,74 @@
...
+}
+
+export const providerKeyByChainId = new Map<number, string>([
+  [1, process.env.RPC1_ALCHEMY_KEY as string],
+  [5, process.env.RPC5_ALCHEMY_KEY as string], // goerli - safe compatable
+])
+
+export const payout = async (args: PayoutParams) => {
+  // collect baseline info
+  const tokenAddress = ethers.utils.getAddress(args.token)
```
zgorizzo69, line 25: shouldn't we make sure the token address is whitelisted ?

src/utils/payout.ts
```diff
@@ -0,0 +1,74 @@
...
+  const isEth = tokenAddress === ethers.constants.AddressZero
+  const destinationAddress = ethers.utils.getAddress(args.to)
+  const mnemonic = process.env.MNEMONIC as string
+  const network = args.network || 1
```
zgorizzo69, line 29: we should also check that we allow the network and that we have the corresponding alchemy key

zgorizzo69 (changes requested):

src/utils/payout.ts
pavlovcik: That's a nice detail. This is to make sure we don't waste gas I presume?

src/utils/payout.ts
pavlovcik: Also nice detail.

.env.example
pavlovcik: We will use PRIVATE_KEY or something of the like, instead of mnemonic.

0xcodercrane:
@3commascapital  would you rebase from main and continue working? we had a bunch of refactors

src/utils/payout.ts
3commascapital: correct. good addition @zgorizzo69

3commascapital: do you mean something specific @zgorizzo69 when you use whitelisted? i was just thinking - balance, and allowed is set to cover the amount. is there anything else?

pavlovcik:
Screenshot 2022-12-02 at 15 37 42

PR #39, opened 2022-11-30
More architecture validations

0xcodercrane:


src/helpers/label.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const createLabel = async (name: string): Promise<void> => {
+  const context = getBotContext();
+  const payload = context.payload as Payload;
+
+  await context.octokit.rest.issues.createLabel({
+    owner: payload.repository.owner.login,
+    repo: payload.repository.name,
+    name,
+    color: COLORS.price,
```
pavlovcik, line 29: Bonus if you able to make a [gradient with the prices](https://github.com/ubiquity/ubiquity-dollar/labels).

pavlovcik:
Unfortunately I'm so unfamiliar with the codebase after your changes that I don't think I can give a very useful review.

src/helpers/label.ts
0xcodercrane: Yes, I can. but I want clear design/spec for color scheme. Right now, we have 2 types of price labels; single price like 100 usdc, 200 usdc, and range price like 100 - 1000 usdc, 200 - 2000 usdc

pavlovcik: You need to calculate the brightness of the colors e.g. `#000` = 0% brightness; `#0f0` = 100% brightness. There are libraries for this. Then you can assert that `0 DAI` = 0% brightness; and `10 000 DAI = 100% brightness.

So this should mean that every DAI = +1% brightness.

e.g.

0 DAI = `#000000`
2 500 DAI = `#004000`
5 000 DAI = `#008000`
7 500 DAI = `#00bf00`
10 000 DAI = `#00ff00`

pavlovcik: Turns out you can directly do math on these in JavaScript pretty easily. 

```typescript

const generateHexValue = (brightness: number) => Math.ceil(0xff * brightness).toString(16)
generateHexValue(0.5) // "80"


```

PR #51, opened 2022-12-17
fix(config): made some changes I thought made sense and ran prettier

pavlovcik:
The most important changes I think are in:

- `.github/workflows/bot.yml`
- `.github/workflows/marketplace.yml`

Because it looks like the secrets' property names are not correct.

The rest was prettier, mostly.


pavlovcik:
I just realized that it doesn't make sense that we must pass in the App ID of the action that is invoked by our bot, does it? Shouldn't GitHub Apps be able to read their own App IDs?

0xcodercrane (approved):
Looks good

PR #63, opened 2022-12-23
Unassigning bounties

0xcodercrane:


.github/workflows/scheduler.yml
```diff
@@ -1,7 +1,7 @@
 name: Run a bot at a regular interval
 on:
   schedule:
-    - cron: '*/1 * * * *'
+    - cron: '* */1 * * *'
```
pavlovcik, line 4: The decision was events based, and not cron so this doesn't look right.

src/configs/price.ts
```diff
@@ -6,21 +6,25 @@ export const DefaultPriceConfig: PriceConfig = {
...
     {
       name: "Time: <2 Weeks",
       weight: 3,
+      value: 3600 * 24 * 14,
       target: "Price: 300-3000 USD",
     },
     {
       name: "Time: <1 Month",
       weight: 4,
+      value: 3600 * 24 * 30,
```
pavlovcik, line 27: Wouldn't it be more accurate to divide `365.25 / 12` to derive the monthly average?

src/handlers/shared.ts
```diff
@@ -1,6 +1,7 @@
 import { getBotContext } from "../bindings";
 import { ActionHandler } from "../types";
 
+export const deadLinePrefix = "The time limit for this bounty is on";
```
pavlovcik, line 4: Where does this string manifest for the user?

src/types/payload.ts
```diff
@@ -27,6 +27,8 @@ export enum Action {
   UNLABELED = "unlabeled",
   UNLOCKED = "unlocked",
   UNPINNED = "unpinned",
+
+  SCHEDULE = "schedule",
```
pavlovcik, line 31: 🤮

.github/workflows/scheduler.yml
0xcodercrane: Yes it was. but we don't have a certain event which could be triggered everyday, every hour. but please let me know the event name which would be triggered regularly. If then, this would be just replacement

pavlovcik: On any our "scheduled" events should piggy back off of

src/configs/price.ts
0xcodercrane: lol If we have to have really strong requirements in exact timeframe, 365.25 / 12 would also be incorrect. we might need to take every 4 year into consideration as well.

.github/workflows/scheduler.yml
0xcodercrane: Which events are scheduled ones?

src/handlers/shared.ts
0xcodercrane: Once a hunter gets assigned by the bounty-admin, the message will be posted on the issue.

.github/workflows/scheduler.yml
pavlovcik: Any logic you plan to run from cron should run as a post hook handler, for example.

src/configs/price.ts
pavlovcik: The reason why I bring this up is because its applicable for (and perhaps more pronounced for accuracy) with "<2 Weeks" as well. I think it applies across the board, and could be particularly useful when we have custom time estimates (crowdsourced from the DevPool e.g. "3 Days")

.github/workflows/scheduler.yml
0xcodercrane: The problem is that we don't know the events triggered at a regular interval. For example, we want to post weekly updates on discord/telegram. This action needs to be happened once a week. It would be best it can happen on Monday but there could be no events until Tuesday. Then weekly updates would be delayed until Tuesday. 

The bounty bot can run in both ways in general. One is the passive way and other one is active way. 
Both have their pros/cons actually but it seems you're just suggesting a passive way which means we could do nothing for which SHOULD happen at a certain time if no events emitted. 

Please clarify me again if you can. If we MUST do it only in passive way, I might need to add `research` label to this issue and spend time to find the proper events (I am not 100% sure I could find them in a short time)

pavlovcik: > For example, we want to post weekly updates on discord/telegram.

This is a good point and I'd like to think about this more from the perspective of scaling this to other (e.g. infinite) partner projects. The final vision will be to have some sort of matchmaking system (imagine if we have thousands of new bounties/tasks posted every week across all partner projects) in this vision, posting the new bounties every week would not make sense. 

Perhaps we can make a separate script run for this particular situation with the intent that we will deprecate it as soon as we have like 10+ partners (imagine posting a list of 50+ bounties every Monday in the chat- that would get kind of ridiculous)

src/configs/price.ts
pavlovcik: Basically if we calculate time with precision we should be able to pick any time interval (1 day, 3 days, 5.5 days etc) and it should be accurate. Right now we only have like four time options, but in the future if we do figure out how to incentivize bounty hunters to gamble on the time estimate of the bounties, we could get accurate time estimates via averages/medians for bounties which will be unlikely to be one of the preset values you have here.

The only way to solve this and to make it work with any time interval is to calculate it with precision- perhaps unix time style.

.github/workflows/scheduler.yml
0xcodercrane: `We have a handful of events that occur on the Ubiquity Dollar repo every day actually. I try to go through the notifications at least once a day and there seems to have consistently been daily activity for weeks now, minus New Years Eve and Christmas Eve I believe.`

This is just rough estimation. I guess we couldn't rely on the irregular events for the regular activities at all.

0xcodercrane: `Perhaps we can make a separate script run for this particular situation with the intent that we will deprecate it as soon as we have like 10+ partners (imagine posting a list of 50+ bounties every Monday in the chat- that would get kind of ridiculous)`

If you want a separate script for regular activities, we can definitely not use cron,  Just make me clear in development stage. Literally separate script is out of scope of this project then.

src/configs/price.ts
0xcodercrane: Oh I understand what you're saying. I agree to make it working with any time inverval. but we will definitely do it on other issues later.

.github/workflows/scheduler.yml
0xcodercrane: @pavlovcik . would you please give me your proof why cron isn't suitable for such regular action based on your experience?

pavlovcik: I Google searched events based vs timetabling and there's a huge amount of articles that point to events based being superior (for most cases.) I could link you random articles from the first page of the Google results but it would be pretty easy for you to find yourself because there are so many.

In addition I asked @transitive-bullshit his thonks on this matter.

image

In conclusion, cron is nice for quick hack jobs. Events-based is how robust, production-ready systems are generally built.

transitive-bullshit: no idea if cron is right for this specific use case; just that in general, I'd expect bots to respond instantly to user-defined bot events, and GH bots like vercel and https://socket.dev focus on responding via webhooks. again, no idea if that's relevant to the functionality in this specific PR. just a general precedent.

0xcodercrane: Agree! we did a bunch of discussions already. thanks for confirmation @pavlovcik  @transitive-bullshit

0xcodercrane: this would be updated as discussed on feat/46-supabase-adapter.

0xcodercrane:
https://github.com/ubiquity/bounty-bot/pull/75

Debu976116 (commented):
Bounties

PR #75, opened 2023-01-11
Feat/46 supabase adapter

0xcodercrane:


supabase/migrations/20230111145014_initial_schema.sql
```diff
@@ -6,6 +6,7 @@ CREATE TYPE issue_status AS ENUM (
 );
 
 CREATE TABLE users (
+    id SERIAL PRIMARY KEY,
```
pavlovcik, line 9: I wanted to use their GitHub username as the primary key. I'm no database expert but when I was working through this same problem, this greatly simplified things. 

To clarify, our entire system is dependent on GitHub with no plans to move to another platform. With this in mind, you will probably save a lot of pain and headache down the road using the GitHub ID as the primary key representing users in our system.

pavlovcik: One unrelated comment I also want to add concerning the User schema, in my scraper I prefixed properties with _ because the data from the UI could have collisions with the user data from their API. But there are also situations where the data is the same but represented differently. 

Off hand I remember follower and following counts from the UI are suffixed in K notation (i.e. `12.9k followers` in the UI vs `12965 followers` from the API.)

The benefit of storing the data from the UI instead of going to the API always is rate limiting. There is none for UI scrapes so it was faster and easier for me to scrape thousands of contacts. I thought it would be easiest to "enrich" the UI scraped data with....your favorite...a CRON job that can ping the API slowly over time and enrich everything.

pavlovcik: @0xcodercrane

src/handlers/processors.ts
```diff
@@ -15,14 +15,14 @@ export const processors: Record<string, Handler> = {
     action: [pricingLabelLogic],
     post: [nullHandler],
   },
-  [Action.SCHEDULE]: {
-    pre: [nullHandler],
-    action: [checkBountiesToUnassign],
-    post: [nullHandler],
-  },
```
pavlovcik, line 22: 😮‍💨

supabase/migrations/20230111145014_initial_schema.sql
0xcodercrane: > I wanted to use their GitHub username as the primary key. I'm no database expert but when I was working through this same problem, this greatly simplified things.
> 
> To clarify, our entire system is dependent on GitHub with no plans to move to another platform. With this in mind, you will probably save a lot of pain and headache down the road using the GitHub ID as the primary key representing users in our system.

Agree with your experience.  In general, I've been using SERIAL type to make indexing faster. But I guess we will not have much difference in terms of search speed.  so I will make username primary key

0xcodercrane: > One unrelated comment I also want to add concerning the User schema, in my scraper I prefixed properties with _ because the data from the UI could have collisions with the user data from their API. But there are also situations where the data is the same but represented differently.
> 
> Off hand I remember follower and following counts from the UI are suffixed in K notation (i.e. `12.9k followers` in the UI vs `12965 followers` from the API.)
> 
> The benefit of storing the data from the UI instead of going to the API always is rate limiting. There is none for UI scrapes so it was faster and easier for me to scrape thousands of contacts. I thought it would be easiest to "enrich" the UI scraped data with....your favorite...a CRON job that can ping the API slowly over time and enrich the contact data.

I guess we can prefix properties with _ but both represents the same number so in case UI scrapper, if we add a simple conversion method, we can get 12900 easily. Also I guess we will not have rate limiting for bounty bot because its works based on event which will have most fields we need. 

In worst case, if we couldn't use API, we can update table by creating a new migration.

pavlovcik: > we can get 12900 easily

Yes I already prototyped this in a Google Sheet when running some analytics on the data but I want precision so this strategy is not acceptable.

The GitHub API limits, if I recall correctly, is 60 times per minute; I agree that rate limiting is unlikely to be an issue.

src/handlers/assign/action.ts
```diff
@@ -60,5 +60,5 @@ export const commentWithAssignMessage = async (): Promise<void> => {
   const commit_msg = `${flattened_assignees} ${deadLinePrefix} ${endDate.toLocaleDateString("en-us")}`;
   log.debug(`Creating an issue comment`, { commit_msg });
 
-  await addCommentToIssue(commit_msg);
+  await addCommentToIssue(commit_msg, payload.issue!.number);
```
pavlovcik, line 63: Fancy syntax. I had to Google it.

src/handlers/wildcard/unassign.ts
```diff
@@ -41,8 +39,10 @@ const checkBountyToUnassign = async (issue: any): Promise<boolean> => {
   const curTimestamp = new Date().getTime();
 
   if (curTimestamp > deadLineOfIssue) {
-    log.debug(`Releasing the bounty back to dev pool because the allocated duration already ended`, { deadLineOfIssue, curTimestamp });
-    await addCommentToIssue(`Releasing the bounty back to dev pool because the allocated duration already ended`);
+    log.debug(
+      `Releasing the bounty back to dev pool because the allocated duration already ended! deadLineOfIssue: ${deadLineOfIssue}, curTimestamp: ${curTimestamp}`
+    );
+    await addCommentToIssue(`Releasing the bounty back to dev pool because the allocated duration already ended`, issue.number);
```
pavlovcik, line 45: It could be nice to have a global strings config for projects to be able to update these messages. It's not a priority and we would certainly be fine to operate without this feature but I'll make an issue for it anyways.

pavlovcik (approved):

src/handlers/wildcard/unassign.ts
0xcodercrane: Totally agree!

PR #78, opened 2023-01-16
fix: priority names

pavlovcik:
"Low" implies that its lower than "Normal" priority which is misleading. 

I normally would have pushed this directly to `development` but I want to make sure that it won't cause any issues with our current labeling. 

I know my other script renamed the labels, but in the process of renaming, it would actually delete (and lose the label data) from the issues, meaning I had to go back and manually re-assign all the labels again. I just want to make sure something like that doesn't happen again. 

If all is clear, please merge and deploy this.


0xcodercrane:
I wanna deploy with all the features in review

PR #79, opened 2023-01-16
Feat/lint staged

pavlovcik:
I wanted to quickly add `lint-staged` as I realized that the prettier doesn't automatically run (and then fails in the CI.) 

Mind as well import our `ubiquity-dollar` lint-staged settings!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage
 lib
 .DS_Store
 dist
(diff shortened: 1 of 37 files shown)


pavlovcik:
I just realized that the `bin/index.js` file will frequently have merge conflicts. Perhaps there is a better way to handle this?

PR #84, opened 2023-01-23
Feat/build on release

pavlovcik:
This should make development faster because now `dist/index.js` is compiled by a `on push` GitHub Action and committed to the repository on the current working branch.

At the moment I'm figuring out how to make it exit on success even if none of the TypeScript files were modified. Right now its crashing because I'm editing the `.yml` files. Those do not cause `dist/index.js` to change in contents so `git -am` does not change anything there.


pavlovcik:
Merge conflicts may occur on pull requests. If this happens, we can probably make a new action that forces a recompile and commit via `rm bin/index.js && yarn build:serverless` etc.

.github/workflows/push.yml
```diff
@@ -0,0 +1,42 @@
...
+        uses: actions/setup-node@v3
+        with:
+          node-version: "18"
+
+      - name: Install
+        run: yarn install
+
+      - name: Prettier
+        run: yarn prettier:check
+
```
0xcodercrane, line 25:
```suggestion
- name: Local Build
  run: yarn build:local
```

0xcodercrane (commented):
We should definitely remove scheduler.yml right. by the way, it has been already removed on development branch lol.

PR #89, opened 2023-01-25
fix: make organization payload optional

rndquu:
I've added the bot to one of my github repos and on adding a label to an issue I got the following error:
```
INFO (event): Payload schema validation failed!!!
    id: "b54e95b0-9cc5-11ed-8c7c-723ed247b6ad"
WARN (event):
    0: {
      "instancePath": "",
      "schemaPath": "#/required",
      "keyword": "required",
      "params": {
        "missingProperty": "organization"
      },
      "message": "must have required property 'organization'"
    }
    id: "b54e95b0-9cc5-11ed-8c7c-723ed247b6ad"
```

So this PR makes the `organization` property optional

### DIFF
--- a/src/types/payload.ts
+++ b/src/types/payload.ts
@@ -223,7 +223,7 @@ export const PayloadSchema = Type.Object({
   label: Type.Optional(LabelSchema),
   sender: UserSchema,
   repository: RepositorySchema,
-  organization: OrganizationSchema,
+  organization: Type.Optional(OrganizationSchema),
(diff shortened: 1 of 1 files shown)


pavlovcik (commented):
Seems fine that it is used on personal repositories but I imagine that it would only really be used on organization repositories.

0xcodercrane:
Ok. It's fine to make it optional because `organization`  isn't getting used right now.

0xcodercrane (approved):

PR #102, opened 2023-02-05
Update/pricing

pavlovcik:


src/configs/price.ts
```diff
@@ -3,61 +3,62 @@ import { PriceConfig } from "../types";
...
+      value: 3600,
+      target: "Price: 100+ USD",
+    },
     {
       name: "Time: <1 Day",
       weight: 1,
       value: 3600 * 24,
-      target: "Price: 100-1000 USD",
+      target: "Price: 100+ USD",
     },
```
0xcodercrane, line 17: seems duplicated in target price `Price: 100+ USD`, why?

0xcodercrane (changes requested):

src/configs/price.ts
pavlovcik: My mistake I guess the other one should be $12.5+ lol

0xcodercrane: by the way, do we need to make a 12.5 USD bounty?

PR #105, opened 2023-02-07
Improve unassigning bounties

0xcodercrane:


src/configs/strings.ts
```diff
@@ -1,3 +1,4 @@
 export const GLOBAL_STRINGS = {
   unassign_comment: "Releasing the bounty back to dev pool because the allocated duration already ended!",
+  ask_update: "Do you have any updates?",
```
pavlovcik, line 3: We should be using camelCase. Only underscores for CONSTANT_VARIABLES

supabase/migrations/20230207013753_update_types.sql
```diff
@@ -0,0 +1,26 @@
...
+ALTER COLUMN completed_at TYPE timestamp without time zone using completed_at::timestamp;
+
+ALTER TABLE public.issues
+ALTER COLUMN closed_at TYPE timestamp without time zone using closed_at::timestamp;
+
+ALTER TABLE public.wallets
+ALTER COLUMN created_at TYPE timestamp without time zone using created_at::timestamp;
+
+ALTER TABLE public.wallets
+ALTER COLUMN updated_at TYPE timestamp without time zone using updated_at::timestamp;
```
pavlovcik, line 26: Can you explain to me these migration files? Are they just for reference? Is this considered the correct way to include the database setup in the codebase? It seems like a good idea but this is just new for me to see.

README.md
```diff
@@ -30,9 +30,10 @@ jobs:
       - name: Ubiquity Bounty Bot
         uses: ubiquity/bounty-bot@c1c6c99336f34ac5e94efaed49c9f218fb7a2d76
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SUPABASE_PROJECT_URL: ${{ secrets.SUPABASE_PROJECT_URL }}
           SUPABASE_PROJECT_KEY: ${{ secrets.SUPABASE_PROJECT_KEY }}
+          FOLLOWUP_TIME: '4 days'
+          DISQUALIFY_TIME: '7 days'
```
pavlovcik, line 36: Shouldn't these be numbers? 
```suggestion
          FOLLOWUP_DAYS: 4
          DISQUALIFY_DAYS: 7
```

supabase/migrations/20230207013753_update_types.sql
0xcodercrane: > Can you explain to me these migration files? Are they just for reference? Is this considered the correct way to include the database setup in the codebase? It seems like a good idea but this is just new for me to see.

It is a normal process to have a migration file for every change in database management.

README.md
0xcodercrane: > Shouldn't these be numbers?

We don't need to care about its unit?

pavlovcik: Ideally, sure. But realistically, I can't imagine that users would mandate intraday follow ups. You think they will want to have a bot hound somebody twice a day and then eliminate within the next day? lol

src/configs/strings.ts
pavlovcik: This doesn't look like you changed it so you shouldn't mark it as resolved

0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/105/commits/eb8abcaa7a839addd4e7a1c2872149745ccb809f

can I make it resolved? lol

README.md
```diff
@@ -63,6 +64,11 @@ jobs:
 SUPABASE_PROJECT_URL=XXX
 SUPABASE_PROJECT_KEY=XXX
 ```
+5. Add `FOLLOW_UP_TIME` and `DISQUALIFY_TIME` to the `.env` file if you don't want to use default ones. Please refer [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) how to set the value
+```
+FOLLOW_UP_TIME="P4D" // 4 days
+DISQUALIFY_TIME="P7D" // 7 days
```
pavlovcik, line 70: It seems very unusual and not as user friendly as doing my suggestion

```
          FOLLOWUP_DAYS: 4
          DISQUALIFY_DAYS: 7
```

pavlovcik (changes requested):
Small changes requested but overall seems good.

One more concern I have is that I don't believe I see logic related to looping through all of the issues/PRs in order to sync time dependent operations. 

For example, I recall that when I saw the first demo of this, that the only time the bot would notice there was a late issue is when activity occurred on the issue itself. It should be checking every issue (I guess in reverse chronological order, meaning the highest issue number first) for any time operations like these follow ups, using the post-event-wildcard-handler.

README.md
pavlovcik: I think I understand your concern about giving the user more flexibility but there is also usability to consider. Also as I mentioned earlier, I really don't see a situation where a project will request intraday follow ups. That would be micromanaging and they would probably be sending messages over Slack or whatever instead for this

pavlovcik: New idea, we could have two secret .env properties `FOLLOWUP_ISO` and `DISQUALIFY_ISO` if these are found, then they overwrite `FOLLOWUP_DAYS`, and `DISQUALIFY_DAYS`?

We can highlight the user friendly one to new partners, and if they ask about smaller intervals we can show them these properties? 

I was inspired to suggest this for testing purposes. Would be easier to see unassign working if it does after a few minutes.

0xcodercrane: agree to have both `_ISO` and `_DAYS`

0xcodercrane: based on internal conversation, we decided to go with the `ms` npm module.

src/bindings/config.ts
```diff
@@ -1,14 +1,28 @@
...
+
+  const durationInSecs =
+    365 * 24 * 3600 * (durationObj.years ?? 0) +
+    30 * 24 * 3600 * (durationObj.months ?? 0) +
+    24 * 3600 * (durationObj.days ?? 0) +
+    3600 * (durationObj.hours ?? 0) +
+    60 * (durationObj.minutes ?? 0) +
+    (durationObj.seconds ?? 0);
+  return durationInSecs * 1000;
+};
```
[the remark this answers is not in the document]
0xcodercrane: If we use `ms`, we don't even need hardcoded calculation

README.md
pavlovcik: In the DMs we concluded that we will use a plain english library parser `@vercel/ms` to enhance user friendliness.

src/configs/shared.ts
```diff
@@ -5,31 +5,18 @@ export const COLORS = {
...
+ * ms('10h')     // 36000000
+ * ms('2.5 hrs') // 9000000
+ * ms('2h')      // 7200000
+ * ms('1m')      // 60000
+ * ms('5s')      // 5000
+ * ms('1y')      // 31557600000
+ * ms('100')     // 100
+ * ms('-3 days') // -259200000
+ * ms('-1h')     // -3600000
+ * ms('-200')    // -200
```
pavlovcik, line 19: Wouldn't negative break things? We should probably throw an error for negative values.

0xcodercrane: good catch

pavlovcik (approved):

src/configs/shared.ts
pavlovcik: I set approval assuming that you will finish this.

0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/105/commits/1081d8f6ae7c55fee84e336b2ccbbcca54837061

PR #108, opened 2023-02-11
feat: weekly summary

### LINKED ISSUE #45: Weekly reviews every week in the DevPool
### Discussed in https://github.com/ubiquity/bounty-bot/discussions/69

example:
> We closed seven pull requests last week! 🚀
This week, we added more tests for the ERC20Ubiquity contract, fixed a bug that was causing incorrect balances to be displayed in the footer, added spell checking for dapp and contract scripts, added a custom 404 page, implemented the ERC-2535 Diamond standard, fixed a bunch of Slither issues, adapted the contract to simplify UBQ farming, and fixed a bunch of warnings.

### How to solve
- [ ] database scheme and adapter needs to be completed prior to any analytics features
- [ ] Telegram bot adapter should be completed prior to this issue.
- [ ] Total closed issues, merged pull requests, the number of new bounties released, total payouts, etc. These information needs to be collected once a week and create a proper description for weekly updates with collected information

Steveantor:
Resolves #45


src/job/summary.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export const summaryJob = new CronJob(
+  "1 2 * * sun",
+  () => {
+    init();
+  },
+  null,
+  false,
+  "Atlantic/Reykjavik" //GMT-UTC matching timezone
+);
```
pavlovcik, line 12: I have a couple of issues with this implementation. Unfortunately you were not apart of my conversations with @0xcodercrane and some others regarding CRON but basically the conclusion is that we will not be using or relying on CRON. Instead, for timed events like these we have a "wildcard post event handler" which runs after every time this bot is invoked. Remember, the bot is only invoked as a response to an "event" e.g. a new comment, a new label, a new pull request etc. 

The idea is to always check after every event if it should post a summary. If not, presumably it should add to the summary queue. In this same logic it should also check if it posted the summary of the weeks that have passed (we can max it to just the last week, so that it won't post infinite summaries since the beginning of time.)

Also I'm skeptical that this implementation will work at all because the bot must be actively performing a task at exactly the same time as `""1 2 * * sun"` to catch this. The Action will only run for the duration of the function which will likely be less than a minute at a time.

src/adapters/summarize/summarize.ts
```diff
@@ -0,0 +1,191 @@
...
+  });
+  summaryInfo =
+    `<code>new issues: ${openedIssues}</code>\n` +
+    `<code>issues resolved: ${closedIssues}</code>\n` +
+    `<code>total user interactions count: ${comments}</code>\n` +
+    `<code>bounties given: ${bountiesUSD} USD</code>\n` +
+    `<code>new pulls: ${openedPRs}</code>\n` +
+    `<code>closed pulls: ${closedPRs}</code>\n` +
+    `<code>merged pulls: ${mergedPRs}</code>\n` +
+    `<code>total commits: ${commits}</code>\n`;
```
pavlovcik, line 133: This seems like a really solid summary implementation. You even are using whatthediff! Is this rendering an image and posting an image summary then?

pavlovcik (changes requested):
It looks like you did a solid job on the core logic but I have an issue with relying on CRON. Fortunately @0xcodercrane already laid down the foundation for you to change what invokes this logic (post-event wildcard handler.)

src/adapters/summarize/summarize.ts
Steveantor: yes, it posts like this
image
@pavlovcik

src/job/summary.ts
Steveantor: So I have replaced the CronJob with wildcard selector @pavlovcik

src/bindings/config.ts
```diff
@@ -25,6 +25,10 @@ export const loadConfig = async (): Promise<BotConfig> => {
       token: process.env.TELEGRAM_BOT_TOKEN ?? "",
       delay: process.env.TELEGRAM_BOT_DELAY ? Number(process.env.TELEGRAM_BOT_DELAY) : DEFAULT_BOT_DELAY,
     },
+    git: {
+      org: process.env.ORG_NAME ?? "ubiquity",
+      repo: process.env.REPO_NAME ?? "ubiquity-dollar",
```
pavlovcik, line 30: We should log a warning if it goes to the fallback, our repo. It can cause confusion for partners setting this up if they forgot to set it to their repo.

pavlovcik (changes requested):
I would like to know how to test this to be 100% confident of my review but overall it looks pretty good.

- I have a small change request as a comment.
- I also presume that this is now implemented without CRON, and instead with the post-event wildcard handler. 
- I also thought i paused whatthediff's access to the repo so I'm surprised that it seems to keep working fine. 
- Lastly it is not clear to me when is the time that it is able to run. Is it after Monday 0:00 UTC?

src/bindings/config.ts
Steveantor: added fallback warning

Steveantor:
> I would like to know how to test this to be 100% confident of my review but overall it looks pretty good.
> 
> * I have a small change request as a comment.
> * I also presume that this is now implemented without CRON, and instead with the post-event wildcard handler.
> * I also thought i paused whatthediff's access to the repo so I'm surprised that it seems to keep working fine.
> * Lastly it is not clear to me when is the time that it is able to run. Is it after Monday 0:00 UTC?

- I'm using whatthediff's public repo analyzer, which should work regardless.
- Use `example.ts` to test.
- It's 00:00Z same as UTC (GitHub uses Z), any day.

pavlovcik:
> I'm using whatthediff's public repo analyzer, which should work regardless.

How did you find this? I spent some time researching this and it never came across my radar!

> Use example.ts to test.

I guess we should setup a tests folder and suffix with `.t.ts` etc.

> It's 00:00Z same as UTC (GitHub uses Z), any day.

Its not clear to me then. Does it run daily? Does it have a check to run max once per week? Is it in the wildcard post event handler?

Steveantor:
> > I'm using whatthediff's public repo analyzer, which should work regardless.
> 
> How did you find this? I spent some time researching this and it never came across my radar!

it's the public analyzer at[wtd@repository-report]( https://whatthediff.ai/repository-report)
> > It's 00:00Z same as UTC (GitHub uses Z), any day.
> 
> Its not clear to me then. Does it run daily? Does it have a check to run max once per week? Is it in the wildcard post event handler?

it has a check to run max once per week, the supabase should provide this
```ts
export const getLastWeeklyTime = async (): Promise<number> => {
  const { supabase } = getAdapters();

  const { data } = await supabase.from("weekly").select("last_time").limit(1).single();
  if (data) {
    return Number(data.last_time);
  } else {
    return 0;
  }
};
//src/adapters/supabase/helpers/client.ts
```
@pavlovcik

PR #120, opened 2023-02-14
Add Web-Assets Support

Steveantor:


pavlovcik:
I assume this is related to being able to load the fonts as we discussed?

Steveantor:
> I assume this is related to being able to load the fonts as we discussed? It could be simpler to save the font and host it with the bot. Either as the font file itself or base64 encoded!

the fonts will be loaded via Base64 #125 , the webAsset is for loading remote background.
@pavlovcik

PR #129, opened 2023-02-16
Automatic Assign & handle wallet registration

0xcodercrane:


0xcodercrane:
This PR is trying to resolve both issues attached above

src/handlers/wildcard/weekly/action.ts
```diff
@@ -8,8 +8,8 @@ import { BotConfig } from "../../../types";
 import { getFallback } from "../../../utils/fallback";
 import { fetchImage } from "../../../utils/webAssets";
 import { weeklyConfig } from "../../../configs/weekly";
+import { wait } from "../../../helpers";
```
pavlovcik, line 11: This wait looks like a timeout which raises a flag. What is this for?

pavlovcik:
Great progress to see! Not sure how I missed the original notification, sorry for the delayed review.

src/handlers/wildcard/weekly/action.ts
0xcodercrane: that fn has been created by @Steveantor  originally. I am not 100% sure but I guess its being used to wait for 1 min between github api calls because of its rate limit.

pavlovcik: It would be better to read the rate limit header that's passed with the GitHub request and change behavior based on that number I think.

PR #138, opened 2023-02-27
Feat/137 integrate permit2

0xcodercrane:


rndquu (approved):

src/configs/shared.ts
```diff
@@ -20,3 +20,8 @@ export const DEFAULT_BOT_DELAY = 100; // 100ms
  */
 export const DEFAULT_FOLLOWUP_TIME = "4 days"; // 4 days
 export const DEFAULT_DISQUALIFY_TIME = "7 days"; // 7 days
+
+export const DEFAULT_CHAIN_ID = 1;
+export const DEFAULT_RPC_ENDPOINT = "https://cloudflare-eth.com";
```
pavlovcik, line 25:
```suggestion
export const DEFAULT_RPC_ENDPOINT = "https://rpc-pay.ubq.fi/v1/mainnet";
```

I want to make dedicated RPC endpoints for all of our applications because its simple, free, and we can isolate logging metrics in case of abuse. I made this one the other day: 

https://github.dev/ubiquity/pay.ubq.fi/blob/773947bad8fd069adf99fb223373fcbdfab4600d/static/scripts/get-contract.js#L3

src/handlers/payout/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    log.info("Skipping to proceed the payment because `assignee` is undefined");
+    return;
+  }
+
+  if (!issueDetailed.priceLabel) {
+    log.info("Skipping to proceed the payment because price not set");
+    return;
+  }
+
+  const priceInETH = issueDetailed.priceLabel!.substring(7, issueDetailed.priceLabel!.length - 4);
```
pavlovcik, line 33:
```suggestion
  const priceInEth = issueDetailed.priceLabel!.substring(7, issueDetailed.priceLabel!.length - 4);
```

If we do the same code style as the dollar repo, its similar to us renaming all identifiers with `NFT` to `Nft`

src/helpers/permit.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Generates permit2 signature data with `spender` and `amountInETH`
+ *
+ * @param spender The recipient address we're going to send tokens
+ * @param amountInETH The token amount in ETH
+ *
+ * @returns Permit2 url including base64 encocded data
+ */
+export const generatePermit2Signature = async (spender: string, amountInETH: string): Promise<string> => {
```
pavlovcik, line 16:
```suggestion
export const generatePermit2Signature = async (spender: string, amountInEth: string): Promise<string> => {
```

src/main.ts
```diff
@@ -12,7 +12,5 @@ const UBIQUITY = `
 
 export default function main(app: Probot) {
   console.log(UBIQUITY);
-  // app.onAny(bindEvents as any);
-  app.on("issue_comment.created", bindEvents);
-  app.on("issue_comment.edited", bindEvents);
+  app.onAny(bindEvents as any);
```
pavlovcik, line 15: `as any` is code smell. You generally should strive to never use `as any` but instead, at least as a generic function definition or ideally a specific signature of a function.

pavlovcik (changes requested):

pavlovcik:
I see that conventional commits didn't pass. I am willing to let it slide but we should take the lint-staged hook from the `ubiquity-dollar` repo to minimize this from happening again. I'll set it as a new issue.

src/handlers/payout/index.ts
```diff
@@ -0,0 +1,44 @@
...
+  const recipient = await getWalletAddress(assignee);
+  if (!recipient) {
+    log.info(`Recipient address is missing`);
+    await addCommentToIssue(`@${assignee} would you please post your wallet address here?`, issue.number);
+    return;
+  }
+
+  const payoutUrl = await generatePermit2Signature(recipient, priceInETH);
+  log.info(`Posing a payout url to the issue, url: ${payoutUrl}`);
+  await addCommentToIssue(`@${assignee} You can receive your payment on ${payoutUrl}`, issue.number);
```
pavlovcik, line 43:
```suggestion
  await addCommentToIssue(`@${assignee} [ [ CLAIM ${CLAIM_AMOUNT} ${CLAIM_TOKEN_SYMBOL} ] ](https://pay.ubq.fi/?claim=${ENCODED_CLAIM})`, issue.number);
```

src/adapters/supabase/helpers/client.ts
```diff
@@ -152,7 +152,7 @@ export const upsertUser = async (user: UserProfile): Promise<void> => {
 /**
  * Performs an UPSERT on the wallet table.
  * @param username The user name you want to upsert a wallet address for
- * @param address The EVM wallet address
+ * @param address The account address
  */
 export const upsertWalletAddress = async (username: string, address: string): Promise<void> => {
```
pavlovcik, line 157: One of my colleagues had a security related concern about updating wallet addresses. Can you walk me through how an exploit could occur, and what measures are being put into place to combat these vectors?

The scenario he presented was if somebody is able to update all the wallets in our system and then front-run/claim new bounties posted. I think its unlikely to occur but I didn't have an answer for him as I am not 100% clear on the full architecture related to this system.

src/configs/shared.ts
pavlovcik: I haven't settled on a scalable naming formula but ideally a standardized way to name all these dedicated RPC endpoints.

src/handlers/payout/index.ts
pavlovcik: Regarding that security related point, we could consider posting their address as part of this comment as well. It could be even faster/easier to notice something is wrong. 
---
Another idea (not great) is that we could consider only supporting ENS and only allowing unique ENS names to be registered. That would make it very tedious for an attacker (but also kind of annoying for bounty hunters)

src/main.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/0a2f5d727febbba6a726141ec04f0e1f83ecb6ab

src/handlers/payout/index.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/0a2f5d727febbba6a726141ec04f0e1f83ecb6ab

src/helpers/permit.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/ca57bf0d3032801ad10d854af0a2708080254b8f

`BigNumber` would be replaced with `BigInt` since the latest ethers version so I feel like `string` would be more extensible

src/main.ts
0xcodercrane: Agree with your opinion not to use `as any` but sometimes it was kinda difficult to define the function definitions assignable what the Probot wants. but yeah let me try it without using `as any`

0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/7bf059daee4f16052e2d1f348e023e0428229622

src/helpers/permit.ts
pavlovcik: I'm pretty sure that the latest Ethers uses the built in JavaScript BigNumber primitive.

src/adapters/supabase/helpers/client.ts
0xcodercrane: we can keep talking on the discussion

src/handlers/payout/index.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/838ada446e5223d63893a4927f5d6ad0e7d4377b

src/handlers/comment/handlers/wallet.ts
```diff
@@ -1,10 +1,12 @@
...
 import { getBotContext } from "../../../bindings";
+import { addCommentToIssue } from "../../../helpers";
 import { Payload } from "../../../types";
 
 export const registerWallet = async (body: string): Promise<void> => {
   const { log, payload: _payload } = getBotContext();
   const payload = _payload as Payload;
+  const issue = payload.issue;
   const sender = payload.sender.login;
   const regex = /(0x[a-fA-F0-9]{40})/g;
```
pavlovcik, line 11: Would you also be able to handle ENS?

pavlovcik: It might make the most sense to ENS resolve at the moment of parsing. And if it resolves, just to save the raw address?

src/handlers/payout/index.ts
pavlovcik: Thanks! Please also add their wallet address to the comment!

src/helpers/permit.ts
0xcodercrane: Even with BigNumber, its a string internally. Lets use string instead of bignumber here.

src/handlers/comment/handlers/wallet.ts
0xcodercrane: This regex doesn't handle ENS

src/handlers/payout/index.ts
0xcodercrane: https://github.com/ubiquity/bounty-bot/pull/138/commits/1db28cd3f74f502ce9f5c6823c285bc1bf1fedfe

src/handlers/comment/handlers/wallet.ts
pavlovcik: I understand that, I'm saying to add it

pavlovcik: I wrote this the other day in normal javascript. It relies on some indie developer's Cloudflare function so please be sure to do error handling/logging properly in case they take down their endpoint.

https://github.com/ubiquity/pay.ubq.fi/blob/773947bad8fd069adf99fb223373fcbdfab4600d/static/scripts/render-transaction.js#L88

0xcodercrane: In lastest commits, it should be handled

pavlovcik (approved):

src/handlers/comment/handlers/wallet.ts
```diff
@@ -1,25 +1,20 @@
...
 import { getBotContext } from "../../../bindings";
 import { addCommentToIssue, resolveAddress } from "../../../helpers";
 import { Payload } from "../../../types";
+import { IssueCommentCommands } from "../commands";
 
+// Extracts ensname from raw text.
 const extractEnsName = (text: string): string | undefined => {
   // Define a regular expression to match ENS names
-  const ensRegex = /^([a-z0-9](?:-?[a-z0-9]){0,61}[a-z0-9])?\.([a-z]{2,})(\.eth)?$/i;
+  const ensRegex = /^(?=.{3,40}$)([a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z]{2,}$/gm;
```
pavlovcik, line 10: ENS can only be up to 40 characters? You sure about that?

PR #145, opened 2023-03-01
fix(weekly): update path resolution

Steveantor:


package.json
```diff
@@ -14,7 +14,7 @@
     "build": "tsc",
     "build:ci": "ncc build src/index.ts -o dist && mv ./dist/index.js ./bin",
     "start": "probot run ./lib/main.js",
-    "start:watch": "nodemon --exec 'yarn start'",
+    "start:watch": "copy-folder-util src/assets/images lib/assets/images && nodemon --exec 'yarn start'",
```
pavlovcik, line 17: This seems really unnecessary to add this dependency. What's the issue with `cp -r` or `rsync`

pavlovcik (changes requested):
I generally disagree with adding dependencies unless they are really needed. In this case, I don't see why we can't use built in binaries like `cp`

package.json
Steveantor: I'm thinking about cross platform support, whichever command or dependency works
@pavlovcik

pavlovcik: I thought that `cp` works for windows lol

pavlovcik:
Let me one up you with an idea, why not use a symlink instead of cp for this?

pavlovcik:
Mixed feelings about merging that in but its such a minor thing I figured no need to block you on it.

PR #147, opened 2023-03-02
refactor: replace copy-util with native

Steveantor:


copy.js
```diff
@@ -0,0 +1,9 @@
+const fs = require("fs");
+
+fs.cp("src/assets/images", "lib/assets/images", { recursive: true }, (err) => {
+  if (err) {
+    throw new Error(err);
+  } else {
+    console.log("successfully copied files");
+  }
+});
```
pavlovcik, line 9: Questionable implementation but better than a new dependency lol

Steveantor: symlinks require privilage, and most projects tend to copy files to avoid inconsistencies.

pavlovcik: Not sure what you mean by privilage. You don't need sudo access to symlink.

Steveantor: was referring to windows, linux by default doesn't need it (may vary depending on the distro/policy)

PR #149, opened 2023-03-02
Refactor/env properties

pavlovcik:
Not sure how exactly you are accessing the GitHub secrets so I am not sure I did this correctly.


src/bindings/config.ts
```diff
@@ -31,9 +31,9 @@ export const loadConfig = async (): Promise<BotConfig> => {
...
       chainId: process.env.CHAIN_ID ? Number(process.env.CHAIN_ID) : DEFAULT_CHAIN_ID,
       rpc: process.env.RPC_PROVIDER_URL || DEFAULT_RPC_ENDPOINT,
-      privateKey: process.env.UBIQUIBOT_PRIVATE_KEY || "",
+      privateKey: process.env.EVM_SIGNING_KEY || "",
       paymentToken: process.env.PAYMENT_TOKEN || DEFAULT_PAYMENT_TOKEN,
-      permitBaseUrl: process.env.PERMIT_BASE_URL || DEFAULT_PERMIT_BASE_URL,
+      permitBaseUrl: process.env.PAYMENT_DAPP_DOMAIN || DEFAULT_PAYMENT_DAPP_DOMAIN,
     },
     unassign: {
       followUpTime: ms(process.env.FOLLOW_UP_TIME || DEFAULT_FOLLOWUP_TIME),
```
pavlovcik, line 39: I think that every config should throw an error if it is unset. Unless you support graceful fallback. For example if no `privateKey` is set, then perhaps we can disable payouts for that bot instance?

0xcodercrane: the app throws itself if the require config params are unset.

0xcodercrane:
This PR needs to be merged with the secrets key update

pavlovcik:
> This PR needs to be merged with the secrets key update

Apologies but I am not sure what you are referring to. Are you saying we need to change the property names on the secrets? Let's work together tomorrow on this?

0xcodercrane:
> > This PR needs to be merged with the secrets key update
> 
> Apologies but I am not sure what you are referring to. Are you saying we need to change the property names on the secrets? Let's work together tomorrow on this?

Yes property names on the secrets. I will be ready to work together

PR #201, opened 2023-03-26
Feat: New Bounty Hunter Tutorial

### LINKED ISSUE #183: Tutorial for new Bounty Hunters
When a hunter posts for the first time we should display the /help menu and a greeting so that they know what they can handle themselves (in this case if they knew about the assign function that would be great)

---
Hi @pavlovcik 

Can you assign this bounty to me? 

_Originally posted by @jpldcarvalho in https://github.com/ubiquity/ubiquity-dollar/issues/578#issuecomment-1459925989_

Steveantor:
Resolves #183

### DIFF
--- a/src/handlers/comment/handlers/first.ts
+++ b/src/handlers/comment/handlers/first.ts
@@ -1,0 +1,23 @@
+import { getBotContext, getLogger } from "../../../bindings";
+import { COMMAND_INSTRUCTIONS } from "../../../configs";
+import { addCommentToIssue } from "../../../helpers";
+import { Payload } from "../../../types";
+
+export const verifyFirstCheck = async (): Promise<void> => {
+  const context = getBotContext();
+  const logger = getLogger();
+  const payload = context.payload as Payload;
+
+  try {
+    const response = await context.octokit.rest.search.issuesAndPullRequests({
+      q: `is:issue repo:${payload.repository.owner.login}/${payload.repository.name} commenter:${payload.sender.login}`,
+      per_page: 2,
+    });
+    if (response.data.total_count >= 2) {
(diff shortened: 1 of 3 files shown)


src/handlers/comment/handlers/first.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  try {
+    const response = await context.octokit.rest.search.issuesAndPullRequests({
+      q: `is:issue repo:${payload.repository.owner.login}/${payload.repository.name} commenter:${payload.sender.login}`,
+      per_page: 2,
+    });
+    if (response.data.total_count >= 2) {
+      const msg = `${COMMAND_INSTRUCTIONS}\n@${payload.sender.login}`;
+      await addCommentToIssue(msg, payload.issue!.number);
+    }
```
pavlovcik, line 19: If I'm understanding this logic correctly, its checking for any comment by the bounty hunter in any issue or pull request. I think this might be a viable strategy and we can always iterate on it if we find something superior. This seems to check out. How can we easily test if its working @0xcodercrane ?

Steveantor: Yes,
It checks for whether the user has previously issue commented on the repo,
If not that means he's a newcomer. 
@pavlovcik

pavlovcik: GitHub UI seems to have a label for "new contributor" I was hoping this is directly accessible from the GitHub API?

Steveantor: I couldn't find the exact spec for the `new contributor` tag.
it might be for first issue or pr
@pavlovcik

PR #208, opened 2023-03-26
feat: unassign command

### LINKED ISSUE #207: `/unassign` command
This should allow a bounty hunter to release a bounty back into the DevPool (assign to @ubiquity-bounties)

              We need to actually create the `/unassign` function for this feature to work!

_Originally posted by @pavlovcik in https://github.com/ubiquity/bounty-bot/pull/196#pullrequestreview-1357989238_

Steveantor:
Resolves #207

### DIFF
--- a/src/handlers/comment/handlers/unassign.ts
+++ b/src/handlers/comment/handlers/unassign.ts
@@ -1,0 +1,33 @@
+import { addAssignees, addCommentToIssue, removeAssignees } from "../../../helpers";
+import { getBotContext, getLogger } from "../../../bindings";
+import { Payload } from "../../../types";
+import { IssueCommentCommands } from "../commands";
+import { GLOBAL_STRINGS } from "../../../configs";
+
+export const unassign = async (body: string) => {
+  const { payload: _payload } = getBotContext();
+  const logger = getLogger();
+  if (body != IssueCommentCommands.UNASSIGN && body.replace(/`/g, "") != IssueCommentCommands.UNASSIGN) {
+    logger.info(`Skipping to unassign. body: ${body}`);
+    return;
+  }
+
+  const payload = _payload as Payload;
+  logger.info(`Received '/unassign' command from user: ${payload.sender.login}`);
+  const issue = (_payload as Payload).issue;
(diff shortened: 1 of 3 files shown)


src/handlers/comment/handlers/unassign.ts
```diff
@@ -0,0 +1,38 @@
...
+    return;
+  }
+
+  const issue_number = issue!.number;
+  const _assignees = payload.issue?.assignees;
+  const assignees = _assignees ?? [];
+  let shouldUnassign = false;
+  assignees.forEach((assignee) => {
+    if (assignee === payload.sender.login) {
+      shouldUnassign = !shouldUnassign && true;
```
pavlovcik, line 29: I'm confused about this `&& true` syntax

pavlovcik:
This needs to "unassign" by removing all assignees, and then assigning @ubiquity-bounties for now. Please make sure to pull the name from the shared "reserved user names" array or whatever it is, instead of hardcoding in the function.

Also we will be deprecating that username soon, so however you implement this should be easy to catch to update as well ideally.

Steveantor:
> This needs to "unassign" by removing all assignees, and then assigning @ubiquity-bounties for now. Please make sure to pull the name from the shared "reserved user names" array or whatever it is, instead of hardcoding in the function.
> 
> Also we will be deprecating that username soon, so however you implement this should be easy to catch to update as well ideally.

This is what it does,
- remove previous assignees
- Add reserved assignees `GLOBAL_STRING.assignees`

@pavlovcik

src/handlers/comment/handlers/unassign.ts
Steveantor: it's basically shouldUnassign = true

0xcodercrane: We want 1 active assignee at a time so you can update it like 
`shouldUnassign = assignees.length > 0 && payload.sender.login === assignees[0]`

0xcodercrane (changes requested):

0xcodercrane (approved):

PR #235, opened 2023-04-07
Feat: Wallet Address Check

### LINKED ISSUE #230: Replace Deadline Message
We currently don't enforce the deadlines so the bot commenting the deadline for the bounty immediately when the issue is assigned isn't useful. We should replace this comment with the following:

It would be useful for the bot to check if the assignee's wallet address is registered. 

If it is not registered, the bot should request that the assignee registers their wallet address, explaining that it is required in order to be paid for the bounty.  

If it is registered, the bot should let the assignee know which address is currently registered.

Steveantor:
Resolves #230


0xcodercrane (approved):
Lgtm!

PR #236, opened 2023-04-07
Dynamic menu

### LINKED ISSUE #233: Dynamic Help Menu
We should have a central object that stores all of the commands, as well as the help menu description of each. This way, if any new commands are created, the developer is forced to also write a description for it and it will automatically populate in the help menu.

Pseudo code because on phone:

```
const userCommands = {

assign: { 
  handler: Function;
  description: String;
},

help: { 
  handler: Function;
  description: String;
},

disableAutoPay: { 
  handler: Function;
  description: String;
},

recruit: { 
  handler: Function;
  description: String;
},

}
```

We could also separate and make two other objects, `treasurerCommands` and `adminCommands` see #221 for more context on these roles.

devpanther:
Resolves #233


src/configs/shared.ts
```diff
@@ -1,3 +1,5 @@
+import { generateHelpMenu } from "../handlers/comment/handlers";
```
0xcodercrane, line 1: you can simply import as `import { generateHelpMenu } from "../handlers"`

src/configs/shared.ts
```diff
@@ -25,13 +27,13 @@ export const DEFAULT_RPC_ENDPOINT = "https://rpc-pay.ubq.fi/v1/mainnet";
...
+//         ex: /assign
 
-- /wallet <WALLET_ADDRESS | ENS_NAME>: Register the hunter's wallet address.
-	ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
-	ex2: /wallet vitalik.eth\n` +
-  "```";
+// - /wallet <WALLET_ADDRESS | ENS_NAME>: Register the hunter's wallet address.
+// 	ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
+// 	ex2: /wallet vitalik.eth\n` +
+//   "```";
```
0xcodercrane, line 39: We should remove them if dynamic menu is working properly

0xcodercrane (changes requested):

src/types/handlers.ts
```diff
@@ -14,3 +14,10 @@ export type Handler = {
   action: ActionHandler[];
   post: PostActionHandler[];
 };
+
+// UserCommandsSchema
+export type UserCommandsSchema = {
```
0xcodercrane, line 19: The name looks wrong to me. If this is `type`, it needs to be like `UsersCommands`. Also I think `HelpCommands` sounds more reasonable.

devpanther:
@0xcodercrane requested changes implemented

0xcodercrane:
> @0xcodercrane requested changes implemented

I pushed a couple of refactors to your branch as well.

0xcodercrane (approved):

PR #242, opened 2023-04-10
GitHub ratelimit

### LINKED ISSUE #134: Rate Limit Header
It would be better to read the rate limit header that's passed with the GitHub response and change behavior based on that number I think. 

_Originally posted by @pavlovcik in https://github.com/ubiquity/bounty-bot/pull/129#discussion_r1115189897_

devpanther:
Fixes #134


pavlovcik (approved):

0xcodercrane (changes requested):
I will review this PR more deeply cause it seems not including all the GitHub API requests. 
Let me keep @devpanther updated on that

pavlovcik:
> > it seems not including all the GitHub API requests.
> 
> I don't get what you mean

I think what @0xcodercrane means is that your implementation should be a general utility function that we call before every request to any GitHub API endpoint across all of the features of the bot.

devpanther:
> > > it seems not including all the GitHub API requests.
> > 
> > 
> > I don't get what you mean
> 
> I think what @0xcodercrane means is that your implementation should be a general utility function that we call before every request to any GitHub API endpoint across all of the features of the bot.

I think that's the only time Github API is called and requires a rate limiter, all other requests are POSTs (fetching issues and removing/adding assignees)..

devpanther:
Screenshot 2023-04-11 at 10 22 39 AM
This allowed @pavlovcik

0xcodercrane:
> I think what @0xcodercrane means is that your implementation should be a general utility function that we call before every request to any GitHub API endpoint across all of the features of the bot.

Yep! exactly!  That's originally what I want. by the way, it sounds wrong its the only place where the Github API is called. 

We have a couple of files more to call Github API. Please take a look at the following example files.
https://github.com/ubiquity/bounty-bot/blob/development/src/helpers/issue.ts
https://github.com/ubiquity/bounty-bot/blob/development/src/helpers/label.ts

devpanther:
Okay, will check em out.. Thanks

devpanther:
@0xcodercrane i see issue file requires a retry.. but is there any need to do that in the label file.

They're just post/get requests called once unlike the get requests in a while loop

PR #259, opened 2023-04-11
258 security issue

0xcodercrane:


pavlovcik:
アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:46]
We won’t custody partner funds
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:46]
Don’t want that liability
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:47]
Technically if they mint our stablecoin we will do a version of custody but that’s beyond the scope of this conversation lol
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:48]
Whitelist is not a good approach that doesn’t scale
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:49]
We should think of an architecture that allows people to use devpool permissionlessly
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:49]
I mean, sure it could be a temporary fix but we should have a discussion and come up with a better idea. I’m sure it won’t be complicated to implement; just requires a little research
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:51]
I made the tutorial page for permits for the reason that partners should be able to set up their own automated payments
 アレクサンダー.eth | U B I Q U I T Y, [Apr 12, 2023 at 08:51]
Automatic payments in pretty sure was only in the alpha milestone of the bot

PR #267, opened 2023-04-22
Feat: Orgwide Keychain Management

### LINKED ISSUE #262: Config split
Right now we have a shared private key with partners for permit2 feature. Partner should be able to set a private key in a private config repository.

Partner should be able to set 2 configs:
1. Organization wide config file located in the `ubiquibot-config` repository at `OWNER/ubiquibot-config/.github/ubiquibot-config.yml` path.
2. Repository wide config file located at a target repository at `OWNER/REPOSITORY/.github/ubiquibot-config.yml` path.

Repository wide config takes precedence over organization wide one. Notice that a repository wide config may not exist.

Sample reference code of how to read file content from a private repository: https://gist.github.com/rndquu/fb13523b8b9536cb7467d96ce6a6f021

So for a partner project the flow will be the following:
1. Create a private repository `ubiquibot-config`
2. Create a config file `.github/ubiquibot-config.yml` with a private key
3. Add the bot to the config repository with the `contents` repository permission

**What should be done**:
1. Implement config split
2. Read private key for the permit2 feature from the config repository 

Notice that here https://github.com/ubiquity/bounty-bot/blob/development/src/bindings/config.ts#L34 it is better to read both from the config and ENV variable for backwards compatibility. When this task is finished we should update the code and read only from the config.

Steveantor:
**.env Setup**

> X25519_PUBLIC_KEY = public key
> X25519_PRIVATE_KEY = private key

Follow [pay.ubq.fi@43](https://github.com/ubiquity/pay.ubq.fi/pull/43) to generate keys

Resolves #262


src/bindings/config.ts
```diff
@@ -13,10 +13,13 @@ import {
 } from "../configs";
 import { ajv } from "../utils";
 import { Context } from "probot";
+import { getPrivateKey } from "../utils/private";
 
 export const loadConfig = async (context: Context): Promise<BotConfig> => {
   let configFile: any = {};
   configFile = await context.config("ubiquibot-config.yml");
+  const keyData = await getPrivateKey();
+  const privateKey = keyData !== undefined ? keyData : process.env.UBIQUITY_BOT_EVM_PRIVATE_KEY !== undefined ? process.env.UBIQUITY_BOT_EVM_PRIVATE_KEY : "";
```
0xcodercrane, line 22: `  const privateKey = keyData ?? process.env.UBIQUITY_BOT_EVM_PRIVATE_KEY  ?? "";`

we can make it simple as above

src/bindings/config.ts
```diff
@@ -51,6 +54,10 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
       autoPayMode: process.env.AUTO_PAY_MODE === "TRUE" ? true : configFile?.autoPayMode ?? true,
       analyticsMode: process.env.ANALYTICS_MODE === "TRUE" ? true : configFile?.analyticsMode ?? false,
     },
+    sodium: {
+      publicKey: process.env.X25519_PUBLIC_KEY ?? "",
+      privateKey: process.env.X25519_PRIVATE_KEY ?? "",
+    },
```
0xcodercrane, line 60: I am not fully understanding why we setup both keys. Both of them wouldn't be put in the partner repo.

src/utils/private.ts
```diff
@@ -0,0 +1,49 @@
...
+    const {
+      sodium: { publicKey, privateKey },
+    } = getBotConfig();
+    const payload = context.payload as Payload;
+
+    if (publicKey === "" || privateKey === "") {
+      return undefined;
+    }
+
+    const { data } = await context.octokit.rest.repos.getContent({
```
0xcodercrane, line 24: Would you give me an example how `ubiquibot-config.yml` looks like? If there's some steps needed for integrators to create this file rather than putting raw private_key, we might need to update docs as well

cuddle-dev (changes requested):
A couple of notes to make sure I am not missing anything from the origin spec.

src/bindings/config.ts
Steveantor: updated

src/bindings/config.ts
Steveantor: We're using `libsodium`, hence we need both the private key and public key,
follow the linked PR to generate one.

It's to be noted that the partners don't need to generate those keys, and those keys are only meant to 
be for bot constants (for instance: the _**installation key**_ is also a bot constant)

src/utils/private.ts
Steveantor: I've linked another PR for the wizard part,
the `ubiquibot-config.yml` should currently look like this:

```json
{
  "PSK": "<Encrypted_Private_Key_Value>"
}
```

Steveantor: I've linked another PR for the wizard part,
the `ubiquibot-config.yml` should currently look like this:

```yml
  PSK: "<Encrypted_Private_Key_Value>"
```

src/utils/private.ts
```diff
@@ -0,0 +1,49 @@
...
+      owner: payload.organization?.login!,
+      repo: CONFIG_REPO,
+      path: KEY_PATH,
+      mediaType: {
+        format: "raw",
+      },
+    });
+
+    if (data) {
+      const cipherText = (data as any).startsWith(KEY_NAME) ? (data as any).replace(`${KEY_NAME}: `, "").replace(`"`, "") : undefined;
```
rndquu, line 34: Isn't it better to use a [yaml parser](https://www.npmjs.com/package/yaml) here?

If there are new variables added then it is no more possible to get only a private key

src/utils/private.ts
```diff
@@ -0,0 +1,49 @@
...
+      if (cipherText !== undefined) {
+        const binPub = sodium.from_base64(publicKey, sodium.base64_variants.URLSAFE_NO_PADDING);
+        const binPriv = sodium.from_base64(privateKey, sodium.base64_variants.URLSAFE_NO_PADDING);
+        const binCipher = sodium.from_base64(cipherText, sodium.base64_variants.URLSAFE_NO_PADDING);
+
+        let walletPrivateKey: string | undefined = sodium.crypto_box_seal_open(binCipher, binPub, binPriv, "text");
```
rndquu, line 40: So the whole idea is that a partner should keep his private key encrypted in a private repo, right?

So for a partner the flow is the following:
1. Take bot's public key
2. Encrypt partner's wallet private key with bot's public key
3. Save encrypted value in a private repo

src/bindings/config.ts
```diff
@@ -31,7 +33,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     payout: {
       chainId: process.env.CHAIN_ID ? Number(process.env.CHAIN_ID) : DEFAULT_CHAIN_ID,
       rpc: process.env.RPC_PROVIDER_URL || DEFAULT_RPC_ENDPOINT,
-      privateKey: process.env.UBIQUITY_BOT_EVM_PRIVATE_KEY || "",
+      privateKey: keyData ?? process.env.UBIQUITY_BOT_EVM_PRIVATE_KEY ?? "",
```
rndquu, line 36: If we make partners to encrypt their wallets' private keys then we don't need a separate private repo for a config because we can simply store an encrypted private key in a standard bot's config

Then this page https://github.com/ubiquity/pay.ubq.fi/pull/43 transforms into a single field which encrypts partner wallet's private key with the bot's public key

Steveantor: The private repo is for additional secrecy, despite them being encrypted.
For the same reason, GitHub doesn't allow reading secret values through API (yes they're encrypted)

src/utils/private.ts
Steveantor: that's right

src/bindings/config.ts
rndquu: Yes, if the bot's PK is leaked then all encrypted wallet PKs in public repos are at risk so it is better to have a private repo

rndquu (commented):
There is no concept of config split in this PR as described in the original issue

I mean if there is only an organization wide config (without any repo wide configs) then the bot won't be able find and use the needed config

Steveantor:
> There is no concept of config split in this PR as described in the original issue
> 
> I mean if there is only an organization wide config (without any repo wide configs) then the bot won't be able find and use the needed config

the PR followed the suggested steps:

1. Create a private repository ubiquibot-config
2. Create a config file .github/ubiquibot-config.yml with a private key
3. Add the bot to the config repository with the contents repository permission

could you describe additional steps that need to be taken?

rndquu:
> > There is no concept of config split in this PR as described in the original issue
> > I mean if there is only an organization wide config (without any repo wide configs) then the bot won't be able find and use the needed config
> 
> the PR followed the suggested steps:
> 
> 1. Create a private repository ubiquibot-config
> 2. Create a config file .github/ubiquibot-config.yml with a private key
> 3. Add the bot to the config repository with the contents repository permission
> 
> could you describe additional steps that need to be taken?

```
Partner should be able to set 2 configs:
1. Organization wide config file located in the `ubiquibot-config` repository at `OWNER/ubiquibot-config/.github/ubiquibot-config.yml` path.
2. Repository wide config file located at a target repository at `OWNER/REPOSITORY/.github/ubiquibot-config.yml` path.

Repository wide config takes precedence over organization wide one. Notice that a repository wide config may not exist.
```

Simply put [here](https://github.com/ubiquity/bounty-bot/blob/development/src/bindings/config.ts#L19) we should:
1. Read organization wide config
2. Read repository wide config
3. Merge those 2 configs (repository wide config values take precedence over organization wide config values)
4. Keep code execution with this combined config

package.json
```diff
@@ -40,16 +40,19 @@
     "husky": "^8.0.2",
     "jimp": "^0.22.4",
     "js-yaml": "^4.1.0",
+    "libsodium-wrappers": "^0.7.11",
```
pavlovcik, line 43: I presume this is for salting given the `sodium` term

Steveantor: yeah like that lol

src/bindings/config.ts
pavlovcik: Public keys should be able to be derived from private keys. We shouldn't burden the users to pass in more information than absolutely needed to operate the bot.

0x4007 (changes requested):
Can you generate the public key from the private key easily or is it a huge project?

src/bindings/config.ts
0xcodercrane: hmm I think I understood. so you were trying to use asymmetric encryption method. that's good.

Steveantor:
> Can you generate the public key from the private key easily or is it a huge project?

It depends on the algorithm,
but what's your use case?

src/utils/private.ts
0xcodercrane: @rndquu is right. we shouldn't parse it manually. A couple of yaml parser exist as npm packages. lets use either of them.

Steveantor: updated with the yml parser

src/bindings/config.ts
Steveantor: > Public keys should be able to be derived from private keys. We shouldn't burden the users to pass in more information than absolutely needed to operate the bot.

the end users don't need to pass those values at all

Steveantor:
updated

0x4007 (approved):

src/utils/private.ts
```diff
@@ -0,0 +1,157 @@
...
+
+export const getWideConfig = async (context: Context): Promise<DataConfig> => {
+  const orgConfig = await getConfigSuperset(context, "org");
+  const repoConfig = await getConfigSuperset(context, "repo");
+
+  const parsedOrg: WideOrgConfig | undefined = await parseYAML(orgConfig);
+  const parsedRepo: WideRepoConfig | undefined = await parseYAML(repoConfig);
+  const PSK = parsedOrg && parsedOrg[KEY_NAME] ? await getPrivateKey(parsedOrg[KEY_NAME]) : undefined;
+
+  const configData: DataConfig = {
```
rndquu, line 122: Can't we merge configs together like `Object.assign(parsedOrg, parsedRepo);` ?

Steveantor: > Can't we merge configs together like `Object.assign(parsedOrg, parsedRepo);` ?

There are additional checks that needs to be passed, Object.assign is prone to errors.

rndquu: My point is that (at least for me) this is really hard to read:
```
parsedRepo && parsedRepo["baseMultiplier"] && !Number.isNaN(Number(parsedRepo["baseMultiplier"]))
        ? Number(parsedRepo["baseMultiplier"])
        : parsedOrg && parsedOrg["baseMultiplier"] && !Number.isNaN(Number(parsedOrg["baseMultiplier"]))
        ? Number(parsedOrg["baseMultiplier"])
        : Number(DefaultPriceConfig["baseMultiplier"]),
``` 

So I'm trying to figure out the way to make it cleaner, perhaps by deep merging 2 yaml configs

Steveantor: you need to tinker around ternary to get used to it.

pavlovcik: We shouldn't use ternary or any shorthand. 

Code readability and maintenance outweighs the benefits of saving a few keystrokes.

Steveantor: refactored

cuddle-dev (approved):

0x4007:
What was the purpose of encrypting the private key again? Why not just add to the key directly to the organization secrets?

I presume it had something to do with being able to log it out in the code but we can do the same thing right now after it’s decrypted.

However, key encryption wasn't in the spec.

For context: I'm currently working on refactoring the whole codebase.

rndquu:
Reasons behind the private key encryption:
1. If an org wide config is accidentally made public then private key encryption prevents funds leakage
2. Malicious organisation members can't read the key

PR #281, opened 2023-05-03
Comment Handler

### LINKED ISSUE #269: Comment callback handler
Be sure to clean up any deprecated code, such as all of the function calls to manually invoke commenting after all of the other operations.

              > Can you make sure that if there is an error of some sort that it lets us know in a comment?

> From an architecture perspective, a little bit like how I proposed the dynamic help menu architecture, perhaps we can make sort of a standard callback function after a command handler is invoked to always leave a comment/response. In this way, the bounty-bot developer would intentionally have to instruct the bot to not leave a comment/response after a bot command is invoked.

_Originally posted by @pavlovcik in https://github.com/ubiquity/bounty-bot/issues/192#issuecomment-1502298930_

devpanther:
resolves #269


devpanther:
Not building, i can't see any errors @pavlovcik @0xcodercrane

0xcodercrane:
image

you have to push yarn.lock if there's a new module added in package.json.  @devpanther

devpanther:
but i think it still fails @0xcodercrane

pavlovcik:
The best way to diagnose this problem is to clone fresh, run `gh pr checkout 281` to check out this branch, and then running it.

devpanther:
> The best way to diagnose this problem is to clone fresh, run `gh pr checkout 281` to check out this branch, and then running it.

I don't think it has anything to do with my PR, its broken repo-wide

pavlovcik:
We're investigating a fix. We have the issue diagnosed. Sorry about that!

devpanther:
> We're investigating a fix. We have the issue diagnosed. Sorry about that!

Can this be reviewed, this contains significant changes to the slash commands and i see alot of merge conflicts if its not a base soon

devpanther:
> We're investigating a fix. We have the issue diagnosed. Sorry about that!

Can this be reviewed, this contains significant changes to the slash commands and i see alot of merge conflicts if its not a base soon @pavlovcik @0xcodercrane

pavlovcik (approved):
Quickly looked and it seems like you implemented correctly!

0xcodercrane:
The issue came from different one. not the issue of this PR.

src/helpers/issue.ts
```diff
@@ -118,7 +118,8 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
 
       await checkRateLimitGit(response?.headers);
 
-      if (response.data) {
+      // Fixing infinite loop here, it keeps looping even when its an empty array
+      if (response.data && response.data.length > 0) {
```
0xcodercrane, line 122: `It keeps looping even when its an empty array` --- I am not sure whether it keep looping if its an empty array but also interesting. If you have more bandwidth, would you be posting the simple proof like 1 ~ 2 line of javascript code

src/handlers/processors.ts
```diff
@@ -35,7 +37,19 @@ export const processors: Record<string, Handler> = {
...
+        const { payload: _payload } = getBotContext();
+        const issue = (_payload as Payload).issue;
+        try {
+          const comment = await handleIssueClosed();
+          return await addCommentToIssue(comment!, issue!.number);
+        } catch (err: any) {
+          return await addCommentToIssue("Error: " + err.message, issue!.number);
+        }
+      },
+    ],
```
0xcodercrane, line 52: I don't agree to have it here. You can update action handler type properly or do whatever you can. but keep the previous one. 

`action: [handleIssueClosed]`

src/handlers/comment/handlers/index.ts
```diff
@@ -34,30 +35,48 @@ export const userCommands: UserCommands[] = [
     id: IssueCommentCommands.ASSIGN,
     description: "Assign the origin sender to the issue automatically.",
     handler: assign,
+    callback: async (issue_number, comment) => {
+      await addCommentToIssue(comment, issue_number!);
+    },
```
0xcodercrane, line 40: This block has been duplicated in the next lines. Please define/implement a callback function and just use it here like the following.

```typescript
    id: IssueCommentCommands.ASSIGN,
    description: "Assign the origin sender to the issue automatically.",
    handler: assign,
    callback: commandCallback
```

you might need to create command callback function somewhere

0xcodercrane (changes requested):

src/helpers/issue.ts
pavlovcik, line 122:
```suggestion
      if (response?.data?.length > 0) {
```

Could be more concise.

devpanther: OKay, I'd go with the concise suggestion from pavlovcik, but I'd do that next time.

I ran the payout command and saw it loop endlessly so just a quick fix

src/handlers/processors.ts
devpanther: Okay, i'd look for something

src/handlers/comment/handlers/index.ts
devpanther: Oh yeah, i got that... refactoring now

src/helpers/issue.ts
devpanther: > Could be more concise.

Yea, i guess that'd be more consise.. Thanks

0xcodercrane (approved):

PR #295, opened 2023-05-04
Auto devpool pr

### LINKED ISSUE #278: Add New Partner Repositories via Pull Request via `installation_repositories.added` Event
The bot should create a PR with a partner project URL to this file https://github.com/ubiquity/devpool/blob/development/projects.json if partner's URL does not exist in that list and if the repository is public.

According to https://github.com/ubiquity/bounty-bot/pull/295#discussion_r1186057386 `installation_repositories.added` event is fired when the bot is added to a new repository. This event handler should automatically add the necessary labels to a new repository, and also register the repository to this list via a pull request.

---

>               > Added more projects we're working on.
> > 
> > I'm debating on the benefits of whitelisting. This approach, we are whitelisting projects and adding them to this directory, but isn't it in our best interest to collect every project automatically and then allow for more sophisticated filtering tools?
> 
> Yes, perhaps we could automatically add new partner projects to https://github.com/ubiquity/devpool/blob/development/projects.json or at least the bot could create a PR with a partner project URL.
> 
> Also it's in partner's best interest to do their best to make sure their project is available to as many developers as possible. Opening a PR is a small effort + it's kind of a common practise. DefiLlama has the [same concept](https://github.com/DefiLlama/DefiLlama-Adapters) of adding partner projects.

_Originally posted by @rndquu in https://github.com/ubiquity/devpool/issues/433#issuecomment-1526519728_

seprintour:
resolves #278

### DIFF
--- a/src/bindings/event.ts
+++ b/src/bindings/event.ts
@@ -92,9 +92,12 @@ export const bindEvents = async (context: Context): Promise<void> => {
     await postAction();
   }
 
-  // Run wildcard handlers
-  logger.info(`Running wildcard handlers: ${wildcardProcessors.map((fn) => fn.name)}`);
-  for (const wildcardProcessor of wildcardProcessors) {
-    await wildcardProcessor();
+  // Skip wildcard handlers for installation event
+  if (eventName !== GithubEvent.INSTALLATION_ADDED_EVENT) {
+    // Run wildcard handlers
+    logger.info(`Running wildcard handlers: ${wildcardProcessors.map((fn) => fn.name)}`);
+    for (const wildcardProcessor of wildcardProcessors) {
+      await wildcardProcessor();
+    }
   }
 };
--- a/src/bindings/event.ts
+++ b/src/bindings/event.ts
@@ -48,20 +48,23 @@ export const bindEvents = async (context: Context): Promise<void> => {
   logger.info("Creating adapters for supabase, telegram, twitter, etc...");
   adapters = createAdapters(botConfig);
 
-  // Validate payload
-  const validate = ajv.compile(PayloadSchema);
-  const valid = validate(payload);
-  if (!valid) {
-    logger.info("Payload schema validation failed!!!", payload);
-    logger.warn(validate.errors!);
-    return;
-  }
+  // Skip validation for installation event
+  if (eventName !== GithubEvent.INSTALLATION_ADDED_EVENT) {
+    // Validate payload
+    const validate = ajv.compile(PayloadSchema);
+    const valid = validate(payload);
+    if (!valid) {
+      logger.info("Payload schema validation failed!!!", payload);
+      logger.warn(validate.errors!);
+      return;
+    }
 
-  // Check if we should skip the event
-  const { skip, reason } = shouldSkip();
-  if (skip) {
-    logger.info(`Skipping the event. reason: ${reason}`);
-    return;
+    // Check if we should skip the event
+    const { skip, reason } = shouldSkip();
+    if (skip) {
+      logger.info(`Skipping the event. reason: ${reason}`);
+      return;
+    }
   }
 
   // Get the handlers for the action
--- a/src/handlers/pull-request/createDevPoolPR.ts
+++ b/src/handlers/pull-request/createDevPoolPR.ts
@@ -1,0 +1,63 @@
+import { getBotContext } from "../../bindings";
+
+export const createDevPoolPR = async () => {
+  console.log("new install");
+  const context = getBotContext();
+  // replace with the owner and name of the repository you want to create the branch in
+  const owner = "seprintour";
+  const repo = "betterbuy";
+
+  // create a new branch
+  const branchName = "my-branch";
+  const baseRef = "master"; // replace with the name of the base branch
+
+  const { data: branch } = await context.octokit.repos.getBranch({
+    owner,
+    repo,
+    branch: "master",
+  });
+
+  const { data: file } = await context.octokit.repos.getContent({
+    owner,
+    repo,
+    path: "projects.json",
+    ref: "master",
+  });
+
+  const curContent = Buffer.from(JSON.parse(JSON.stringify(file)).content!, "base64").toString();
+
+  console.log(curContent);
+
+  // const mainSha = branch.commit.sha;
+
+  // await context.octokit.git.createRef({
+  //   owner,
+  //   repo,
+  //   ref: `refs/heads/${branchName}`,
+  //   sha: mainSha,
+  // });
+
+  // // create a new file on the branch
+  // const path = 'path/to/my/file.txt';
+  // const content = 'This is my new content';
+  // const createFileResponse = await context.octokit.repos.createOrUpdateFileContents({
+  //   owner,
+  //   repo,
+  //   path,
+  //   message: 'Add file',
+  //   content: Buffer.from(content).toString('base64'),
+  //   branch: branchName
+  // });
+  // console.log(createFileResponse.data);
+
+  // // create the pull request
+  // const pullRequest = await context.octokit.pulls.create({
+  //   owner,
+  //   repo,
+  //   title: 'My pull request',
+  //   head: branchName,
+  //   base: baseRef,
+  //   body: 'This is my pull request',
+  // });
+  // console.log(pullRequest.data);
+};
--- a/src/handlers/pull-request/createDevPoolPR.ts
+++ b/src/handlers/pull-request/createDevPoolPR.ts
@@ -2,9 +2,13 @@ import { getBotContext, getLogger } from "../../bindings";
 import { Payload } from "../../types";
 
 export const createDevPoolPR = async () => {
-  const context = getBotContext();
   const logger = getLogger();
+
+  const context = getBotContext();
   let payload = context.payload as Payload;
+
+  let devPoolOwner = "ubiquity";
+  let devPoolRepo = "devpool";
 
   if (payload.repositories_added?.length === 0) {
     return;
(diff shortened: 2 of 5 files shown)


seprintour:
@pavlovcik @rndquu Ready for review

src/handlers/pull-request/createDevPoolPR.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // Get the current projects json file
+  const { data: file } = await context.octokit.repos.getContent({
+    owner: devPoolOwner,
+    repo: devPoolRepo,
+    path,
+    ref: "development",
+  });
+
+  const contentFile = JSON.parse(JSON.stringify(file));
```
pavlovcik, line 42: Object.assign is probably better if you want to copy an object.

src/handlers/pull-request/createDevPoolPR.ts
```diff
@@ -0,0 +1,83 @@
...
+  await context.octokit.pulls.create({
+    owner: devPoolOwner,
+    repo: devPoolRepo,
+    title: `Add ${repository.full_name} to repo`,
+    head: branchName,
+    base: baseRef,
+    body: "",
+  });
+
+  logger.info("New repository added to devpool");
```
pavlovcik, line 82:
```suggestion
  logger.info("New repository added to DevPool Directory");
```

src/bindings/event.ts
```diff
@@ -48,20 +48,23 @@ export const bindEvents = async (context: Context): Promise<void> => {
...
-  // Validate payload
-  const validate = ajv.compile(PayloadSchema);
-  const valid = validate(payload);
-  if (!valid) {
-    logger.info("Payload schema validation failed!!!", payload);
-    logger.warn(validate.errors!);
-    return;
-  }
+  // Skip validation for installation event
+  if (eventName !== GithubEvent.INSTALLATION_ADDED_EVENT) {
```
pavlovcik, line 52: Can you explain to me what is happening here?

src/handlers/pull-request/createDevPoolPR.ts
pavlovcik, line 1: I would like to start enforcing kebab case for file names.

pavlovcik (changes requested):

src/handlers/pull-request/createDevPoolPR.ts
seprintour: I'm not copying, i'm just parsing here..

src/handlers/pull-request/createDevPoolPR.ts
seprintour: I think its a good idea for the file names, but i don't really find it working well on the variable names

src/handlers/pull-request/createDevPoolPR.ts
pavlovcik: Shouldn't `file` already be parsed? Why are you stringifying it.

src/bindings/event.ts
pavlovcik: I think we need the wildcard processor but I'm not 100% sure @0xcodercrane

src/handlers/pull-request/createDevPoolPR.ts
pavlovcik: Just file names. Besides, it's impossible to kebab case variable names

src/bindings/event.ts
seprintour: > I think we need the wildcard processor but I'm not 100% sure @0xcodercrane

`checkBountiesToUnassign`, `collectAnalytics`, `checkWeeklyUpdate` is not something we want to run on install.. 

Its just been setup, no  bounties to unassign, no analytics and no weekly update.. so it'd often break there

src/handlers/pull-request/createDevPoolPR.ts
seprintour: > Shouldn't `file` already be parsed? Why are you serializing it?

We need to parse it, but for the stringify part.. it was only to make it parsable, the object returned from the getContent wasn't allowing a direct parse cos it was more string-like

src/types/payload.ts
```diff
@@ -16,6 +16,9 @@ export enum GithubEvent {
 
   // pull_request
   PULL_REQUEST_OPENED = "pull_request.opened",
+
+  // installation event
+  INSTALLATION_ADDED_EVENT = "installation_repositories.added",
```
rndquu, line 21: Check the following case:
1. Partner creates `repository 1`
2. Partner adds bot to the `repository 1`
3. Partner creates `repository 2`
4. Partner sets bot's repository access so that bot could use `repository 2`

Does the `installation_repositories.added` fired on step 4 ?

seprintour: Yea, its fired.. Any addition of new repository gets it fired.

Not organizational addition

src/types/payload.ts
```diff
@@ -229,6 +232,7 @@ export const PayloadSchema = Type.Object({
   repository: RepositorySchema,
   organization: Type.Optional(OrganizationSchema),
   installation: Type.Optional(InstallationSchema),
+  repositories_added: Type.Optional(Type.Array(RepositorySchema)),
```
0xcodercrane, line 235: It seems not right. the property should be a part of [`WebhookPayloadWithRepository`](https://github.com/probot/probot/blob/952f88b/src/context.ts#L177). Would you be making sure the property exists in `WebhookPayloadWithRepository` if I am missing something.

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -0,0 +1,83 @@
...
+  await context.octokit.pulls.create({
+    owner: devPoolOwner,
+    repo: devPoolRepo,
+    title: `Add ${repository.full_name} to repo`,
+    head: branchName,
+    base: baseRef,
+    body: "",
+  });
+
+  logger.info("New repository added to DevPool Directory");
```
0xcodercrane, line 82: The comment seems not right because the PR has been created on above. Lets make sure logging the PR link once it gets created so we can debug it properly from the logs

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -0,0 +1,83 @@
...
+import { Payload } from "../../types";
+
+export const createDevPoolPR = async () => {
+  const logger = getLogger();
+
+  const context = getBotContext();
+  let payload = context.payload as Payload;
+
+  let devPoolOwner = "ubiquity";
+  let devPoolRepo = "devpool";
```
0xcodercrane, line 11: you might use `const` instead of `let` if its never being changed.

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (payload.repositories_added?.length === 0) {
+    return;
+  }
+
+  let repository = payload.repositories_added![0];
```
0xcodercrane, line 17: same as above

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -0,0 +1,83 @@
+import { getBotContext, getLogger } from "../../bindings";
```
0xcodercrane, line 1: please create `index.ts` in `src/handlers/pull-request/index.ts` with the following contents. 

```typescript
export * from "./create-devpool-pr.ts"
```

0xcodercrane (changes requested):
The PR looks good totally but have some changes needed. 

Once it gets merged in the development, we want to see its working in [sandbox](https://github.com/ubiquity-bounties/ubiquibot-sandbox) repo

src/types/payload.ts
seprintour: > WebhookPayloadWithRepository

It doesn't exist

0xcodercrane: I want to keep the comment unresolved until things get clear. would you be clarifying why you added it `PayloadSchema` even if you think its in `RepositorySchema`.

seprintour: > I want to keep the comment unresolved until things get clear. would you be clarifying why you added it `PayloadSchema` even if you think its in `RepositorySchema`.

I don't understand this

seprintour: `repositories_added` contains objects that matches the `RepositorySchema`, so i am using it

seprintour: besides `repository` had a type `RepositorySchema` anyway and this new field is just like an array of repositories

src/handlers/pull-request/createDevPoolPR.ts
pavlovcik: Can you paste the exact value here? It doesn't make any sense to me because I never had to do this other than for attempting to copy an object.

pavlovcik:
> I need your input @pavlovcik before merging it  because your request seems not approved yet. 

Theres at least one conversation I tagged you in for your input you need to reply to.

src/handlers/pull-request/createDevPoolPR.ts
seprintour: > Can you paste the exact value here? It doesn't make any sense to me because I never had to do this other than for attempting to copy an object.

I used object assign it was a better way, no errors and it works.. i've made a new commit

src/bindings/event.ts
seprintour: Here @0xcodercrane

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -39,7 +39,9 @@ export const createDevPoolPR = async () => {
     ref: "development",
   });
 
-  const contentFile = JSON.parse(JSON.stringify(file));
+  const contentFile = Object.assign(file);
+
+  console.log(contentFile);
```
pavlovcik:
```suggestion

```

pavlovcik (changes requested):

src/handlers/pull-request/create-devpool-pr.ts
seprintour: I already removed it

src/handlers/pull-request/create-devpool-pr.ts
```diff
@@ -39,7 +39,9 @@ export const createDevPoolPR = async () => {
     ref: "development",
   });
 
-  const contentFile = JSON.parse(JSON.stringify(file));
+  const contentFile = Object.assign(file);
```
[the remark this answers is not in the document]
seprintour: Screenshot 2023-05-09 at 6 08 09 PM

After doing this, type error.. works without the `{}`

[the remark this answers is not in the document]
pavlovcik: Yes because you need to set its type in typescript.

```suggestion
  const contentFile = Object.assign({} as typeof file, file);
```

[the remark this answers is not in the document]
pavlovcik: You should never use `any` as a type, as it kills the point of using typescript. You should create the interface/type and assign then type your variable correctly.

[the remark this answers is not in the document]
seprintour: > You should never use `any` as a type, as it kills the point of using typescript. You should create the interface/type and assign, then assign the type to your variable correctly.

Okay, i created a type

0xcodercrane:
> > I need your input @pavlovcik before merging it  because your request seems not approved yet.
> 
> Still need [your input](https://github.com/ubiquity/bounty-bot/pull/295#discussion_r1185820969) you need to reply to.

Answered

pavlovcik (approved):

pavlovcik:
Should this pull request be reverted? @0xcodercrane @Draeieg

0xcodercrane:
Lets keep the PR and wait for @seprintour to create a hotfix pr if needed.

seprintour:
> Lets keep the PR and wait for @seprintour to create a hotfix pr if needed.

Can i handle the QA?, This works so maybe i should handle the PR and test.. remember this will need 2 repos with the staging bot installed and one of them (directory) with a little bit more permissions to create pr, edit file and contents

0xcodercrane:
@seprintour. yes you can definitely handle the QA. please follow the instructions [here](https://github.com/ubiquity/ubiquibot#how-to-qa-any-additions-to-the-bot)

PR #297, opened 2023-05-09
Feat: Access Control

### LINKED ISSUE #288: Access Control Manager
Working alongside the "roles" idea (`billing_manager`, `member`, `admin`, `recruiting_manager` etc) we should be able to configure specific permissions on the user level for bot interactions and persist those to our database. To clarify, because "roles" are built into GitHub, those are useful for controlling GUI capabilities. The following refers to our custom capabilities via the bot.
- We need to intercept the "change label" event and check WHICH user is setting WHICH label in order to enforce this permissions system.

```
Roles
    Labels
        set priority
        set time
        set price
    Commands
        set payment multiplier
```

For each user we should be able to mark each list item prefixed with "set" as a boolean (true or false.) PER repository. Organization admins should automatically get full access control.

---

The purpose of this is to enable levels of delegation (and controlling funding from permits) within the organization, per repository. 
- It makes sense for technical talent to be able to review and edit bounties to provide a time estimate. 
- It makes sense for project managers to be able to assign a priority level.
- We don't have the #128 feature implemented yet, but my vision is to allow for direct price setting as well, which should have its own access control.
- #221 for context on the payment multiplier.

devpanther:
resolves #288

### DIFF
--- a/supabase/migrations/20230509035114_create_access_table.sql
+++ b/supabase/migrations/20230509035114_create_access_table.sql
@@ -1,0 +1,10 @@
+CREATE TABLE weekly (
+    user_name character varying(255) NOT NULL PRIMARY KEY,
+    priority_access  boolean,
+    time_access boolean,
+    price_access boolean,
+    multiplier_access boolean,
+    repository text,
+    created_at text,
+    updated_at text
+);
--- a/supabase/migrations/20230509035114_create_access_table.sql
+++ b/supabase/migrations/20230509035114_create_access_table.sql
@@ -1,10 +1,10 @@
 CREATE TABLE access (
     user_name character varying(255) NOT NULL PRIMARY KEY,
+    repository text,
     priority_access  boolean,
     time_access boolean,
     price_access boolean,
     multiplier_access boolean,
-    repository text,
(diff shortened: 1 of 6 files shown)


supabase/migrations/20230509035114_create_access_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE access (
+    user_name character varying(255) NOT NULL PRIMARY KEY,
+    repository text,
+    priority_access  boolean,
+    time_access boolean,
+    price_access boolean,
+    multiplier_access boolean,
+    created_at text,
+    updated_at text
```
0xcodercrane, line 9: Lets make the type of both `timestamp`. Please take a look for the reference.  https://github.com/ubiquity/bounty-bot/blob/development/supabase/migrations/20230207013753_update_types.sql

0xcodercrane (changes requested):
I think it has not been requested for a review but wanted to make sure you're in the right direction in terms of migration.  please ignore me if you're already updating it as I requested locally.

0xcodercrane (approved):

Issue #285, opened 2023-05-03
Make permit2 compatible with gnosis chain


rndquu:
Depends on https://github.com/ubiquity/bounty-bot/issues/284

We should update our [permit2 script](https://github.com/ubiquity/bounty-bot/blob/development/src/helpers/permit.ts) so that it could support multiple networks (specifically gnosis chain and ethereum mainnet).

What should be done:
1. Use permit2 address depending on the bot's config `chainId` param [here](https://github.com/ubiquity/bounty-bot/blob/development/src/helpers/permit.ts#L6)
2. Use payment token depending on the bot's config `chainId`param [here](https://github.com/ubiquity/bounty-bot/blob/development/src/helpers/permit.ts#L27). For ethereum mainnet use `DAI`, for gnosis chain use `WXDAI`.

PR #316, opened 2023-05-18
feat: make permit2 compatible with gnosis

rndquu:
resolves https://github.com/ubiquity/ubiquibot/issues/285


src/configs/shared.ts
```diff
@@ -25,7 +25,6 @@ export const DEFAULT_DISQUALIFY_TIME = "7 days"; // 7 days
 export const DEFAULT_CHAIN_ID = 1; // ethereum
 export const DEFAULT_RPC_ENDPOINT = "https://rpc-bot.ubq.fi/v1/mainnet";
 
-export const DEFAULT_PAYMENT_TOKEN = "0x6B175474E89094C44Da98b954EedeAC495271d0F";
```
pavlovcik, line 28: Shouldn't we replace this with another default instead of removing?

src/helpers/payout.ts
```diff
@@ -0,0 +1,45 @@
...
+ * 1. Gelato network should support the added chain (https://docs.gelato.network/developer-services/relay/api#oracles)
+ * 2. Gelato network should support the added payment token (https://docs.gelato.network/developer-services/relay/api#oracles-chainid-paymenttokens)
+ */
+
+import { Static } from "@sinclair/typebox";
+import { DEFAULT_RPC_ENDPOINT } from "../configs";
+import { PayoutConfigSchema } from "../types";
+
+// chain ids
+const CHAIN_GNOSIS = 100;
```
pavlovcik, line 19: I just realized that the interesting chains to integrate with for me personally:

Mainnet: 1
Optimism: 10
Gnosis Chain: 100

src/configs/shared.ts
rndquu: We don't need the `DEFAULT_PAYMENT_TOKEN` variable anymore so it was deleted. Now we have a separate payment token for [each chain](https://github.com/ubiquity/ubiquibot/pull/316/files#diff-7a573dc385193e6a51c6e9a66e89d361cb981e4c10d27e77d398aa669d9f980dR22).

src/helpers/payout.ts
rndquu: Adding a new chain is basically 6 lines of code:
```js
const CHAIN_OPTIMISM = 10;
const PAYMENT_TOKEN_CHAIN_OPTIMISM = "0x00"; // some token supported by gelato
if (chainId === CHAIN_OPTIMISM) {
    payoutConfig.rpc = "https://rpc.optimism.com";
    payoutConfig.paymentToken = PAYMENT_TOKEN_CHAIN_OPTIMISP;
}
```

Then a partner would need to:
1. Update `chain-id` in the config
2. [Approve permit2 contract to spend tokens](https://dao.ubq.fi/ubiquibot-setup#42f485283d0d463c8a2c8289ecfb8248)

pavlovcik (approved):

0xcodercrane:
Here's another thing I am thinking for the gnosis chain. so basically users want to DAI instead of WXDAI on gnosis because it is the native token. Its not a blocker but we might need to think about that on users end

rndquu:
> Here's another thing I am thinking for the gnosis chain. so basically users want to DAI instead of WXDAI on gnosis because it is the native token. Its not a blocker but we might need to think about that on users end

There is no DAI in gnosis chain per se. When you bridge DAI from ethereum mainnet to gnosis chain those bridged DAI are converted to XDAI (`ERC20 token DAI in mainnet = gas token XDAI in gnosis`) which is a native token in gnosis chain used for gas payments. So basically supporting XDAI (native gas token) means that we should also support payments in ETH which means 2 codebases supporting payments both in native (gas) tokens and ERC20 tokens.

0xcodercrane:
It could be done by creating a permit contract with a wrapper feature on gnosis. its simple but yeah we can do it later

pavlovcik:
xDAI isn't an ERC20

0xcodercrane:
> xDAI isn't an ERC20

Right but WXDAI is an ERC20 lol

pavlovcik:
Yes, I know. There's some miscommunication but I'll delegate my vote to rndquu because I believe that they did quite a bit of research on the topic.

PR #318, opened 2023-05-19
fix: fix permit2 payouts

rndquu:
At the moment permit2 payouts are broken. 

When issue is closed and permit2 URL is about to be generated the bot's config [here](https://github.com/ubiquity/ubiquibot/blob/development/src/utils/private.ts#L84) is an empty object because it is not yet initialized. 

That is why we read sodium's private key from the env variable instead of the bot's config.


pavlovcik:
I think this review is @Steveantor territory

Steveantor (approved):

Issue #310, opened 2023-05-16
Re-Enable GitHub Action Runtime Environment


pavlovcik:
We should enable support from running the bot natively off of GitHub Actions again, but this time using `npx tsx index.ts` in order to eliminate the complexities around compilation, and then updating the `dist/index.js` file.

The bot should use GitHub Actions as the runtime environment for forks automatically for testing purposes. 

For production we should continue using a server less backend (we are currently using Netlify) this is because, from what I understand, the only way to easily add the bot from the GitHub Marketplace and immediately start using is if it is using an external backend. Otherwise, keeping all of the infrastructure on GitHub is preferred.

0xcodercrane:
Hmm yes we can enable it actually.

0xcodercrane:
making a progress on it

pavlovcik:
But the last update was last week. Perhaps it would be in everybody's best interest if you only were assigned and focused on one task at a time?

0xcodercrane:
resuming the work here

pavlovcik:
Hey we should discuss this before you close and claim it. I expected this to be a huge project to make it run on the netlify backend for production, but then to only run on GitHub Actions for the bounty hunter forks. Can you verify that is the status?

0xcodercrane:
TL;DR; the status is fine. so we have 2 interfaces; [netlify function](https://github.com/ubiquity/ubiquibot/blob/development/.netlify/functions/webhooks/webhooks.ts) and [github adapter action](https://github.com/ubiquity/ubiquibot/blob/development/adapters/index.ts). Both are being used separately

0xcodercrane:
The pr is still being tested locally but i think I made a mistake in the merge. reopen it

pavlovcik:
What's the status here?

PR #320, opened 2023-05-22
Reenable github action

0xcodercrane:


.github/workflows/build.yml
```diff
@@ -20,3 +20,23 @@ jobs:
...
+        if: github.ref == 'refs/heads/development'
+        run: |
+          git config --local user.email "UbiquiBot[dev]@github.com"
+          git config --local user.name "UbiquiBot[dev]"
+          yarn build:ci
+          git add -A
+          git diff-index --quiet HEAD || git commit -am "build: update executable" --no-verify
+          git push git@github.com:${{github.repository}}.git ${{github.ref}}
+        env:
+          GITHUB_TOKEN: ${{ secrets.UBIQUITY_BOUNTY_BOT_DEV_PRIVATE_KEY_OPENSSH }}
```
pavlovcik, line 42: This is all unnecessary because we can run the bot using `npx tsx` so there's no need to commit the `dist/index.js` file anymore. This is in the spec https://github.com/ubiquity/ubiquibot/issues/310 @0xcodercrane

pavlovcik (changes requested):

PR #335, opened 2023-05-27
No dependency on yarn version

0xcodercrane:


.yarn/install-state.gz
pavlovcik, line 1: I assume this is a netlify workaround because you definitely should not be including this type of file in the repository under normal circumstances.

PR #338, opened 2023-05-29
hotfix of permit check

0xcodercrane:


pavlovcik (commented):
I linked another bounty. 

To clarify, I believe that the order of the checks matter so that the bot can provide the most relevant warning/error message in the comment for why the permit generation was aborted. 

In this reference scenario, the bot should first check if the issue was closed as completed. If it is not closed as completed, then the bot should cite that as the reason the permit was not generated above any other reason. 

I suppose we should come up with a list of all the reasons and put them in a single ordered list in a bounty in order to provide the most relevant warning/error message.

PR #343, opened 2023-05-29
Create automate-payments.js

drew7wonders:
Add a new feature to the Ubiquibot that allows contributors to claim, send, or manage their compensation more efficiently.


pavlovcik:
You need to link the original issue

PR #344, opened 2023-05-30
Fix duplicate posting of permit url

0xcodercrane:


src/handlers/comment/handlers/payout.ts
```diff
@@ -27,7 +27,7 @@ export const payout = async (body: string) => {
   }
 
   const IssueComments = await getAllIssueComments(issue.number);
-  if (IssueComments === null) {
+  if (IssueComments.length === 0) {
```
pavlovcik, line 30: Should it check for `null` and empty array? Why was it `null` before?

src/handlers/payout/index.ts
```diff
@@ -72,6 +72,13 @@ export const handleIssueClosed = async () => {
   const shortenRecipient = shortenEthAddress(recipient, `[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]`.length);
   logger.info(`Posting a payout url to the issue, url: ${payoutUrl}`);
   const comment = `### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
+  const comments = await getAllIssueComments(issue.number);
+  const commentContents = comments.map((i) => i.body);
+  const exist = commentContents.find((content) => content.includes(comment));
+  if (exist) {
+    logger.info(`Skip to generate a permit url because it has been already posted`);
+    return `Skip to generate a permit url because it has been already posted`;
```
pavlovcik, line 80:
```suggestion
    return `Permit generation skipped because it was already posted to this issue.`;
```

src/handlers/comment/handlers/payout.ts
0xcodercrane: I am not sure why it was `null` someone made it. but `getAllIssueComments` returns an array of comments so no need to make the return object `null`,  empty array is better to handle.

pavlovcik: According to git blame it was @Steveantor. 

Can you explain why you put `null` here?

pavlovcik:
```suggestion
  if (IssueComments?.length === 0) {
```

This also solves for both `null` and empty array.

src/helpers/issue.ts
```diff
@@ -99,11 +99,11 @@ export const getCommentsOfIssue = async (issue_number: number): Promise<Comment[
   return result;
 };
 
-export const getAllIssueComments = async (issue_number: number): Promise<Comment[] | null> => {
+export const getAllIssueComments = async (issue_number: number): Promise<Comment[]> => {
   const context = getBotContext();
   const payload = context.payload as Payload;
 
-  let result: Comment[] | null = [];
```
pavlovcik, line 106: These also seem obviously intentional. I believe that the API sometimes returns `null` @0xcodercrane

0xcodercrane: Its getting handle properly here.

0xcodercrane:
Why did I remove `null` from the following codebase? 
`export const getAllIssueComments = async (issue_number: number): Promise<Comment[] | null> => {}`

There's no mandatory requests in code style so its free to make it as you want. but I removed `null` from the following benefits.

1/ Simplify the function body itself.
```typescript
- export const getAllIssueComments = async (issue_number: number): Promise<Comment[] | null> => {
+ export const getAllIssueComments = async (issue_number: number): Promise<Comment[]> => {

- let result: Comment[] | null = [];
+ let result: Comment[] = [];

catch (e: unknown) {
-   result = unknown
    shouldFetch = false;
}
```

2/ make its usage simpler
I've added both lines to use it. 
```
  const commentContents = comments.map((i) => i.body);
  const exist = commentContents.find((content) => content.includes(comment));
```
If `getAllIssueComments` still returns `null`, I had to add a null check for both lines.

src/helpers/issue.ts
0xcodercrane: https://github.com/ubiquity/ubiquibot/pull/344#issuecomment-1567720041

src/handlers/comment/handlers/payout.ts
0xcodercrane: https://github.com/ubiquity/ubiquibot/pull/344#issuecomment-1567720041

pavlovcik (approved):

PR #345, opened 2023-05-30
Continuous Deployment

0xcodercrane:


pavlovcik (approved):

src/handlers/payout/index.ts
```diff
@@ -17,7 +17,7 @@ export const handleIssueClosed = async () => {
   if (!issue) return;
 
   if (issue.state_reason !== StateReason.COMPLETED) {
-    logger.info("No need to generate a permit url for an uncompleted issue");
+    logger.info("Permit generation skipped because the issue was not closed as completed");
     return "Permit generation skipped because the issue was not closed as completed";
```
pavlovcik: You should just write the string in a variable and use the variable. Anyways its a nit-pick

0xcodercrane: Definitely. we have lots of nit-picks like this so I am planning to refactor all of them in an another issue

Issue #342, opened 2023-05-29
Don't Add Range Labels


pavlovcik:
This pull request https://github.com/ubiquity/ubiquibot/pull/249 doesn't solve the problem I just tested removing the range label on https://github.com/ubiquity/research/issues/11 and the bot added it back. Please remove the "range labels" feature as they cause more confusion than add value. 

### Context

- The bot will add a "range label" e.g. `Price: 300+ USD` when only one of the two required labels are added (either `Time` or `Priority`)
- The intent was to set the expectations for the bounty hunter that the minimum price is likely to be what is specified with the range label. 
- Unfortunately we do not actually generate permits for range labels. Permits require both the `Time` and `Priority` to be set. This has caused issues in the past with bounty hunters not getting payment permits generated and also with changing prices. Please remove this feature.

PR #348, opened 2023-05-30
Don't add range labels

0xcodercrane:
Resolves https://github.com/ubiquity/ubiquibot/issues/342


rndquu (approved):
We should also remove `target` from all other configs when this PR is deployed to production

0xcodercrane:
> We should also remove `target` from all other configs when this PR is deployed to production
> 
> Added #349

It won't break even if we keep it but should remove it because its no longer needed

pavlovcik (approved):

PR #352, opened 2023-05-31
Base rate update watcher

### LINKED ISSUE #341: Repository Price Labels Should Update When `base-rate` Updated
It has caused confusion where the `Price` label does not match the generated permit claim amount. This is because the bot currently does not monitor changes to the `base-rate` and then check the `Price` labels to see if they are accurate. People are sensitive about money so it is worth putting extra attention to detail on this one. 

I'm not sure of the exact best approach here (ideally this logic runs if the `ubiquibot-config.yml` was detected to have been updated.)

I believe this is the most efficient approach from a API rate limit perspective: 

1. Start when the `ubiquibot-config.yml` has been updated
2. Take note of the old base rate
3. Take note of the new base rate
4. Query the repositories labels API for a list of all of the labels
5. Directly edit the `Price` labels noting the difference between the old and new base rates.

Context: https://github.com/ubiquity/ubiquibot/issues/329#issuecomment-1565436831

devpanther:
resolves #341


0xcodercrane:
Lets make sure the netlify build succeeds before starting a review @devpanther

devpanther:
> Lets make sure the netlify build succeeds before starting a review @devpanther

That was automatic, did not start one and besides the netlify error is not from my changes

src/helpers/label.ts
```diff
@@ -53,3 +53,47 @@ export const getLabel = async (name: string): Promise<boolean> => {
...
+  try {
+    for (const label of labels) {
+      if (label.name.startsWith("Price: ")) {
+        const labelParts = label.name.split(": ");
+        const prefix = labelParts[0];
+        const valueAndSuffix = labelParts[1].split(" ");
+
+        let labelValue = parseFloat(valueAndSuffix[0]);
+
+        let updatedLabelValue =
```
rndquu, line 69: @devpanther perhaps it is simpler to calculate label prices from scratch without calculating the difference

rndquu:
> > That was automatic, did not request and besides the netlify error is not from my changes
> 
> That is my fault because I updated https://github.com/ubiquity/ubiquibot/blob/development/.github/CODEOWNERS which apparently automatically requests reviews, even on drafts. Perhaps there is a way to disable that, or, we will need to remove this feature entirely.
> 
> I hope that somebody else can handle this research because I'm underwater on some other matters for the rest of the week.

From the [docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners):
```
Code owners are not automatically requested to review draft pull requests
```

What happened:
1. @devpanther opened a ready to review PR
2. I converted the PR to draft as not ready

TLDR; @devpanther pls for future PRs open a draft PR first

src/helpers/label.ts
devpanther: > @devpanther perhaps it is simpler to calculate label prices from scratch without calculating the difference

how do you mean from scratch

rndquu: you can use an [existing function](https://github.com/ubiquity/ubiquibot/blob/development/src/handlers/shared/pricing.ts#L3) to calculate an updated price

devpanther: > you can use an [existing function](https://github.com/ubiquity/ubiquibot/blob/development/src/handlers/shared/pricing.ts#L3) to calculate an updated price

Doesn't it need an updated config to give the correct prices?

rndquu: Yes it does. As far as I understand we already have an updated config on push event so we can recalculate price labels based on it.

So on push event we:
1. check whether `base-multiplier` has been updated
2. at this point we have access to the updated config so we can recalculate price labels

devpanther: > Yes it does. As far as I understand we already have an updated config on push event so we can recalculate price labels based on it.
> 
> So on push event we:
> 
> 1. check whether `base-multiplier` has been updated
> 2. at this point we have access to the updated config so we can recalculate price labels

Yea, I saw the problem.. i was still using the camel case config

devpanther:
Ready for review @0xcodercrane @rndquu

src/bindings/event.ts
```diff
@@ -35,11 +37,15 @@ export const bindEvents = async (context: Context): Promise<void> => {
   if (!logger) {
     return;
   }
+
   logger.info(`Config loaded! config: ${JSON.stringify({ price: botConfig.price, unassign: botConfig.unassign, mode: botConfig.mode, log: botConfig.log })}`);
   const allowedEvents = Object.values(GithubEvent) as string[];
-  const eventName = `${name}.${payload.action}`;
+  const eventName = payload.action ? `${name}.${payload.action}` : name; // some events wont have actions as this grows
```
0xcodercrane, line 43: looks like a nice catch!

src/helpers/label.ts
```diff
@@ -53,3 +56,65 @@ export const getLabel = async (name: string): Promise<boolean> => {
 
   return false;
 };
+
+// Function to update labels based on the base rate difference
+export async function updateLabelsFromBaseRate(owner: string, repo: string, context: Context, labels: Label[], previousBaseRate: number) {
```
0xcodercrane, line 61: Here's a nitpick in terms of coding style. 
```suggestion
export const updateLabelsFromBaseRate = async (owner: string, repo: string, context: Context, labels: Label[], previousBaseRate: number) =>{
```

src/helpers/label.ts
```diff
@@ -1,8 +1,11 @@
...
+import { Context } from "probot";
+import { getBotConfig, getBotContext, getLogger } from "../bindings";
 import { COLORS } from "../configs";
-import { Payload } from "../types";
+import { calculateBountyPrice } from "../handlers";
+import { Label, Payload } from "../types";
+import { deleteLabel } from "./issue";
 
-export const listLabelsForRepo = async (per_page?: number, page?: number): Promise<string[]> => {
+export const listLabelsForRepo = async (per_page?: number, page?: number, all?: boolean): Promise<Label[] | string[]> => {
```
0xcodercrane, line 8: could we make it returning only `Label[]` like 
```suggestion
export const listLabelsForRepo = async (per_page?: number, page?: number): Promise<Label[]> => {
```
There would be a few parts to use `string[]` but we can get them easily from `Label[]`. we can reduce the complexity in use of the `listLabelsForRepo` function.

0xcodercrane (changes requested):
Looks good. there are a couple of requests.

devpanther:
@0xcodercrane requests resolved

devpanther:
Mistakenly closed with comment

0xcodercrane (approved):

0xcodercrane:
overall looks good to me. waiting for @pavlovcik 's input.

src/bindings/event.ts
pavlovcik: @0xcodercrane should we make an issue for this? I do not have enough context on this issue to create it myself if needed.

pavlovcik:
Unfortunately this is a huge pull request and I am unable to test it. I perused through and everything checks out but I do not fully understand all of the changes. I delegate my review to @0xcodercrane

src/handlers/push/update-base.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  // fetch all labels
+  const repoLabels = await listLabelsForRepo();
+
+  if (repoLabels.length === 0) {
+    logger.debug("No labels on this repo");
+    return;
+  }
+
+  updateLabelsFromBaseRate(owner, repo, context, repoLabels as Label[], previousBaseRate);
```
pavlovcik, line 35:
```suggestion
 await updateLabelsFromBaseRate(owner, repo, context, repoLabels as Label[], previousBaseRate);
```

I noticed that this is an async function. Shouldn't this be `await`ed?

src/helpers/file.ts
```diff
@@ -0,0 +1,61 @@
...
+          repo,
+          file_sha: previousFile.sha!,
+        });
+        return previousFileContent.data.content;
+      }
+    }
+    return "";
+  } catch (error) {
+    logger.debug("Error retrieving previous file content:", error);
+    return "";
```
pavlovcik, line 59: Returning empty strings doesn't seem right. Shouldn't we be returning useful messages and/or errors for the bot to comment?

src/helpers/label.ts
```diff
@@ -14,7 +17,7 @@ export const listLabelsForRepo = async (per_page?: number, page?: number): Promi
   });
 
   if (res.status === 200) {
-    return res.data.map((i) => i.name);
+    return res.data;
```
pavlovcik, line 20: This seems like a big change from the API where it apparently used to return a string, and then now its returning a `Label` object. Is this correct? Why is this only being changed now? I feel like the code in this revision and the last revision might not be compatible?

pavlovcik (commented):
I left some low-context comments!

src/helpers/label.ts
devpanther: They're compatible, just one file made use of this and now it's made compatible for easy use of the function

src/helpers/file.ts
devpanther: Only comments return response/errors, this is an event run function.. It's clearly logged

I have no use of the error when reading a file, once it returns empty string. The process does not continue any way

src/handlers/push/update-base.ts
devpanther: Yea, it can but this is the last function so it doesn't matter if there's an await or not

pavlovcik: The runtime can terminate before the response arrives unless you `await`.

pavlovcik:
@0xcodercrane check the comments and merge this in if everything checks out for you.

src/handlers/push/update-base.ts
0xcodercrane: oh good catch @pavlovcik. it needs to be `await`

devpanther: > oh good catch @pavlovcik. it needs to be `await`

Awaited it

devpanther: > oh good catch @pavlovcik. it needs to be `await`. Even if it is the last function, it should be `await` because the bot runs as a lambda function so needs to wait until all the functions get through. If not, it just terminates without running. lets accept the request changes @devpanther 

Yea @0xcodercrane, I've Awaited it..

devpanther: > please post the commit link and mark it resolved

https://github.com/ubiquity/ubiquibot/pull/352/commits/de72859b8e7c27a82f97c6271dec55c8654b9142

src/helpers/file.ts
0xcodercrane: we can go with the current implementation but I guess it would be more ideal to return `undefined` instead of empty string. just a comment here

devpanther:
All reviews resolved @0xcodercrane

src/helpers/file.ts
devpanther: > we can go with the current implementation but I guess it would be more ideal to return `undefined` instead of empty string. just a comment here

Undefined might just end up breaking the Buffer read

PR #360, opened 2023-06-01
chore: logging

0xcodercrane:


src/handlers/payout/test.json
pavlovcik, line 1: This probably should not have been committed to the repository

0xcodercrane: It shouldn't

pavlovcik (dismissed):

PR #363, opened 2023-06-02
Autopay command

### LINKED ISSUE #290: `/autopay` Command
1. Add a new command `/autopay true|false` that enables/disables permit2 URL payouts for a particular issue.
2. The bot should respond with a success message
3. On issue close and `autopay == false` the bot should display a warning message of why a permit URL is not generated
4. Update help menu accordingly

Notice that there is the `auto-pay-mode` bot's param which enables/disables automatic payments for the whole org/repo. The `/autopay true|false` command should take precedence on the particular issue over the bot's config.

seprintour:
resolves #290

### DIFF
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -133,6 +133,11 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
   return result;
 };
 
+export const issueAutopayCheck = async (issue_number: number) => {
+  let comments = await getAllIssueComments(issue_number);
+  console.log(comments);
+};
+
 export const removeAssignees = async (issue_number: number, assignees: string[]): Promise<void> => {
   const context = getBotContext();
   const logger = getLogger();
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -141,7 +141,7 @@ export const issueAutopayCheck = async (issue_number: number) => {
 
   for (const item of comments) {
     const { body, user } = item;
-    if (user?.login === "ubiquibot[bot]" && user?.type === "Bot") {
+   if (user?.type === "Bot") {
       const match = body.match(regex);
 
       if (match && match[1]) {
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -134,8 +134,30 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
 };
 
 export const issueAutopayCheck = async (issue_number: number) => {
+  const context = getBotContext();
+  console.log(context);
   let comments = await getAllIssueComments(issue_number);
-  console.log(comments);
+  let lastAutopayStatus = null;
+
+  const regex = /Autopay is (disabled|enabled)/;
+
+  for (const item of comments) {
+    const { body, user } = item;
+    if (user?.login === "ubiquibot[bot]" && user?.type === "Bot") {
+      const match = body.match(regex);
+
+      if (match && match[1]) {
+        const status = match[1];
+        if (status === "enabled") {
+          lastAutopayStatus = true;
+        } else if (status === "disabled") {
+          lastAutopayStatus = false;
+        }
+      }
+    }
+  }
+
+  return lastAutopayStatus === null ? true : lastAutopayStatus;
 };
 
 export const removeAssignees = async (issue_number: number, assignees: string[]): Promise<void> => {
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -134,8 +134,6 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
 };
 
 export const issueAutopayCheck = async (issue_number: number) => {
-  const context = getBotContext();
-  console.log(context);
   let comments = await getAllIssueComments(issue_number);
   let lastAutopayStatus = null;
 
--- a/src/handlers/comment/commands.ts
+++ b/src/handlers/comment/commands.ts
@@ -5,6 +5,7 @@ export enum IssueCommentCommands {
   WALLET = "/wallet", // register wallet address
   PAYOUT = "/payout", // request permit payout
   MULTIPLIER = "/multiplier", // set bounty multiplier (for treasury)
+  AUTOPAY = "/autopay",
 
   // Access Controls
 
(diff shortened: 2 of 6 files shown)


seprintour:
@0xcodercrane ready for review

src/helpers/issue.ts
```diff
@@ -133,6 +133,31 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
...
 
+export const issueAutopayCheck = async (issue_number: number) => {
+  let comments = await getAllIssueComments(issue_number);
+  let lastAutopayStatus = null;
+
+  const regex = /Autopay is (disabled|enabled)/;
+
+  for (const item of comments) {
+    const { body, user } = item;
+    if (user?.login === "ubiquibot[bot]" && user?.type === "Bot") {
```
0xcodercrane, line 144: if (user?.type === "Bot") {
```suggestion
    if (user?.login === "ubiquibot[bot]" && user?.type === "Bot") {
```

pavlovcik: Your suggestion doesn't seem right. It means that it will run this check on any bot comment, right?

seprintour: > Your suggestion doesn't seem right. It means that it will run this check on any bot comment, right?

I added that so it only comes fromubiquibot, i feared someone could exploit by commenting as a bot.
 
But bots are github apps.. and there is no they are exploiting any of the github apps installed just to comment autopay

0xcodercrane (approved):

src/helpers/issue.ts
pavlovcik: >  and there is no they are exploiting any of the github apps installed just to comment autopay

I don't understand your statement.

seprintour: > > and there is no they are exploiting any of the github apps installed just to comment autopay
> 
> I don't understand your statement.

It means to comment as a bot, you need a github app installed on the repo and that is very unlikely..

pavlovcik: Installing any GitHub App on the repo is very easy to do I don't understand what you're talking about.

seprintour: But can only be installed by admin, so that neutralises the risks..

pavlovcik: @0xcodercrane I don't understand this issue what do you think?

rndquu (approved):

seprintour:
Reviews @0xcodercrane ?

src/helpers/issue.ts
```diff
@@ -133,6 +133,31 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
   return result;
 };
 
+export const issueAutopayCheck = async (issue_number: number) => {
+  let comments = await getAllIssueComments(issue_number);
+  let lastAutopayStatus = null;
+
+  const regex = /Autopay is (disabled|enabled)/;
```
pavlovcik, line 140: This seems brittle. Perhaps we should have a more robust system for bot comments as a "database" or at least we should share the string between the read and write function so that we will not have a situation where a developer only updates one of the two, breaking this system.

seprintour: @pavlovcik i did not fully get your point above, are you saying we should store the autopay states to a DB instead?

0xcodercrane:
@pavlovcik It has already been reverted [here](https://github.com/ubiquity/ubiquibot/pull/372).

seprintour:
@pavlovcik It was tested here https://github.com/seprintour/betterbuy/issues/1

seprintour:
I don't know why it throws handler error on lambda till now.. it works perfectly on the repo

And yes, the two regex are different

seprintour:
1. [Set to false here](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579316061)
2. [Saved as disabled](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579316124)
Result. [Autopay disabled](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579316676)

3. [Set to true here](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579315454)
4. [Saved as enabled](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579315524)
Result. [Asking for wallet](https://github.com/seprintour/betterbuy/issues/1#issuecomment-1579315753) (passed autopay)

seprintour:
> Yeah I just realize those two regexes aren't even compatible so this code shouldn't work

So this is wrong, one is a comment handler and one is just a way to save the state on the comment while there is a function that watches for (disabled|enabled)

pavlovcik:
Should this pull request be reverted? @0xcodercrane @Draeieg

0xcodercrane:
I suggest keeping the PR merged and creating a hotfix PR if needed. 
@seprintour you can also give us proof of work by following [this](https://github.com/ubiquity/ubiquibot#how-to-qa-any-additions-to-the-bot). @Draeieg  will verify the work here

pavlovcik:
If we do reverts then the bot can track it in the future and XP can be deducted, for example.

seprintour:
> @seprintour
> 
> > at least we should share the string between the read and write function
> 
> * [Autopay command #363 (comment)](https://github.com/ubiquity/ubiquibot/pull/363#discussion_r1219324880)
> * [Autopay command #363 (comment)](https://github.com/ubiquity/ubiquibot/pull/363#discussion_r1219325638)
> 
> Export and then import the regex and use the same regex in both to make the code less brittle.
> 
> Yeah I just realize those two regexes aren't even compatible so this code shouldn't work. One is searching for `true` or `false` and the other is searching for `disabled` or `enabled`. How were you able to test this?
> 
> @0xcodercrane you might need to revert.
> 
> @Draeieg it would be helpful if you can also test this and confirm.

I don't think you should call for a revert without the author explaining the code.. the code works, i would not push a code that doesn't work and besides i attached a proof on my repo

seprintour:
As it has already been reverted by just merely looking at the code, how would it make it to QA for @Draeieg ?

I am 100% certain this code works so i am going to leave it at this.. i have several PR's that were open without proper tests

This https://github.com/ubiquity/ubiquibot/issues/278 is on of them, and apparently from my investigation it has been working

PR #364, opened 2023-06-03
feat: added 'add review' comment

### LINKED ISSUE #307: "Please Leave a Review"
Upon payment permit generation the bot should say something like "If you enjoy the DevPool experience, please follow [Ubiquity on GitHub](https://github.com/ubiquity) and star [this repo](https://github.com/ubiquity/devpool-directory) to show your support. It helps a lot!"

wannacfuture:
# Description

- I added 'Please add review' comment after payment permit generation.

closes #307

### DIFF
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -42,6 +42,10 @@ export const issueClosedCallback = async (): Promise<void> => {
   const issue = (_payload as Payload).issue;
   try {
     const comment = await handleIssueClosed();
+    await addCommentToIssue(
+      `If you enjoy the DevPool experience, please follow <a href="https://github.com/ubiquity">Ubiquity on GitHub</a> and star <a href="https://github.com/ubiquity/devpool-directory">this repo</a> to show your support. It helps a lot!`,
+      issue!.number
+    );
     return await addCommentToIssue(comment!, issue!.number);
   } catch (err: any) {
     return await addCommentToIssue("Error: " + err.message, issue!.number);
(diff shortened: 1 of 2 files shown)


pavlovcik:
> I added 'Please add review' comment before payment permit generation.

What was your reasoning for putting it before? I do not have a strong opinion on this matter, but I feel that I would put it at the very end to make sure that the comment is not missed.

wannacfuture:
> > I added 'Please add review' comment before payment permit generation.
> 
> What was your reasoning for putting it before? I do not have a strong opinion on this matter, but I feel that I would put it at the very end to make sure that the comment is not missed.

Yeah, seems u're right. Fixed.

src/handlers/comment/handlers/index.ts
```diff
@@ -42,7 +42,11 @@ export const issueClosedCallback = async (): Promise<void> => {
   const issue = (_payload as Payload).issue;
   try {
     const comment = await handleIssueClosed();
-    return await addCommentToIssue(comment!, issue!.number);
+    await addCommentToIssue(comment!, issue!.number);
+    await addCommentToIssue(
```
pavlovcik, line 46: It seems that these should be consolidated into a single comment. What is :45 commenting exactly?

wannacfuture: :45 is the returning comment about payment permit when the user closed issue. And if we consolidate them into one, it will be hard to impact review comment. Is that OK?
I mean, the payment permit result and "If you enjoy the experience ..." comment will be one comment.

pavlovcik (approved):

src/handlers/comment/handlers/index.ts
pavlovcik: Do you have an example on your fork so I could see and compare?

PR #365, opened 2023-06-04
Feat/default label

### LINKED ISSUE #325: Adding Default Labels on New Issue
A nice config to have would be to allow repository maintainers to specify the default labels that the bot should automatically add for any newly created issue if there are no labels. 

For example:

When I create new issues, I do several things before pressing the "Submit new issue" button generally speaking: 1. Write the issue title 2. Write the issue body 3. Add labels.

**"3." Means that it is possible to create a new issue with no labels (or with the proper `Priority: ` and `Time: ` labels.)**

It would enhance the bounty issuing experience if we can set a repository default in `ubiquibot-config.yml` to automatically add specific labels on issues opened without any labels. In this instance, I would be interested to have the bot automatically add `Priority: 0 (Normal)` and `Time: <1 Hour` (or `Time: <1 Day`) so that I can rapid fire issuing bounties.

wannacfuture:
# Description

- added default label on new issue

closes #325

### DIFF
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -1,3 +1,4 @@
+import { getBotConfig } from "../../../bindings";
 import { Payload, UserCommands } from "../../../types";
 import { IssueCommentCommands } from "../commands";
 import { assign } from "./assign";
@@ -7,7 +8,7 @@ import { unassign } from "./unassign";
 import { registerWallet } from "./wallet";
 import { setAccess } from "./set-access";
 import { multiplier } from "./multiplier";
-import { addCommentToIssue } from "../../../helpers";
+import { addCommentToIssue, createLabel } from "../../../helpers";
 import { getBotContext } from "../../../bindings";
 import { handleIssueClosed } from "../../payout";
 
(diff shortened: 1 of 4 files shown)


src/handlers/comment/handlers/index.ts
```diff
@@ -42,12 +43,37 @@ export const issueClosedCallback = async (): Promise<void> => {
...
+  const { payload: _payload } = getBotContext();
+  const config = getBotConfig();
+  const issue = (_payload as Payload).issue;
+  const labels = issue!.labels;
+  try {
+    const timeLabels = config.price.timeLabels.filter((item) => labels.map((i) => i.name).includes(item.name));
+    const priorityLabels = config.price.priorityLabels.filter((item) => labels.map((i) => i.name).includes(item.name));
+
+    if (timeLabels.length === 0) await createLabel("Time: <1 Hour");
+    if (priorityLabels.length === 0) await createLabel("Priority: 0 (Normal)");
```
pavlovcik, line 70: The default labels here should load from our config actually. If there is nothing in the config then this should not do anything.

pavlovcik (changes requested):
This pull request looks really good! Thank you for your contribution. There is one change we need though. Thank you.

src/handlers/comment/handlers/index.ts
wannacfuture: yeah, there is no problem. But what is the variable name? I mean, I can I get the data?

For example, config.default_timeLabel, config.default_priorityLabel???

wannacfuture: Yeap! I've fixed it by picking up the lowest price label.

pavlovcik (approved):
The code looks great to me. Thank you. It isn't easy for me to test so I'll wait for @0xcodercrane to sign off on this!

0xcodercrane (approved):
Looks good

PR #366, opened 2023-06-04
feat: make default config

### LINKED ISSUE #327: Add default yml config
More info [here](https://github.com/ubiquity/pay.ubq.fi/issues/68)

What should be done:
1. Add a default bot's `yml`config so that the [config generation page](https://pay.ubq.fi/) could fetch it via github API
2. Assign the default values [here](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L15) from the default config `yml` file instead of the hardcoded param values

Venoox:
resolves #327


src/utils/helpers.ts
pavlovcik, line 1: Returning null everywhere doesn't seem right.

src/utils/private.ts
pavlovcik, line 1: These fallbacks also don't seem right.

default_config.yml
pavlovcik, line 1: Don't we have a standard naming convention as `ubiquibot-config.yml`?

pavlovcik (commented):
There are several things that do not seem right. I will need to delegate my review to my colleagues.

default_config.yml
rndquu: @Venoox  Yes, we've got a kebab case naming convention for files so could you rename `default_config.yml` to smth like `ubiquibot-config-default.yml` ?

src/utils/helpers.ts
```diff
@@ -1,93 +1,92 @@
-import { DEFAULT_CHAIN_ID, DefaultPriceConfig } from "../configs";
```
rndquu, line 1: If `DefaultPriceConfig` is not used anywhere then we are free to delete this file

src/utils/helpers.ts
rndquu: @Venoox Perhaps it is better to read a default yml config here?

I mean:
```
const defaultConfig = readFileSync("ubiquibot-config-default.yml", "utf8");
const parsedDefaultConfig: WideRepoConfig = await parseYAML(defaultConfig);
```

And then instead of returning `null` we could use default values from `parsedDefaultConfig`

rndquu (changes requested):

src/utils/helpers.ts
Venoox: You're right, that seems better

src/utils/private.ts
```diff
@@ -118,24 +117,22 @@ export const getScalarKey = async (X25519_PRIVATE_KEY: string | undefined): Prom
 export const getWideConfig = async (context: Context) => {
   const orgConfig = await getConfigSuperset(context, "org");
   const repoConfig = await getConfigSuperset(context, "repo");
-  const defaultConfig = readFileSync("default_config.yml", "utf8");
 
-  const parsedOrg: WideOrgConfig | undefined = await parseYAML(orgConfig);
-  const parsedRepo: WideRepoConfig | undefined = await parseYAML(repoConfig);
+  const parsedOrg: WideOrgConfig | undefined = parseYAML(orgConfig);
+  const parsedRepo: WideRepoConfig | undefined = parseYAML(repoConfig);
```
pavlovcik, line 122: Is that correct to switch this to a synchronous function?

rndquu (approved):

src/utils/private.ts
rndquu: yes, `YAML.parse` [does not return](https://eemeli.org/yaml/#yaml-parse) a promise

pavlovcik (approved):

0xcodercrane (approved):

0xcodercrane:
reverting this PR because I've found something not correct. so there was a path issue to read ubiquibot-config-default.yml in lambda function. 

`Jun 6, 09:11:32 AM: 2023-06-06T01:11:32.613Z	undefined	ERROR	Uncaught Exception 	{"errorType":"Error","errorMessage":"ENOENT: no such file or directory, open 'ubiquibot-config-default.yml'","code":"ENOENT","errno":-2,"syscall":"open","path":"ubiquibot-config-default.yml","stack":["Error: ENOENT: no such file or directory, open 'ubiquibot-config-default.yml'","    at Object.openSync (node:fs:601:3)","    at readFileSync (node:fs:469:35)","    at Object.<anonymous> (/var/task/.netlify/functions/webhooks/webhooks.js:254005:49)","    at Module._compile (node:internal/modules/cjs/loader:1254:14)","    at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)","    at Module.load (node:internal/modules/cjs/loader:1117:32)","    at Module._load (node:internal/modules/cjs/loader:958:12)","    at Module.require (node:internal/modules/cjs/loader:1141:19)","    at require (node:internal/modules/cjs/helpers:110:18)","    at Object.<anonymous> (/var/task/webhooks.js:1:18)"]}
`

@Venoox

rndquu:
> reverting this PR because I've found something not correct. so there was a path issue to read ubiquibot-config-default.yml in lambda function.
> 
> `Jun 6, 09:11:32 AM: 2023-06-06T01:11:32.613Z undefined ERROR Uncaught Exception {"errorType":"Error","errorMessage":"ENOENT: no such file or directory, open 'ubiquibot-config-default.yml'","code":"ENOENT","errno":-2,"syscall":"open","path":"ubiquibot-config-default.yml","stack":["Error: ENOENT: no such file or directory, open 'ubiquibot-config-default.yml'"," at Object.openSync (node:fs:601:3)"," at readFileSync (node:fs:469:35)"," at Object.<anonymous> (/var/task/.netlify/functions/webhooks/webhooks.js:254005:49)"," at Module._compile (node:internal/modules/cjs/loader:1254:14)"," at Module._extensions..js (node:internal/modules/cjs/loader:1308:10)"," at Module.load (node:internal/modules/cjs/loader:1117:32)"," at Module._load (node:internal/modules/cjs/loader:958:12)"," at Module.require (node:internal/modules/cjs/loader:1141:19)"," at require (node:internal/modules/cjs/helpers:110:18)"," at Object.<anonymous> (/var/task/webhooks.js:1:18)"]} `
> 
> @Venoox

So why netlify throws this error? Perhaps we should have adjusted the path here https://github.com/ubiquity/ubiquibot/pull/366/files#diff-d9b0865f2e8f841410cd4faf3e1e4a1fb7cb0b545f583f6adaad5841d7eaa58dR6 ?

Venoox:
I think this task would take more than 1 hour since we need to figure out how to include files in the build process

rndquu:
> I think this task would take more than 1 hour since we need to figure out how to include files in the build process

@Venoox could you:
1. reopen a PR https://github.com/ubiquity/ubiquibot/pull/371
2. set a relative path [here](https://github.com/ubiquity/ubiquibot/pull/366/files#diff-d9b0865f2e8f841410cd4faf3e1e4a1fb7cb0b545f583f6adaad5841d7eaa58dR6)

I guess this way there should not be any errors

PR #368, opened 2023-06-05
build: set payouts to gnosis chain

rndquu:
This PR enables payouts for gnosis chain (i.e. all bounty hunters will start to receive rewards in gnosis chain's WXDAI)


pavlovcik (approved):
The future is now

Issue #315, opened 2023-05-18
Default Label Colors


pavlovcik:
Make sure that all the labels match GitHub colors. The default label color is gray. Only Price labels should be green.

- `ededed` - gray - Default
- `1f883d` - green - `Price: `

PR #370, opened 2023-06-05
feat: added default label color and updated price label color

sweetapplepumpkinpie:
Completed #315


pavlovcik (approved):
I have some hesitation around making special behavior around the word "price" specifically as I believe that code, like the functions themselves, should use generalized terminology to describe their purpose abstracted away from business logic. 

Basically I'm debating on whether we should refer to the color as "highlighted" instead of "price"

0xcodercrane (approved):

PR #371, opened 2023-06-06
Revert "feat: make default config"

0xcodercrane:
Reverts ubiquity/ubiquibot#366


pavlovcik:
Why was this reverted

0xcodercrane:
> Why was this reverted?

It has a path issue that broke the staging bot and blocked @Draeieg 's QA tasks. I wouldn't be able to fix it so reverted it.
Please take a look at the following comment.
https://github.com/ubiquity/ubiquibot/pull/366#issuecomment-1577753888

PR #377, opened 2023-06-06
fix: fix ERR_REQUIRE_ESM

rndquu:
This PR fixes this [error](https://github.com/ubiquity/ubiquibot/issues/290#issuecomment-1573450082)

The `mdast-util-from-markdown` is an ESM module. Our project setup does not support such modules (we support only commonjs modules). As soon as probot [moves to ESM](https://github.com/probot/probot/issues/1684) we can also [enable full ESM support](https://github.com/ubiquity/ubiquibot/issues/378) for our project.


0xcodercrane:
This error is really frustrating because it is happening to most of the bounty hunters recently. I just want to make sure this will resolve it finally. @rndquu if you are 100% sure, lets merge it

0xcodercrane (approved):

PR #380, opened 2023-06-07
feat: comment out payout command

### LINKED ISSUE #353: Comment out the `/payout` command
We should comment out the `/payout` command (don't forget the help menu) and uncomment this command when gnosis safe is ready to be used.

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/290#issuecomment-1569831915_

Venoox:
resolves #353

I commented out the payout command in the user command list which also removes it from the help menu. 
I didn't remove the actual implementation of the command since it will be enabled once Gnosis Safe is supported

### DIFF
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -3,7 +3,6 @@ import { Payload, UserCommands } from "../../../types";
 import { IssueCommentCommands } from "../commands";
 import { assign } from "./assign";
 import { listAvailableCommands } from "./help";
-import { payout } from "./payout";
 import { unassign } from "./unassign";
 import { registerWallet } from "./wallet";
 import { setAccess } from "./set-access";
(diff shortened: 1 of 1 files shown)


src/handlers/comment/handlers/index.ts
```diff
@@ -3,7 +3,6 @@ import { Payload, UserCommands } from "../../../types";
 import { IssueCommentCommands } from "../commands";
 import { assign } from "./assign";
 import { listAvailableCommands } from "./help";
-import { payout } from "./payout";
```
pavlovcik, line 6: Perhaps this should also be left as a comment.

pavlovcik (changes requested):
I think you should also add a note (as a comment in the code) why these are commented out and link the original issue. 

That way future developers will know not to remove the commented out code.

src/handlers/comment/handlers/index.ts
0xcodercrane: agree with @pavlovcik 's request

rndquu (approved):

pavlovcik (approved):

0xcodercrane (approved):

PR #389, opened 2023-06-11
Complete re-enabling github action

0xcodercrane:
Resolves #https://github.com/ubiquity/ubiquibot/issues/310

### DIFF
--- a/src/handlers/push/index.ts
+++ b/src/handlers/push/index.ts
@@ -26,6 +26,6 @@ export const runOnPush = async () => {
   // check for modified or added files and check for specified file
   if (payload.head_commit.modified.includes(BASE_RATE_FILE) || payload.head_commit.added.includes(BASE_RATE_FILE)) {
     // update base rate
-    updateBaseRate(context, payload, BASE_RATE_FILE);
+    await updateBaseRate(context, payload, BASE_RATE_FILE);
   }
 };
--- a/.github/workflows/bot.yml
+++ b/.github/workflows/bot.yml
@@ -20,7 +20,7 @@ jobs:
   calculate_bounty_job:
     # ignore events invoked by bots
     if: >-
-      github.event.pull_request.payload.sender.type != 'Bot'
+      github.event.pull_request.payload.sender.type != 'Bot' && github.repository != 'ubiquity/ubiquibot'
     runs-on: ubuntu-latest
     name: Calculate Bounty with UbiquiBot
     steps:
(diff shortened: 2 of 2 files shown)


0xcodercrane:
[here](https://github.com/0xcodercrane/ubiquibot/issues/7)'s the QA for this

0xcodercrane:
The problem with `tsx adapters/index.ts` is that it requires `yarn install` and `yarn build`. this slow down the execution time of the action. but i leave it for now bc i guess its the original spec to use `tsx adapters/index.ts` @pavlovcik

pavlovcik:
> The problem with `tsx adapters/index.ts` is that it requires `yarn install` and `yarn build`. this slow down the execution time of the action. but i leave it for now bc i guess its the original spec to use `tsx adapters/index.ts` @pavlovcik 

`yes | npx tsx adapters/index.ts`

`yarn build` it's inconvenient for me to check from my phone now what it does but typically this just compiles typescript into javascript. This is not a necessary step when using the `tsx` runtime

0xcodercrane:
you're right. we can remove `yarn build`

src/handlers/push/index.ts
```diff
@@ -26,6 +26,6 @@ export const runOnPush = async () => {
   // check for modified or added files and check for specified file
   if (payload.head_commit.modified.includes(BASE_RATE_FILE) || payload.head_commit.added.includes(BASE_RATE_FILE)) {
     // update base rate
-    updateBaseRate(context, payload, BASE_RATE_FILE);
+    await updateBaseRate(context, payload, BASE_RATE_FILE);
```
pavlovcik, line 29: I'm pretty sure I fixed this in the original PR. How was it removed?

0xcodercrane: when I catch it up, it was not there. so I think I added it again.

pavlovcik: I made a mistake and realized that it is a different function with a similar name that I fixed in https://github.com/ubiquity/ubiquibot/pull/352#discussion_r1216587731

0xcodercrane: Yeah that was the only part but I can go through it.

0xcodercrane: I don't think we should fix `await` problems in this PR. I suggest to do that out of this PR.

pavlovcik (dismissed):

pavlovcik:
CI isn't passing.

PR #393, opened 2023-06-13
feat: added time config for max issue

### LINKED ISSUE #388: Max bounty hunter issues improvement - Configurable Time
As an extension of #298 I realized that we should make the time configurable. Please add the config property so our partners can customize this behavior. If you need a default setting, we can make it 24 hours. 

Lastly, we should be able to disable this feature as well. I'm unsure how this will be implemented, perhaps if the config property is not found, or the value is not parseable/valid it can disable. 

> Code looks good. You achieved the specification, thanks for that, but I think I will like to create a new bounty to make the 24 hour period configurable (this also means we need to add a new config property.

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/pull/387#pullrequestreview-1473321237_

wannacfuture:
Resolves #388 
Test successful.
You can see the test result at https://github.com/wannacfuture/Battleship/issues/19
There, I can't assign to that issue even I have one pull request which is created over 24 hours before, because TIME_RANGE_FOR_MAX_ISSUE_ENABLED was set `false` so bot doesn't count PRs.
Also I added TIME_RANGE_FOR_MAX_ISSUE for flexible time configuration for max-bounty-issue.

### DIFF
--- a/src/configs/shared.ts
+++ b/src/configs/shared.ts
@@ -5,7 +5,8 @@ export const COLORS = {
   price: "1f883d",
 };
 export const DEFAULT_BOT_DELAY = 100; // 100ms
-
+export const TIME_RANGE_FOR_MAX_ISSUE = 24;
+export const TIME_RANGE_FOR_MAX_ISSUE_ENABLED = true;
 /**
  * ms('2 days')  // 172800000
  * ms('1d')      // 86400000
--- a/src/configs/shared.ts
+++ b/src/configs/shared.ts
@@ -5,8 +5,8 @@ export const COLORS = {
   price: "1f883d",
 };
 export const DEFAULT_BOT_DELAY = 100; // 100ms
-export const TIME_RANGE_FOR_MAX_ISSUE = 24;
-export const TIME_RANGE_FOR_MAX_ISSUE_ENABLED = true;
+export const DEFAULT_TIME_RANGE_FOR_MAX_ISSUE = 24;
+export const DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED = false;
 /**
  * ms('2 days')  // 172800000
  * ms('1d')      // 86400000
--- a/src/configs/shared.ts
+++ b/src/configs/shared.ts
@@ -6,7 +6,7 @@ export const COLORS = {
 };
 export const DEFAULT_BOT_DELAY = 100; // 100ms
 export const DEFAULT_TIME_RANGE_FOR_MAX_ISSUE = 24;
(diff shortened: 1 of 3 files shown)


src/helpers/issue.ts
```diff
@@ -338,6 +339,7 @@ export const getAssignedIssues = async (username: string) => {
 };
 
 export const getOpenedPullRequestWithNoReviewsOver24HoursPassedAfterCreated = async (username: string) => {
```
pavlovcik, line 341: To increase code clarity I recommend splitting this into separate functions `getOpenedPullRequests` 
- and then filter by no reviews i.e. `noReviews` 
- and then filter again by `timePassed`

src/configs/shared.ts
```diff
@@ -5,7 +5,8 @@ export const COLORS = {
   price: "1f883d",
 };
 export const DEFAULT_BOT_DELAY = 100; // 100ms
-
+export const TIME_RANGE_FOR_MAX_ISSUE = 24;
+export const TIME_RANGE_FOR_MAX_ISSUE_ENABLED = true;
```
pavlovcik, line 9:
```suggestion
export const DEFAULT_TIME_RANGE_FOR_MAX_ISSUE = 24;
export const DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED = true;
```

pavlovcik (changes requested):
Left comments.

wannacfuture:
> Left comments.

@pavlovcik , fixed.

0xcodercrane:
The codebase looks good to me. @Draeieg  would you be handling the QA with @wannacfuture  here?

0xcodercrane (approved):

Draeieg:
> The codebase looks good to me. @Draeieg  would you be handling the QA with @wannacfuture  here?

He already provided an example QA, I'd say we can merge to development and I'll replicate in staging just to be sure

src/handlers/comment/handlers/assign.ts
```diff
@@ -28,7 +22,7 @@ export const assign = async (body: string) => {
     return;
   }
 
-  const opened_prs = await getOpenedPullRequestWithNoReviewsOver24HoursPassedAfterCreated(payload.sender.login);
+  const opened_prs = await getAvailableOpenedPullRequests(payload.sender.login);
 
   logger.info(`Opened Pull Requests with no reviews but over 24 hours have passed: ${JSON.stringify(opened_prs)}`);
```
ubiquity-bounties, line 27: This log comment should be updated because the `24 hours` is configurable.  @wannacfuture

Issue #390, opened 2023-06-12
Ensure Awaits


pavlovcik:
> I made a mistake and realized that it is a different function with a similar name that I fixed in https://github.com/ubiquity/ubiquibot/pull/352#discussion_r1216587731.

> Can you go through the codebase and make sure that we are `await`ing every async function like these?

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/pull/389#discussion_r1226217798_

0xcodercrane:
> Do you have any updates @0xcodercrane? If you would like to release the bounty back to the DevPool, please comment `/unassign`

Yes, spending my time on other stuff.

pavlovcik:
I think you should assign yourself only on things you're actively working on. I understand that you have dynamic priorities but maybe it's best to unassign yourself before assigning to the latest most urgent task?

0xcodercrane:
> Do you have any updates @0xcodercrane? If you would like to release the bounty back to the DevPool, please comment `/unassign`

I am fixing the warnings that `yarn lint` caught up

pavlovcik:
Can you finish this bounty this week?

PR #399, opened 2023-06-13
Feat/ensure async await

0xcodercrane:
Resolves #https://github.com/ubiquity/ubiquibot/issues/390


pavlovcik (commented):
Very cool that GitHub Actions highlights the issues in the code review tab now due to your `eslint` presumably.

src/handlers/access/labels-access.ts
```diff
@@ -33,12 +33,12 @@ export const handleLabelsAccess = async () => {
 
     if (payload.action === "labeled") {
       // remove the label
-      removeLabel(payload.label?.name!);
+      await removeLabel(payload.label?.name!);
```
pavlovcik, line 36: Maybe we should ban using non-null assertions via eslint because I see them being abused quite a bit in this codebase.

pavlovcik (commented):
Looks like you're moving in the right direction. Perhaps we can make a new bounty to clean up the rest of the codebase according to the GitHub Actions / build inline notifications I see on the code view here.

rndquu (approved):

.eslintrc.js
pavlovcik, line 1: We should use `.ts` files exclusively, or `.json`

src/helpers/issue.ts
```diff
@@ -18,32 +19,36 @@ export const clearAllPriceLabelsOnIssue = async (): Promise<void> => {
...
       labels: [labelName],
     });
   } catch (e: unknown) {
-    logger.debug(`Adding a label to issue failed!, reason: ${(e as any)?.message}`);
+    logger.debug(`Adding a label to issue failed!, reason: ${e}`);
   }
 };
 
-export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open", per_page: number = 30, page: number = 1) => {
+export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open", per_page = 30, page = 1) => {
```
pavlovcik, line 51: Is `30` the max?

pavlovcik (dismissed):
Looks solid just would prefer to change that linter config to `.ts` or `.json`

0xcodercrane:
> Looks solid just would prefer to change that linter config to `.ts` or `.json`

I am not sure `.json` would work. but the same thing that we have `commitlint.config.js`.

src/helpers/issue.ts
pavlovcik: @0xcodercrane

PR #401, opened 2023-06-14
feat: feature to interpret the multiplier reason

### LINKED ISSUE #376: Multiplier Reason
We should edit the `/multiplier ` command to also accept string input after the numerical value in order to record the reason that the contributor has a custom multiplier. This is to increase transparency for all DAO contributors for why this particular contributor has a different compensation rate.

For example:

```
/multiplier 0.5 @pavlovcik "paid a base salary as a core contributor"
```

Ideally the command should be intelligent enough to understand what each is even if they are out of order.

- numbers are parsed as the multiplier value.
- strings beginning with "@" are usernames.
- the remainder is the reason.

The bot should confirm its interpretation in a response comment.

For example:

> Successfully changed the payout multiplier for `pavlovcik` to `0.5`. The reason provided is `paid a base salary as a core contributor`.

This will be displayed to users in #375

sweetapplepumpkinpie:
Resolves #376

QA issue https://github.com/sweetapplepumpkinpie/ubiquibot/issues/1

### DIFF
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -31,7 +31,7 @@ export const commentParser = (body: string): IssueCommentCommands[] => {
   // TODO: As a starting point, it may be simple but there could be cases for the comment to includes one or more commands
   // We need to continuously improve to parse even complex comments. Right now, we implement it simply.
   const commandList = Object.values(IssueCommentCommands) as string[];
-  const result = commandList.filter((command: string) => body.includes(command));
+  const result = commandList.filter((command: string) => (command === "/multiplier" ? body.startsWith(command) : body.includes(command)));
   return result as IssueCommentCommands[];
 };
 
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -31,7 +31,7 @@ export const commentParser = (body: string): IssueCommentCommands[] => {
   // TODO: As a starting point, it may be simple but there could be cases for the comment to includes one or more commands
   // We need to continuously improve to parse even complex comments. Right now, we implement it simply.
   const commandList = Object.values(IssueCommentCommands) as string[];
-  const result = commandList.filter((command: string) => (command === "/multiplier" ? body.startsWith(command) : body.includes(command)));
+  const result = commandList.filter((command: string) => body.startsWith(command));
   return result as IssueCommentCommands[];
 };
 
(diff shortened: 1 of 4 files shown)


src/handlers/comment/handlers/index.ts
```diff
@@ -31,7 +31,7 @@ export const commentParser = (body: string): IssueCommentCommands[] => {
   // TODO: As a starting point, it may be simple but there could be cases for the comment to includes one or more commands
   // We need to continuously improve to parse even complex comments. Right now, we implement it simply.
   const commandList = Object.values(IssueCommentCommands) as string[];
-  const result = commandList.filter((command: string) => body.includes(command));
+  const result = commandList.filter((command: string) => (command === "/multiplier" ? body.startsWith(command) : body.includes(command)));
```
rndquu, line 34: Isn't it better to implement this update for all commands? 
```
const result = commandList.filter((command: string) => body.startsWith(command));
```

sweetapplepumpkinpie: As you can see the comments, there could be cases for the comment to includes one or more commands.

rndquu: right now we don't support multiple commands in a single comment

I'm trying to understand why we should have a special treat **only** for the `/multiplier` command

As far as I understand it is simpler to use `body.startsWith(command)`

@0xcodercrane help

0xcodercrane: We don't support multiple commands in a single comment that's right. In terms of hardcoded `/multiplier`, I agree to remove it from the line. If there's a case to use it for comparison, you should use it like [IssueCommentsCommands.MULTIPLIER](https://github.com/ubiquity/ubiquibot/blob/development/src/handlers/comment/commands.ts#L7)

sweetapplepumpkinpie:
@0xcodercrane Done
QA - https://github.com/sweetapplepumpkinpie/ubiquibot/issues/1

0xcodercrane:
> @0xcodercrane Done QA - [sweetapplepumpkinpie#1](https://github.com/sweetapplepumpkinpie/ubiquibot/issues/1)

It looks good to me. but let me wait for @Draeieg to react with 👍  because he is responsible for QA.

sweetapplepumpkinpie:
Hi @Draeieg Could you review my PR?

Draeieg:
> > @0xcodercrane Done QA - [sweetapplepumpkinpie#1](https://github.com/sweetapplepumpkinpie/ubiquibot/issues/1)
> 
> It looks good to me. but let me wait for @Draeieg to react with 👍  because he is responsible for QA.

Looks good and seems easy to verify in our end

Issue #176, opened 2023-03-07
Unit Testing


pavlovcik:
Deciding if we need to setup GitHub Actions Runner Unit testing in a Docker instance because the set price feature was broken due to the labels being changed. These sort of capabilities must be functional!

We should set up unit tests.

pavlovcik:
Good luck @sweetapplepumpkinpie I have a feeling that this one might be quite complicated.

sweetapplepumpkinpie:
Thanks @pavlovcik. will complete this asap

sweetapplepumpkinpie:
Sorry for the late start. I just thought that I would continue after the last issue is merged.
So the basic unit tests should cover all of the project?

pavlovcik:
The purpose of this bounty is more to set up a foundation so they can be just a few simple ones.

molecula451:
let's work on it!! great @pavlovcik any particular reason why we want a meta dockerized unit testing? (just for the thoughts here, i will handle it anyway)

pavlovcik:
It's a best guess on how to closely match the GitHub actions runner environment. I did a small bit of research before and found that there are docker containers specifically to act like a GitHub Action runner. 

Do you have any other suggestions?

Maybe @rndquu might have some ideas.

rndquu:
Let's clear the terminology:
- Unit tests: fast and small tests which mock most of the external dependencies
- Integration tests: tests running in a simulated environment without mocking external dependencies 

This issue implies setting up a foundation for unit testing, i.e.:
1. Setup `jest` for the project
2. Add basic tests for the following commands:
 - `/assign`
 - `/unassign`
 - `/wallet`
 - `/payout`
 - `/multiplier`
3. Add a new script to the `package.json` that runs unit tests
4. Add unit tests execution to a CI workflow (perhaps to the "Build"  workflow)

At later steps, when we are about to implement integration tests, we could consider using docker for simulating github environment. Perhaps there are better ways for implementing integration tests.

pavlovcik:
Okay so let's remove Docker from the spec and just run `jest` in the CI.

rndquu:
Closing this one because:
- jest (along with jest CI) will be setup [here](https://github.com/ubiquity/ubiquibot/pull/640) 
- unit (functional) tests for basic commands will be implemented [here](https://github.com/ubiquity/ubiquibot/issues/634)

PR #413, opened 2023-06-18
feat: fixed followup and unassignment

### LINKED ISSUE #408: Unreliable Followups/Unassignment
The bot seems to be inconsistent on who it follows up with and then unassigns when dealing with bounties that are assigned but have no updates happening. This is an important feature that should be functional, or else we will have many bounties being reserved and neglected. 

I have a feeling that this stopped being reliable after implementing the capability for the bot to monitor linked pull requests for updates. But this feeling may be unfounded cause I haven't done the research to confirm it. 

              Thanks @pavlovcik. will complete this asap

_Originally posted by @sweetapplepumpkinpie in https://github.com/ubiquity/ubiquibot/issues/176#issuecomment-1583449657_
            

---

I will compile examples as I find them.

- https://github.com/ubiquity/ubiquity-dollar/issues/516

wannacfuture:
Resolves #408 

QA issue1: https://github.com/wannacfuture/Battleship/issues/20#issuecomment-1596167479
There, the user assigned that issue 5 days ago, but never did any activity. So the bot follows him up.
QA issue2: https://github.com/wannacfuture/Battleship/issues/17#issuecomment-1596165596
There, the use assigned that issue last week but never did any activity. So the bot releases the bounty

### _**IMPORTANT**_
Here, the bot monitors the activity of bounty hunter.
And last activity time means the latest time among:
- last comment time on the issue from the bounty hunter.
- committed time of last commit on the linked pull request from the bounty hunter.
- created time of last comment on the linked pull request from the bounty hunter.

Test successful.

### DIFF
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -374,7 +374,7 @@ export const getOpenedPullRequests = async (username: string) => {
   return prs.filter((pr) => !pr.draft && pr.user?.login === username);
 };
 
-export const getCommitsOnPullRequest = async (pull_number: number) => {
+export const getCommitsOnPullRequest = async (pullNumber: number) => {
   const logger = getLogger();
   const context = getBotContext();
   const payload = getBotContext().payload as Payload;
@@ -382,7 +382,7 @@ export const getCommitsOnPullRequest = async (pull_number: number) => {
     const { data: commits } = await context.octokit.rest.pulls.listCommits({
       owner: payload.repository.owner.login,
       repo: payload.repository.name,
-      pull_number,
+      pull_number: pullNumber,
     });
     return commits;
   } catch (e: unknown) {
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -374,6 +374,23 @@ export const getOpenedPullRequests = async (username: string) => {
   return prs.filter((pr) => !pr.draft && pr.user?.login === username);
 };
 
+export const getCommitsOnPullRequest = async (pull_number: number) => {
+  const logger = getLogger();
+  const context = getBotContext();
+  const payload = getBotContext().payload as Payload;
+  try {
+    const { data: commits } = await context.octokit.rest.pulls.listCommits({
+      owner: payload.repository.owner.login,
+      repo: payload.repository.name,
+      pull_number,
+    });
+    return commits;
+  } catch (e: unknown) {
+    logger.debug(`Fetching pull request commits failed!, reason: ${e}`);
+    return [];
+  }
+};
+
 export const getAvailableOpenedPullRequests = async (username: string) => {
   if (!DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED) return [];
   const context = getBotContext();
--- a/src/handlers/wildcard/unassign.ts
+++ b/src/handlers/wildcard/unassign.ts
@@ -1,9 +1,14 @@
-import { getBotConfig, getBotContext, getLogger } from "../../bindings";
+import { getBotConfig, getLogger } from "../../bindings";
 import { GLOBAL_STRINGS } from "../../configs/strings";
-import { addCommentToIssue, getCommentsOfIssue, listIssuesForRepo, removeAssignees } from "../../helpers";
-import { gitIssueParser } from "../../helpers/parser";
-import { Comment, IssueType, Payload } from "../../types";
-import { deadLinePrefix } from "../shared";
+import {
+  addCommentToIssue,
+  getCommentsOfIssue,
+  getCommitsOnPullRequest,
+  getOpenedPullRequestsForAnIssue,
+  listIssuesForRepo,
+  removeAssignees,
+} from "../../helpers";
+import { Comment, IssueType } from "../../types";
 
 /**
  * @dev Check out the bounties which haven't been completed within the initial timeline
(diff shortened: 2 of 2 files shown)


pavlovcik:
The linked examples look good but what was the root cause of the issue?

[> This bounty requires that you figure out the root cause of the issue, as well as provide a fix.](https://github.com/ubiquity/ubiquibot/issues/408#issuecomment-1594158464)

src/helpers/issue.ts
```diff
@@ -374,6 +374,23 @@ export const getOpenedPullRequests = async (username: string) => {
   return prs.filter((pr) => !pr.draft && pr.user?.login === username);
 };
 
+export const getCommitsOnPullRequest = async (pull_number: number) => {
```
pavlovcik, line 377:
```suggestion
export const getCommitsOnPullRequest = async (pullRequest: number) => {
```

We use camel case!

pavlovcik (changes requested):

wannacfuture:
Yea, the root cause of the issue was that the mentioned comment of that bounty hunter is created less than a week before.
So the bot identified that issue as it has activity within the week so, it didn't unassign the hunter.

src/helpers/issue.ts
wannacfuture: fixed.

wannacfuture:
@0xcodercrane , could you review this one?

0xcodercrane (approved):
The codebase looks good. QA is also good but @Draeieg needs to verify it finally.

pavlovcik:
> QA is also good but @Draeieg needs to verify it finally.

Shouldn't you request a review from Draeieg if you need QA?

wannacfuture:
@Draeieg ? 😳

Draeieg (approved):

pavlovcik:
> Yea, the root cause of the issue was that the mentioned comment of that bounty hunter is created less than a week before.
> So the bot identified that issue as it has activity within the week so, it didn't unassign the hunter.

I don't believe that's correct. The reason why I made the issue is because it had already been a full week, and I expected that they should have been unassigned.

wannacfuture:
> I don't believe that's correct. The reason why I made the issue is because it had already been a full week, and I expected that they should have been unassigned. @Draeieg do you have any theories?

No, it had already been a full week, but there is a comment of the hunter on the issue within the week before.
Could you take a closer look? @pavlovcik ?

pavlovcik:
The easiest way to prove your point to me is if you can link the two comments you are talking about here.

wannacfuture:
on the issue https://github.com/ubiquity/ubiquibot/issues/176#issue-1614068275 , 
the bot commented about the time limit on https://github.com/ubiquity/ubiquibot/issues/176#issuecomment-1581824756 .
You see? the created time is `Jun 8 2023 12:09 PM GMT+9`
then Lets move to this comment : https://github.com/ubiquity/ubiquibot/issues/176#issuecomment-1583449657
the created time is `Jun 9 2023 7:03 AM GMT+9`
Then, as you can see - @pavlovcik , you mentioned about that issue that it is not unassigned, but actually at that time - `Jun 16 2023 6:57 AM GMT +9` I think you can calculate the time - it has been less than a week!! right?
then the hunter added comment at `Jun 16 2023 5:44 PM GMT +9` (plz don't think the bot works real-time, lol)
if the hunter added comment the day after, or something the bot would have unassigned it but luckily it wasn't.

pavlovcik:
The bot is supposed to follow up with a comment within four days. The comment examples you provided are more than four days apart.

rndquu:
@0xcodercrane do we have [FOLLOW_UP_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) and [DISQUALIFY_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L49) env variables set in the netlify prod environment? If so then what are their values?

wannacfuture:
@rndquu , The bot add comment : "Do you have any update ...?" after FOLLOW_UP_TIME has passed.
And the bot unassines the hunter after DISQUALIFY_TIME has passed.

Seems something went wrong at that moment(FOLLOW_UP_TIME has not been set or something).
I think we need to take a look for some period if my MR is working. @pavlovcik

pavlovcik:
> @0xcodercrane do we have [FOLLOW_UP_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) and [DISQUALIFY_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L49) env variables set in the netlify prod environment? If so then what are their values?

You can verify in the future by logging into @ubiquity-bounties and then logging into netlify. 

### Production
image

### Staging
image

rndquu:
> > @0xcodercrane do we have [FOLLOW_UP_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) and [DISQUALIFY_TIME](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L49) env variables set in the netlify prod environment? If so then what are their values?
> 
> * You can verify in the future by logging into @ubiquity-bounties and then logging into netlify.
> * Great diagnosis rndquu!
> 
> * I just set `FOLLOW_UP_TIME` to `4`
> * I just set `DISQUALIFY_TIME` to `7`
> 
> Does that mean that this pull request was unnecessary?

> You can verify in the future by logging into @ubiquity-bounties

You mean authenticate by login and password? I don't have a password for the @ubiquity-bounties account

> I just set `FOLLOW_UP_TIME` to `4`,  I just set `DISQUALIFY_TIME` to `7`

This breaks unassignment because [here](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) `ms` is used which converts `ms(4)` to `4ms` string value which breaks unassignment. You should either delete `FOLLOW_UP_TIME` and `DISQUALIFY_TIME` env variables either set them to their correct default [string values](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24)

> Does that mean that this pull request was unnecessary?

No, this PR is necessary. I simply wanted to check whether env variables (`FOLLOW_UP_TIME` and `DISQUALIFY_TIME`) are set and that their values are correct.

pavlovcik:
> You mean authenticate by login and password? I don't have a password for the @ubiquity-bounties account

Shared in direct message.

> This breaks unassignment because [here](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) ms is used which converts ms(4) to 4ms string value which breaks unassignment. You should either delete FOLLOW_UP_TIME and DISQUALIFY_TIME env variables (this way default values will be used) either set them to their correct default [string values](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24)

Okay then try logging in and resetting the values to what you think makes sense.

rndquu:
> > You mean authenticate by login and password? I don't have a password for the @ubiquity-bounties account
> 
> Shared in direct message.
> 
> > This breaks unassignment because [here](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) ms is used which converts ms(4) to 4ms string value which breaks unassignment. You should either delete FOLLOW_UP_TIME and DISQUALIFY_TIME env variables (this way default values will be used) either set them to their correct default [string values](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24)
> 
> Okay then try logging in and resetting the values to what you think makes sense.

@0xcodercrane notice that I've updated env variables in prod netlify deploy to these values (which are used by [default](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24) if `FOLLOW_UP_TIME` and `DISQUALIFY_TIME` are not set at all):
```
FOLLOW_UP_TIME="4 days"
DISQUALIFY_TIME="7 days"
```

@Draeieg FYI on staging we have the following "unassign" setup:
```
FOLLOW_UP_TIME="24 hrs"
DISQUALIFY_TIME="25 hrs"
```

Draeieg:
> > > You mean authenticate by login and password? I don't have a password for the @ubiquity-bounties account
> > 
> > Shared in direct message.
> > 
> > > This breaks unassignment because [here](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts#L48) ms is used which converts ms(4) to 4ms string value which breaks unassignment. You should either delete FOLLOW_UP_TIME and DISQUALIFY_TIME env variables (this way default values will be used) either set them to their correct default [string values](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24)
> > 
> > Okay then try logging in and resetting the values to what you think makes sense.
> 
> @0xcodercrane notice that I've updated env variables in prod netlify deploy to these values (which are used by [default](https://github.com/ubiquity/ubiquibot/blob/development/src/configs/shared.ts#L24) if `FOLLOW_UP_TIME` and `DISQUALIFY_TIME` are not set at all):
> ```
> FOLLOW_UP_TIME="4 days"
> DISQUALIFY_TIME="7 days"
> ```
> 
> @Draeieg FYI on staging we have the following "unassign" setup:
> ```
> FOLLOW_UP_TIME="24 hrs"
> DISQUALIFY_TIME="25 hrs"
> ```

Yes, I was the one who requested those times to facilitate QA

Draeieg:
hey dm me at https://t.me/Draeieg  @wannacfuture 
7529d9aa-e610-4b30-8fbd-b948fd6ea3d7_text

pavlovcik:
@Draeieg I can give you direct access to change environment variables if needed just dm me.

PR #414, opened 2023-06-19
chore: comment

0xcodercrane:
Resolves # Assign issue

### DIFF
--- a/src/bindings/config.ts
+++ b/src/bindings/config.ts
@@ -62,11 +62,11 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     botConfig.mode.autoPayMode = false;
   }
 
-  const validate = ajv.compile(BotConfigSchema);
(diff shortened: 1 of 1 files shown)


pavlovcik:
What exactly was the original rationale behind schema validation?

All errors should be dumping to the bot comment so that it's clear there was an error to the user. 

Simply removing this feature I think this is a bad idea. Instead we should fix the edge cases or loosen up the validation for certain schemas as needed.

pavlovcik (changes requested):
What exactly was the original rationale behind schema validation?

All errors should be dumping to the bot comment so that it's clear there was an error to the user.

Simply removing this feature I think this is a bad idea. Instead we should fix the edge cases or loosen up the validation for certain schemas as needed.

0xcodercrane:
From the discussion in dm, we decidied to add a global error handler and post it to the corresponding issue.

PR #420, opened 2023-06-20
chore: june cleaning

pavlovcik:
General cleanup. Can you make sure that its fine that I deleted the `bin/index.js` file etc?

It's not clear to me how the bot will respond to comment events in the GitHub Action runtime.


.github/workflows/bot.yml
```diff
@@ -12,9 +12,9 @@ on:
     types:
       - created
       - edited
-  pull_request:
-    types:
-      - opened
+  # pull_request:
+  #   types:
+  #     - opened
```
pavlovcik, line 17: @0xcodercrane this threw an error when I opened the pull request. Why is this here?

.github/workflows/bot.yml
```diff
@@ -33,15 +33,18 @@ jobs:
...
-        
+
       - name: Authenticate as UbiquiBot
         uses: tibdex/github-app-token@v1.7.0
         id: get_installation_token
         with:
           app_id: 346074
+          # https://github.com/organizations/ubiquibot/settings/apps/ubiquibot-qa/installations
+          # https://github.com/apps/ubiquibot-qa/installations/new/permissions?target_id=4975670
+          # https://github.com/settings/installations/38790399
```
pavlovcik, line 47: I am unsure which is the correct link but this is to install the app instance on a fork so that this authenticates. These are the three links from the whole install flow.

0xcodercrane:
Looks like a little big refactors. I am gonna review the PR after https://github.com/ubiquity/ubiquibot/pull/422 and https://github.com/ubiquity/ubiquibot/pull/399 merged.

package.json
```diff
@@ -13,10 +13,12 @@
   "scripts": {
     "build": "tsc",
     "build:serverless": "ncc build src/index.ts -o dist",
-    "build:ci": "ncc build adapters/index.ts -o dist && mv ./dist/index.js ./bin",
+    "build:ci": "ncc build adapters/github-actions.ts -o dist",
+    "postbuild:ci": "mv ./dist/index.js ./bin",
```
rndquu, line 17: The `bin` folder is removed so `mv ./dist/index.js ./bin` will move `index.js` to the `bin` **file** instead of the `bin` folder. So in order to keep the `bin` folder indexed we could add an empty `.gitkeep` file to the `bin` folder

.github/workflows/bot.yml
pavlovcik: @0xcodercrane

package.json
pavlovcik: I decided to build to project root and run from there instead.

PR #438, opened 2023-06-27
refactor: remove unused files and cleanup

pavlovcik:


api/github/webhooks/index.ts
pavlovcik, line 1: @0xcodercrane I deleted this because you said its okay to.

0xcodercrane (approved):

PR #443, opened 2023-06-30
feat: added qa automation yml

wannacfuture:
Added QA automation YML
closes https://github.com/ubiquibot/QA-Improvements/issues/1

### DIFF
--- a/.github/workflows/qa-automation.yml
+++ b/.github/workflows/qa-automation.yml
@@ -1,0 +1,28 @@
+name: Add pull requests to project
+
+on:
+  pull_request:
+    types:
+      - review_requested
+
+jobs:
+  add-to-project:
+    name: Add issue to project
+    runs-on: ubuntu-latest
+    steps:
+      - name: add pr to project
+        if: contains(github.event.pull_request.requested_reviewers.*.login, 'Draeieg')
+        uses: actions/add-to-project@v0.5.0
+        with:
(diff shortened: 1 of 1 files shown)


pavlovcik (commented):
Maybe this should only be on the `development` branch?

You should also associate this with the correct issue if possible.

wannacfuture:
>Maybe this should only be on the development branch?
Correct. 
>You should also associate this with the https://github.com/ubiquibot/QA-Improvements/issues/1 if possible.
You mean, adding issue instead of PR?
@Draeieg  said PR also works for him. I wonder if its okay for you if I try that after checking this pr works.

.github/workflows/qa-automation.yml
```diff
@@ -0,0 +1,28 @@
+name: Add pull requests to project
+
+on:
+  pull_request:
+    types:
+      - review_requested
```
pavlovcik, line 6: You need to specify the branch, probably `development`

pavlovcik (changes requested):

.github/workflows/qa-automation.yml
wannacfuture: You mean, only for the prs whose base is 'development'?

wannacfuture: fixed.

PR #462, opened 2023-07-05
penalty for reopened issue

### LINKED ISSUE #439: Penalty for reopened issues
There is an [issue](https://github.com/ubiquity/ubiquity-dollar/issues/643) for which we had to pay twice because it was not correctly implemented for the 1st time. [Here](https://github.com/ubiquity/ubiquity-dollar/issues/643#issuecomment-1607375963) is how it happened. While proper QA eliminates such cases we should also add a penalty for a bounty hunter.

What should be done:
- if issue is reopened and permit was earlier generated then subtract "penalized" issue reward from the next bounty hunter's solved issue (if there is one)
- the bot should post a "penalty" message before generating a permit URL so that a bounty hunter knew that there is a penalty and the reward is reduced

Venoox:
Resolves #439 

QA: https://github.com/Venoox/ubiquibot/issues/13 https://github.com/Venoox/ubiquibot/issues/16


src/handlers/comment/handlers/index.ts
```diff
@@ -81,6 +83,82 @@ export const issueCreatedCallback = async (): Promise<void> => {
...
+    // write penalty to db
+    try {
+      await addPenalty(assignee, repository.full_name, tokenAddress, chainId.toString(), amount);
+    } catch (err) {
+      logger.error(`Error writing penalty to db: ${err}`);
+      return;
+    }
+
+    await addCommentToIssue(
+      `@${assignee} please be sure to review this conversation and implement any necessary fixes. Unless this is closed as completed, its payment of ${formattedAmount} ${tokenSymbol} will be deducted from your next bounty.`,
```
pavlovcik, line 154:
```suggestion
      `@${assignee} please be sure to review this conversation and implement any necessary fixes. Unless this is closed as completed, its payment of **${formattedAmount} ${tokenSymbol}** will be deducted from your next bounty.`,
```

src/handlers/payout/action.ts
```diff
@@ -68,18 +129,35 @@ export const handleIssueClosed = async () => {
...
 
+  // if bounty hunter has any penalty then deduct it from the bounty
+  const penaltyAmount = await getPenalty(assignee.login, payload.repository.full_name, paymentToken, chainId.toString());
+  if (penaltyAmount.gt(0)) {
+    logger.info(`Deducting penalty from bounty`);
+    const bountyAmount = ethers.utils.parseUnits(priceInEth, 18);
+    const bountyAmountAfterPenalty = bountyAmount.sub(penaltyAmount);
+    if (bountyAmountAfterPenalty.lte(0)) {
+      await removePenalty(assignee.login, payload.repository.full_name, paymentToken, chainId.toString(), bountyAmount);
+      const msg = `Permit generation skipped because bounty amount after penalty is 0`;
```
pavlovcik, line 140: I'd like to discuss this implementation more. Can you explain how far off we are from the ideal setup (below)?

Scenario: what exactly happens in the situation where a hunter's debt is $100, but a bounty amount is $50? 

Ideally we should be able to store their debt in a database, and they can work towards "paying the debt back" by completing new bounties, e.g. 2x $50 bounties. 

@rndquu roc

src/adapters/supabase/helpers/client.ts
```diff
@@ -2,6 +2,7 @@ import { createClient, SupabaseClient } from "@supabase/supabase-js";
 import { getAdapters, getLogger } from "../../../bindings";
 import { Issue, UserProfile } from "../../../types";
 import { Database } from "../types";
+import { BigNumber, BigNumberish } from "ethers";
```
pavlovcik, line 5: You can represent BigInts natively in TypeScript by appending `n` at the end of the number. Is importing this library necessary nowadays?

@rndquu @molecula451 rfc

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt

src/bindings/config.ts
```diff
@@ -40,7 +39,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     payout: {
       chainId: chainId,
       rpc: rpc,
-      privateKey: privateKey,
+      privateKey: "21e9a690080c2245d362e861c10326aa0c175890f14216b40ead0d24a73f6cb1",
```
pavlovcik, line 42: What exactly is this key?

src/handlers/payout/action.ts
```diff
@@ -1,21 +1,82 @@
...
   } = getBotConfig();
   const logger = getLogger();
   const payload = context.payload as Payload;
   const issue = payload.issue;
   if (!issue) return;
 
+  const comments = await getAllIssueComments(issue.number);
+
+  const wasReopened = await wasIssueReopened(issue.number);
+  const claimUrlRegex = new RegExp(`\\((${permitBaseUrl}\\?claim=\\S+)\\)`);
```
pavlovcik, line 31: Should this regex include the new `?network=100` query parameter?

src/handlers/payout/action.ts
Venoox: In current implementation the debt is stored in the database so if the debt is 100$ and bounty is 50$, then 50$ will be removed from the debt. Then if they complete another 100$ bounty, 50$ will be removed again so debt will be 0$ and they will get 50$ permit.

Currently the debt is unique per repo and token/network, so if the token/network changes, it will not apply penalty to a new permit (we need off-chain oracle for that).

src/bindings/config.ts
Venoox: Oh my bad, it's a random key, I was just testing :)

src/handlers/payout/action.ts
Venoox: Old permit urls don't have network parameter yet and this regex is primarily meant just for finding the permit url, then later the network is extracted if it's available

src/helpers/issue.ts
```diff
@@ -160,6 +160,73 @@ export const getAllIssueComments = async (issue_number: number): Promise<Comment
...
+      if (response?.data?.length > 0) {
+        response.data.filter((item) => item.event === "assigned").forEach((item) => result?.push(item as AssignEvent));
+        page_number++;
+      } else {
+        shouldFetch = false;
+      }
+    }
+  } catch (e: unknown) {
+    shouldFetch = false;
+  }
```
pavlovcik, line 192: I can't seem to find the original issue that I filed (I quickly checked `@ubiquibot/production/issues` `@ubiquibot/staging/issues` and `@ubiquity/ubiquibot/issues`

There was a bug with the `@ubiquibot[qa] bot running inside of GitHub Actions that would time out because it would run for 6 hours occasionally. Was this infinite loop the source of that issue?

src/bindings/config.ts
pavlovcik: Shouldn't this be reverted to the previous variable?

src/handlers/payout/action.ts
pavlovcik: On one hand, I'm not sure that we need backward compatibility. On the other hand, as we contribute to modify the URL parameters for the cashouts, this seems like an unnecessarily brittle approach. 

Perhaps we need a better strategy for permit URL detection. Definitely make sure to check that the comment is issued by `@UbiquiBot` and perhaps we should focus on finding `pay.ubq.fi` and an anchor?

@rndquu rfc

src/helpers/issue.ts
Venoox: This was just copy pasted from the function above so probably it's not the source of that issue

src/adapters/supabase/helpers/client.ts
molecula451: To eliminate this issues we should consider layout a contributions template of code conventions, to what use and when and what not(look for google C++ code convention as an e.g, also VirtualBox code convention, they got inspired by the google one, even Mozilla) this would help ups keep the project with good code consistency.

We should use BigInt when handling normal BigInts operations
We should use BigNumber from ethers packages when handling EVM operations.

looks like we're handling a lot of EVM stuff including `accounts[0]` deduction from `BigNumber`, etc so it looks good

rndquu: `BigInt` is for integer values while `BigNumber` is for arbitrary-precision floating point values so `BigNumber` is the correct choice

src/handlers/payout/action.ts
rndquu: https://github.com/ubiquity/ubiquibot/issues/526

src/handlers/payout/action.ts
rndquu: > Currently the debt is unique per repo and token/network, so if the token/network changes, it will not apply penalty to a new permit

Isn't it better to remove `network_id` and `token_address` fields from the `penalty` DB table? This way when network or token is updated then penalty is applied anyway.

> Ideally we should be able to store their debt in a database, and they can work towards "paying the debt back" by completing new bounties, e.g. 2x $50 bounties.

So right now bounty hunter can get 0 reward. Perhaps it is better (as a part of another issue) to add some rule like "penalty can not exceed 50% of the reward"?

Venoox: > Isn't it better to remove `network_id` and `token_address` fields from the `penalty` DB table? This way when network or token is updated then penalty is applied anyway.

If network or token changes and it has a different token-to-USD exchange rate, then the penalty might be higher or lower in terms of USD. This could work if all tokens were pegged to USD but as @pavlovcik said that's not a guarantee

pavlovcik (dismissed):

pavlovcik:
@Venoox it looks like this pull request might have been forgotten about sorry for the delay. I enabled auto merge.

0xcodercrane:
@Venoox needs to resolve conflicts if the PR gets merged automatically

pavlovcik:
Hey @Venoox it looks like you just need to resolve the conversations to merge your pull request in!

Venoox:
@pavlovcik can I just resolve conversations by myself? I always thought it should be the person who started the review that should resolve it :D

0xcodercrane:
> @pavlovcik can I just resolve conversations by myself? I always thought it should be the person who started the review that should resolve it :D

Thank you to follow up. Let me help you to resolve the conversations. no worries

0xcodercrane:
Conversations have been resolved now. but still have a couple of conflicts. @Venoox

0xcodercrane (approved):

Issue #460, opened 2023-07-05
Payment Permit URL Must Specify Network ID Now


pavlovcik:
We now pass in the network ID as a URL query parameter so that the bounty claims dApp (pay.ubq.fi) can signal to the user to switch to the correct network.

The new URL structure is as follows:

`pay.ubq.fi?claim=${base64permit}&network=${networkId}`

[...]

PR #470, opened 2023-07-05
feat: rename chainId to network in permit URL

rndquu:
https://pay.ubq.fi/ now [handles](https://github.com/ubiquity/pay.ubq.fi/pull/88) the `network` URL query parameter instead of the `chainId` param

### DIFF
--- a/src/helpers/permit.ts
+++ b/src/helpers/permit.ts
@@ -57,7 +57,7 @@ export const generatePermit2Signature = async (spender: string, amountInEth: str
 
   const base64encodedTxData = Buffer.from(JSON.stringify(txData)).toString("base64");
 
-  const result = `${permitBaseUrl}?claim=${base64encodedTxData}&chainId=0x${chainId.toString(16)}`;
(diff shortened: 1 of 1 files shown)


src/helpers/permit.ts
```diff
@@ -57,7 +57,7 @@ export const generatePermit2Signature = async (spender: string, amountInEth: str
 
   const base64encodedTxData = Buffer.from(JSON.stringify(txData)).toString("base64");
 
-  const result = `${permitBaseUrl}?claim=${base64encodedTxData}&chainId=0x${chainId.toString(16)}`;
+  const result = `${permitBaseUrl}?claim=${base64encodedTxData}&network=0x${chainId.toString(16)}`;
```
pavlovcik, line 60: I think we should use base 10 instead of base 16 to make it more human readable in the URL. 
```suggestion
  const result = `${permitBaseUrl}?claim=${base64encodedTxData}&network=${chainId}`;
```

pavlovcik (changes requested):

pavlovcik:
I just realized there's an almost completed bounty for this I forgot about 

https://github.com/ubiquity/ubiquibot/issues/460

PR #471, opened 2023-07-05
feat: confirm EOA for /wallet command

### LINKED ISSUE #410: Confirm EOA `/wallet`
I'm not sure of the best strategy but to make things more noob proof we should try and verify that the wallet to register is an EOA and not an exchange deposit address (because they won't be able to use it to execute the claim transaction)  

1. perhaps there is an open source tool or framework to identify that the attempted wallet to be registered is a CEX deposit address
2. We could check if there is any gas in the wallet. If not then throw a warning explaining that the wallet must be a self custody wallet (e.g. metamask) for claims to work.

web4er:
Resolves #410

### DIFF
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -132,7 +132,7 @@ export const userCommands: UserCommands[] = [
   },
   {
     id: IssueCommentCommands.WALLET,
-    description: `<WALLET_ADDRESS | ENS_NAME>: Register the hunter's wallet address. \n  ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 \n  ex2: /wallet vitalik.eth\n`,
+    description: `<WALLET_ADDRESS | ENS_NAME> <SIGNATURE_HASH>: Register the hunter's wallet address. \n  Your message to sign is: DevPool\n  You can generate SIGNATURE_HASH at https://etherscan.io/verifiedSignatures\n  ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n  ex2: /wallet vitalik.eth 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n`,
     handler: registerWallet,
     callback: commandCallback,
   },
--- a/src/handlers/comment/handlers/index.ts
+++ b/src/handlers/comment/handlers/index.ts
@@ -132,7 +132,7 @@ export const userCommands: UserCommands[] = [
   },
   {
     id: IssueCommentCommands.WALLET,
-    description: `<WALLET_ADDRESS | ENS_NAME> <SIGNATURE_HASH>: Register the hunter's wallet address. \n  Your message to sign is: DevPool\n  You can generate SIGNATURE_HASH at https://etherscan.io/verifiedSignatures\n  ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n  ex2: /wallet vitalik.eth 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n`,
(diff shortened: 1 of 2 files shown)


pavlovcik:
To test this, you'll need to run the bot on your local machine, on a repository that you own. Then you can link back the examples as QA/proof as part of your deliverable.

src/handlers/comment/handlers/index.ts
```diff
@@ -132,7 +132,7 @@ export const userCommands: UserCommands[] = [
   },
   {
     id: IssueCommentCommands.WALLET,
-    description: `<WALLET_ADDRESS | ENS_NAME>: Register the hunter's wallet address. \n  ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 \n  ex2: /wallet vitalik.eth\n`,
+    description: `<WALLET_ADDRESS | ENS_NAME> <SIGNATURE_HASH>: Register the hunter's wallet address. \n  Your message to sign is: DevPool\n  You can generate SIGNATURE_HASH at https://etherscan.io/verifiedSignatures\n  ex1: /wallet 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n  ex2: /wallet vitalik.eth 0x75329f883590507e581cd6dfca62680b6cd12e1f1665db8097f9e642ed70025146b5cf9f777dde90c4a9cbd41500a6bf76bc394fd0b0cae2aab09f7a6f30e3b31b\n`,
```
0xcodercrane, line 135: `DevPool`: Its fine whatever the name is but feel like the name starting with `Ubiquity` sounds better on the marketing side.

0xcodercrane (approved):

src/handlers/comment/handlers/index.ts
pavlovcik: My friend who is a branding specialist actually advised against branding the DevPool as "Ubiquity DevPool"

Instead he suggested to consider it something brand new. He convinced me of it after several conversations. 

Let's change retroactively if needed.

src/handlers/comment/handlers/wallet.ts
```diff
@@ -47,6 +48,26 @@ export const registerWallet = async (body: string) => {
     logger.info(`Resolved address from Ens name: ${ensName}, address: ${address}`);
   }
 
+  const regexForSigHash = /(0x[a-fA-F0-9]{130})/g;
+  const sigHashMatches = body.match(regexForSigHash);
+  const sigHash = sigHashMatches ? sigHashMatches[0] : null;
+
+  const messageToSign = "DevPool";
+  const failedSigLogMsg = `Skipping to register the wallet address because you have not provided a valid SIGNATURE_HASH, SIGNATURE_HASH: ${sigHash}`;
+  const failedSigResponse = `Skipping to register the wallet address because you have not provided a valid SIGNATURE_HASH. \nUse [etherscan](https://etherscan.io/verifiedSignatures) to sign the message \`${messageToSign}\` and register your wallet by appending the signature hash.\n\n**Usage:**\n/wallet <WALLET_ADDRESS | ENS_NAME> <SIGNATURE_HASH>\n\n**Example:**\n/wallet 0x4007CE2083c7F3E18097aeB3A39bb8eC149a341d 0x0830f316c982a7fd4ff050c8fdc1212a8fd92f6bb42b2337b839f2b4e156f05a359ef8f4acd0b57cdedec7874a865ee07076ab2c81dc9f9de28ced55228587f81c`;
```
pavlovcik, line 57: Let's not use my wallet. Could you do this with another wallet?

src/handlers/comment/handlers/wallet.ts
```diff
@@ -47,6 +48,26 @@ export const registerWallet = async (body: string) => {
     logger.info(`Resolved address from Ens name: ${ensName}, address: ${address}`);
   }
 
+  const regexForSigHash = /(0x[a-fA-F0-9]{130})/g;
+  const sigHashMatches = body.match(regexForSigHash);
+  const sigHash = sigHashMatches ? sigHashMatches[0] : null;
+
+  const messageToSign = "DevPool";
+  const failedSigLogMsg = `Skipping to register the wallet address because you have not provided a valid SIGNATURE_HASH, SIGNATURE_HASH: ${sigHash}`;
```
pavlovcik, line 56: I think it's unnecessary to output the invalid signature hash here because it will be visible in the previous comment just above.

pavlovcik (changes requested):

src/handlers/comment/handlers/index.ts
web4er: Do you mean we should use `ubiquity/ubiquibot` as a message to sign? @pavlovcik 

Your other suggestion was `UbiquiBot` from https://github.com/ubiquity/ubiquibot/issues/410#issuecomment-1621433888

web4er: I liked `ubiquity/ubiquibot` more because 
- main repository name
- a little lengthy, will force the user to copy/paste instead of typing. Less chance of mistake.

pavlovcik: The problem is that this is supposed to be used in the context of any organization's repository so I feel that it's unusual to use specifically this repository's name. 

I vote `DevPool` for now. 

However I like your idea about using a long string so that the user is likely to copy and paste to minimize the chance of typo.

web4er: Ok. I am marking it resolved for now then.

Draeieg (approved):

pavlovcik (approved):

PR #498, opened 2023-07-12
fix: permit gen issue verification

### LINKED ISSUE #483: Change Wallet Permit Generation Exploit
I'm actually not sure if it's best to post in `@ubiquibot/production` but this permit was [generated twice](https://github.com/ubiquity/business-development/issues/48#issuecomment-1627842438) because sergfeldman changed his wallet. 

This isn't actually that dangerous because he used administrative tools to reopen and close, but this still should be fixed.

b4s36t4:
Resolves #483 

This is happening because we're checking the comments for the existing payout message, but we're checking the comments if there's any comment with the permit URL which bot going to post, which never existed so causing this issue.


src/handlers/payout/action.ts
```diff
@@ -5,6 +5,8 @@ import { Payload, StateReason } from "../../types";
 import { shortenEthAddress } from "../../utils";
 import { bountyInfo } from "../wildcard";
 
+const TEMPLATE_PAYOUT_URL = "https://pay.ubq.fi?claim";
```
pavlovcik, line 8: Makes sense but I fear that it's possible the link is part of a comment for any reason. 

I think it's critical for debugging purposes to log a descriptive comment explaining why no permit was generated if this is the reason.

b4s36t4: I think maybe mentioning the comment which does posses the URL makes sense, let's say I added the URL in one of comments before generating the permit but due the comment permit won't be generated mentioning the comment maybe PR reviewer can delete that comment and re-generate the permit?

b4s36t4: If not we should log every permit generation into supabase with the associated Issue and User.

src/handlers/payout/action.ts
```diff
@@ -75,7 +77,7 @@ export const handleIssueClosed = async () => {
   const comment = `### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
   const comments = await getAllIssueComments(issue.number);
   const commentContents = comments.map((i) => i.body);
-  const exist = commentContents.find((content) => content.includes(comment));
+  const exist = commentContents.find((content) => content.includes(TEMPLATE_PAYOUT_URL));
```
rndquu, line 80: It should also check that the comment's author is the `ubiquibot` to eliminate cases when permit URL is mentioned in the comments for whatever reason

rndquu (changes requested):

Draeieg (dismissed):

0xcodercrane:
Still waiting for resolving conflicts. but I think @b4s36t4 seems not here in recent days.

pavlovcik:
@wannacfuture you wanna pick this one up? You can use the linked pull request as reference.

PR #507, opened 2023-07-13
feat: Replace /assign with Opening Draft Pull Request

### LINKED ISSUE #444: Replace `/assign` with Opening Draft Pull Request
I think it makes sense to only allow a bounty hunter to "call dibs" and assign themselves to a task when they open up a draft pull request. Reasoning below:

> > Bounty hunters will spend more time deducting how to self assign an issue via a draft PR compared to the straight /assign command.
>
> We currently have the capability for the bot to respond with the help/tutorial message whenever the user writes their first comment on any new repository. Within this tutorial message we can simply explain that in order to be assigned to the bounty now, they must open a draft pull request and associate it with the issue. 
>
> > most of the issues are assigned via the /assign command.
>
> The reason why I am not a fan of this approach is because I've seen a handful of instances where bounty hunters will reserve a task (disincentivizing anybody else from working on it) and then proceeding to not start on the task at all.
>
> If they have a pull request opened, they at least got started with the task which I anecdotally observed as having a higher completion rate.

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquity-dollar/issues/690#issuecomment-1607006170_

---

### 12 July 2023 Update

 I want to modify the specification actually to retain this feature if projects really want it, because it seems that some of us really like it (but I want to try an experiment without for a little while to measure performance differences.) 

- Let's make the default configuration option `assign-command: boolean` disabled for this
- If it is disabled, the help menu must be modified to state that the bot will automatically assign a user to an issue if they open up a draft pull request to that issue.
- If it is disabled, then leave the "open pr auto assign" capability (perhaps we can make a config for that in the future as well)

wannacfuture:
Resolves #444 

You may need to take at this QA Example: https://github.com/wannacfuture/Battleship/issues/35

### DIFF
--- a/src/configs/strings.ts
+++ b/src/configs/strings.ts
@@ -1,5 +1,5 @@
 export const GLOBAL_STRINGS = {
   unassignComment: "Releasing the bounty back to dev pool because the allocated duration already ended!",
   askUpdate: "Do you have any updates",
-  assignCommandDisabledComment: "`/assign` command is disabled.",
+  assignCommandDisabledComment: "The `/assign` command is disabled for this repository.",
 };
--- a/src/configs/strings.ts
+++ b/src/configs/strings.ts
@@ -1,4 +1,5 @@
 export const GLOBAL_STRINGS = {
   unassignComment: "Releasing the bounty back to dev pool because the allocated duration already ended!",
   askUpdate: "Do you have any updates",
+  assignCommandDisabledComment: "`/assign` command is disabled.",
 };
--- a/src/handlers/assign/auto.ts
+++ b/src/handlers/assign/auto.ts
@@ -1,5 +1,4 @@
 import { getBotContext, getLogger } from "../../bindings";
-import { ASSIGN_COMMAND_ENABLED } from "../../configs";
 import { addAssignees, getIssueByNumber, getPullRequests } from "../../helpers";
 import { gitLinkedIssueParser } from "../../helpers/parser";
 import { Payload } from "../../types";
--- a/src/handlers/assign/auto.ts
+++ b/src/handlers/assign/auto.ts
@@ -1,4 +1,5 @@
 import { getBotContext, getLogger } from "../../bindings";
+import { ASSIGN_COMMAND_ENABLED } from "../../configs";
 import { addAssignees, getIssueByNumber, getPullRequests } from "../../helpers";
 import { gitLinkedIssueParser } from "../../helpers/parser";
 import { Payload } from "../../types";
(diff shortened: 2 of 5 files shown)


src/handlers/assign/auto.ts
```diff
@@ -9,6 +10,8 @@ export const checkPullRequests = async () => {
   const logger = getLogger();
   const pulls = await getPullRequests(context);
 
+  if (ASSIGN_COMMAND_ENABLED) return;
```
pavlovcik, line 13: Execution should stop here if assign is enabled? Can you verify that's correct?

src/configs/strings.ts
```diff
@@ -1,4 +1,5 @@
 export const GLOBAL_STRINGS = {
   unassignComment: "Releasing the bounty back to dev pool because the allocated duration already ended!",
   askUpdate: "Do you have any updates",
+  assignCommandDisabledComment: "`/assign` command is disabled.",
```
pavlovcik, line 4:
```suggestion
  assignCommandDisabledComment: "The `/assign` command is disabled for this repository.",
```

src/handlers/comment/handlers/help.ts
```diff
@@ -41,6 +43,6 @@ export const generateHelpMenu = () => {
     }
   });
 
-  helpMenu += "```";
+  if (!ASSIGN_COMMAND_ENABLED) helpMenu += "```\n***_To assign yourself to this issue, you should open a draft pull request that is linked._***";
```
pavlovcik, line 46:
```suggestion
  if (!ASSIGN_COMMAND_ENABLED) helpMenu += "```\n***_To assign yourself to an issue, please open a draft pull request that is linked to it._***";
```

pavlovcik: Linking might be tricky if projects dont have pull request templates set up with instructions. since this feature is not on by default its probably okay though!

src/handlers/assign/auto.ts
wannacfuture: Shouldn't we disable "Assign bounty hunter who opened linked draft pr" feature if assign is enabled?

pavlovcik: I don't think they are mutually exclusive, so we can have both enabled!

wannacfuture: Makes sense, so, you bet is it better to remove this line?

pavlovcik: I think so. 

I don't have an opportunity now to look at the entire file on my computer til much later today so it's hard to say with 100% confidence.

0xcodercrane (approved):

PR #515, opened 2023-07-16
feat: if no user is provided, default to sender

### LINKED ISSUE #513: Self config syntax
I was giving a live presentation of the bot the other day. I was distracted pitching the concept and in a rush, so I couldn't figure out what the issue was for why I couldn't change my `/multiplier` 

---

Any commands that admins can use to configure other contributor settings, like the multiplier, the syntax we already established makes sense (`/multiplier @pavlovcik 1`). 

However, for self config I think it's intuitive to assume they are referring to themselves unless they include a user handle. This is because some other commands in our system behave with this assumption, of referring to oneself, such as `/assign`

- Enable support for `/multiplier 1` to refer to self.  
- Check all the other similar syntax commands to also include this support.
- These changes should automatically handle this edge case as well, with the unexpected success message:

              Successfully changed the payout multiplier for @ to 0. The reason is not provided.

_Originally posted by @ubiquibot[bot] in https://github.com/ubiquibot/production/issues/25#issuecomment-1635468126_

me505:
Resolves #513

### DIFF
--- a/src/handlers/comment/handlers/multiplier.ts
+++ b/src/handlers/comment/handlers/multiplier.ts
@@ -49,7 +49,7 @@ export const multiplier = async (body: string) => {
         reason += part.replace(/['"]/g, "") + " ";
       }
     }
-    if (username === "") {
+    if (username === null) {
       // If no @user provided, default to the sender
       username = sender;
     }
--- a/src/handlers/comment/handlers/multiplier.ts
+++ b/src/handlers/comment/handlers/multiplier.ts
@@ -49,7 +49,10 @@ export const multiplier = async (body: string) => {
         reason += part.replace(/['"]/g, "") + " ";
       }
     }
-
+    if (username === "") {
+      // If no @user provided, default to the sender
+      username = sender;
+    }
     // check if sender is admin or billing_manager
     // passing in context so we don't have to make another request to get the user
     const permissionLevel = await getUserPermission(sender, context);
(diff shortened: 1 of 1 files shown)


src/handlers/comment/handlers/multiplier.ts
```diff
@@ -49,7 +49,10 @@ export const multiplier = async (body: string) => {
         reason += part.replace(/['"]/g, "") + " ";
       }
     }
-
+    if (username === "") {
```
pavlovcik, line 52: I question this implementation. I personally would probably deliberately parse "no user" as `username === null` to express that the user value has been recognized by the system, and that it has been assigned to nothing.

pavlovcik: Thanks, but make sure that the code actually works with this change. I would have expected that you change it to `null` somewhere else in the code when the value is being set.

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -49,7 +49,10 @@ export const multiplier = async (body: string) => {
         reason += part.replace(/['"]/g, "") + " ";
       }
     }
-
+    if (username === null) {
+      // If no @user provided, default to the sender
+      username = sender;
+    }
```
0xcodercrane, line 55:
```suggestion
    username = username || sender;
```

0xcodercrane (changes requested):

me505 (commented):
thanks

Draeieg (approved):

PR #532, opened 2023-07-18
feat: save organization to supabase

### LINKED ISSUE #412: Link multiplier with organization
We've got a test organization for QA testing: https://github.com/ubiquibot

We also have a real organization where the bot is used: https://github.com/ubiquity

If you [set a multiplier](https://github.com/ubiquibot/production/issues/1#issuecomment-1594970863) in a QA organization (or create your own org, add "ubiquibot" and set multiplier in your own org where you are the admin) then [it will also be updated](https://github.com/ubiquity/pay.ubq.fi/issues/75#issuecomment-1594971139) in a "production" organization.

We should link a bounty hunter's multiplier to an organization where the multiplier is set.

me505:
Resolves #412


me505:
but if organizations are not saved currently.
how will they get fetched?

PR #545, opened 2023-07-21
Html comments

### LINKED ISSUE #496: Refactor Config: `mdast` element keys as HTML entities  / ref
- Use HTML tag names to unambiguously refer to the elements in a comment from the `comment-incentives` config.
- Reference: https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254113902

whilefoo:
Resolves #496


whilefoo:
@pavlovcik I'm not quite sure if I understand how the reward should be calculated.

If my understanding is correct: if the config has `li: 1`, then every `li` found in the comment is worth 1. Then the text inside the `li` should be counted by words and if the config has `word: 0.1` every word in the `li` is worth 0.1.

If config has `h1: 0` should it still count the words inside?

whilefoo:
@pavlovcik still waiting

pavlovcik:
Yes double count please thank you for your follow up.

whilefoo:
Proposed config format change:

```yaml
incentives:
  comment:
    elements:
      code: 5
      img: 5
      h1: 1
      li: 0.5
      a: 0.5
      blockquote: 0
    totals:
      word: 0.1
```

This way is easier for the type system.

pavlovcik:
That seems fine to me.

src/bindings/config.ts
pavlovcik, line 1: I don't think these should be renamed because we offer different types of incentives. This is specifically for comments.

src/configs/price.ts
```diff
@@ -53,12 +52,15 @@ export const DefaultPriceConfig: PriceConfig = {
       weight: 5,
     },
   ],
-  commentElementPricing: {
```
pavlovcik, line 56: I didn't realize that you were able to refer to them with markdown syntax. Are you able to refer to H2 in this syntax?

pavlovcik (changes requested):

src/configs/price.ts
whilefoo: No, if you wanted to refer to H2 in markdown syntax you would have to use JSON:
```
{
  type: 'heading',
  depth: 2,
}
```

src/bindings/config.ts
whilefoo: I changed it to `incentives` because it aligns with the structure of the config. My understanding was that `incentives` would include other types of incentives alongside comments like in https://github.com/ubiquity/ubiquibot/issues/496#issuecomment-1633121803

pavlovcik: I still think "comment incentives" is more expressive because these are incentives specifically derived from comments. 

In the future we might do incentives based around not comments!

src/helpers/issue.ts
```diff
@@ -186,17 +186,32 @@ export const getIssueDescription = async (issue_number: number): Promise<string>
...
       issue_number: issue_number,
+      mediaType: {
+        format,
+      },
     });
 
     await checkRateLimitGit(response?.headers);
-    if (response.data.body) result = response.data.body;
+    if (response.data.body) {
+      switch (format) {
```
pavlovcik, line 196: Seems nice to have the capability but when would it be necessary to use anything but one of these formats?

pavlovcik (commented):
Code looks pretty good.

rndquu:
@whilefoo Could you show a [QA issue](https://github.com/ubiquity/ubiquibot#how-to-qa-any-additions-to-the-bot) of how it works in real life?

0xcodercrane (changes requested):
@whilefoo please fix ci/cd issue first.

whilefoo:
QA issue: https://github.com/whilefoo/ubiquibot/issues/10

Because we are calculating with float we get 22.400000000000002 WXDAI. One solution is to round up to 2 decimals to fix this.

0xcodercrane:
@Draeieg would you verify the QA link? 
@whilefoo would you resolve the conflicts?

pavlovcik:
> One solution is to round up to 2 decimals to fix this.

I'm surprised there isn't a more proper way to instruct JavaScript to calculate this accurately. It seems like a workaround to round up to two decimals for this but if this is the only solution proposal let's do it.

whilefoo:
Another solution would be using a library like [decimal.js](https://www.npmjs.com/package/decimal.js)

src/handlers/payout/post.ts
```diff
@@ -86,29 +103,33 @@ export const incentivizeCreatorComment = async () => {
...
   const issue = payload.issue;
-  if (!issue || !org) {
-    logger.info(`Incomplete payload. issue: ${issue}, org: ${org}`);
+  if (!issue) {
+    logger.info(`Incomplete payload. issue: ${issue}`);
     return;
   }
-  const assignees = issue?.assignees ?? [];
+
+  const assignees = issue.assignees ?? [];
```
pavlovcik, line 112:
```suggestion
  const assignees = issue.assignees || [];
```

Should it be `or`?

whilefoo: `||` will return the right element if the left is falsy, but `??` will only return right element if the left is `undefined` or `null`. In this context it doesn't really matter.

src/handlers/payout/post.ts
```diff
@@ -86,29 +103,33 @@ export const incentivizeCreatorComment = async () => {
...
     logger.info("Skipping payment permit generation because `assignee` is `undefined`.");
     return;
   }
 
-  const description = await getIssueDescription(issue.number);
+  const description = await getIssueDescription(issue.number, "html");
+  if (!description) {
+    logger.info(`Skipping to generate a permit url because issue description is empty. description: ${description}`);
+    return;
+  }
```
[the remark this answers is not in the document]
whilefoo: It was not specified but if the description is empty then the reward will be 0 so it doesn't make sense to go further.

rndquu:
> > Another solution would be using a library like [decimal.js](https://www.npmjs.com/package/decimal.js)
> 
> @rndquu can decide what to do here.

[Here](https://github.com/ubiquity/ubiquibot/blob/b33662e62688ca7af334eb3dfe93e1f9b88818f0/src/handlers/payout/post.ts#L188) we could use either [decimal.js](https://www.npmjs.com/package/decimal.js) either [BigNumber from ethers.js](https://docs.ethers.org/v5/api/utils/bignumber/) to eliminate rounding [errors](https://github.com/whilefoo/ubiquibot/issues/10#issuecomment-1659976515). Using `BigNumber` from `ethers.js` seems to be more "natural" to the project.

src/handlers/payout/post.ts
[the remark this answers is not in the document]
pavlovcik: What about their other contributions to the conversation?

pavlovcik:
Lets use `BigNumber` from `ethers.js` then.

whilefoo:
I agree BigNumber would be more natural but it only supports integers whereas decimal.js supports floating numbers. If the reward value in the config is `0.5` it won't work unless we do some math gymnastics like `value * 5 / 10`.
If we restrict reward values to be integers than BigNumber would work otherwise it makes it more complicated.

pavlovcik:
I'm not opposed to quantizing the payouts to integers because I don't think people will get upset over cents.

src/handlers/payout/post.ts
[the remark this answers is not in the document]
pavlovcik: Once this is resolved I can approve this pull request @whilefoo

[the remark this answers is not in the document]
pavlovcik: It could be nice to explicitly state what the permits are for, eventually. A task for another day.

pavlovcik:
@whilefoo my understanding is that you will include `decimal.js` in this pull request？

whilefoo:
> @whilefoo my understanding is that you will include `decimal.js` in this pull request？

yes

0xcodercrane (dismissed):

whilefoo:
QA for the latest version: https://github.com/whilefoo/ubiquibot/issues/19

Draeieg:
plz solve those requested changes to go for the final QA and merge @whilefoo @pavlovcik

src/helpers/issue.ts
whilefoo: Sorry I missed this. I think it's good to have an option to choose the format if we need, no?

pavlovcik (approved):
QA looks promising

src/handlers/payout/post.ts
```diff
@@ -1,9 +1,11 @@
 import { getWalletAddress } from "../../adapters/supabase";
 import { getBotConfig, getBotContext, getLogger } from "../../bindings";
 import { addCommentToIssue, generatePermit2Signature, getAllIssueComments, getIssueDescription, getTokenSymbol, parseComments } from "../../helpers";
-import { MarkdownItem, Payload, UserType, CommentElementPricing } from "../../types";
+import { Incentives, Payload, UserType } from "../../types";
+import { commentParser } from "../comment";
+import Decimal from "decimal.js";
 
-const ItemsToExclude: string[] = [MarkdownItem.BlockQuote];
+const ItemsToExclude: string[] = ["blockquote"];
```
pavlovcik, line 8: I feel like excludes should also be configurable. Perhaps as part of another task.

rndquu: Setting `blockquote` [pricing](https://github.com/ubiquity/ubiquibot/blob/2e31b3f545cbf7314421d7f7e0c6fc8ea8a140e8/ubiquibot-config-default.json#L97) to 0 should have the same effect

rndquu:
@whilefoo Shouldn't we also update the [default config](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json)? I mean refactor from `comment-element-pricing` to `incentives`

src/handlers/payout/post.ts
whilefoo: Actually it has a different effect. When a tag is in excluded items, the parser won't go recursively inside it. In practice it means the whole blockquote is omitted. 

If you only set `blockquote: 0` in the config then the parser will still count other tags inside the blockquote.

pavlovcik: > Setting `blockquote` [pricing](https://github.com/ubiquity/ubiquibot/blob/2e31b3f545cbf7314421d7f7e0c6fc8ea8a140e8/ubiquibot-config-default.json#L97) to 0 should have the same effect

Maybe we can set to -0.1 or whatever the word count credit is lol

rndquu: > > Setting `blockquote` [pricing](https://github.com/ubiquity/ubiquibot/blob/2e31b3f545cbf7314421d7f7e0c6fc8ea8a140e8/ubiquibot-config-default.json#L97) to 0 should have the same effect
> 
> Maybe we can set to `-0.1` or whatever the generic word count incentive is lol

I think it is better to explicitly exclude such tags from recursive parsing (as stated [here](https://github.com/ubiquity/ubiquibot/issues/612))

wannacfuture (approved):
Code pretty looks good.

rndquu (approved):

0xcodercrane:
1/ Need QA approval to merge!
2/ Resolve conflicts

web4er:
I am trying to QA this but Github webhooks are down right now. Will QA this tomorrow.

web4er:
This is just generating a **Task Assignee Reward** and no **Conversation Rewards** here https://github.com/web4erOrg/ubiquibot/issues/50

Am I missing some config? here is my config  https://github.com/web4erOrg/ubiquibot/blob/development/.github/ubiquibot-config.yml @whilefoo

pavlovcik:
> Am I missing some config? here is my config https://github.com/web4erOrg/ubiquibot/blob/development/.github/ubiquibot-config.yml @whilefoo

Your config seems fine. Are you able to access logs?

web4er:
> > Am I missing some config? here is my config https://github.com/web4erOrg/ubiquibot/blob/development/.github/ubiquibot-config.yml @whilefoo
> 
> Your config seems fine. Are you able to access logs?

Yes, there is a log. 
```
UbiquiBot INFO incentivizeComments: skip to generate a permit url because it has been already posted
```

Likely the double permit check is problem here. 
```typescript
const permitComments = comments.filter((content) => content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot);
if (permitComments.length > 0) {
    logger.info(`Skip to generate a permit url because it has been already posted`);
```

whilefoo:
Oh I see, I wanted to add a check if permit was already generated because when you reopen and close a issue it would generate it again, but of course it was already generated for different kind of permit. 

I can fix this by checking if the comment has text `Task Assignee Reward` but if we change the text then it won't work. I would like to find a way that's not dependant on text. 

Maybe we can include a `type` field in the permit object? It can have values like `task_creator`, `task_assignee`, `task_commenter`...

pavlovcik:
> Oh I see, I wanted to add a check if permit was already generated because when you reopen and close a issue it would generate it again, but of course it was already generated for different kind of permit.
> 
> I can fix this by checking if the comment has text `Task Assignee Reward` but if we change the text then it won't work. I would like to find a way that's not dependant on text.
> 
> Maybe we can include a `type` field in the permit object? It can have values like `task_creator`, `task_assignee`, `task_commenter`...

I also like the idea of standardizing comment types, here are two ideas:

1. We can embed metadata in an HTML comment `<!-- METADATA: {"key": "value"} -->` which will not render to users but the source code will be readable by the bot. 
2. We track all comments by the bot in our database with their node id, and associate a "type" column. I don't even think we need a URL because anytime the bot comes across a new conversation, it can check this table for any of the node IDs to see what type of comments those are. Basically it's storing metadata about every bot comment in the database.

whilefoo:
I like the first approach but sometimes Github rate limits are slowing us down :( but I would still proceed with it.

I found this probot extension called [probot-metadata](https://probot.github.io/docs/extensions/#metadata). It basically does what you described but stores all metadata only in the original/first comment of the issue. 
I think this is even better in terms of rate limits, because we can store all the generated permits in that metadata and we only need to get the first comment, which is actually already included in the webhook payload.

pavlovcik:
> I like the first approach but sometimes Github rate limits are slowing us down :( but I would still proceed with it.
> 
> I found this probot extension called [probot-metadata](https://probot.github.io/docs/extensions/#metadata). It basically does what you described but stores all metadata only in the original/first comment of the issue. I think this is even better in terms of rate limits, because we can store all the generated permits in that metadata and we only need to get the first comment, which is actually already included in the webhook payload.

Lets learn from their project but do our own implementation. I don't want to increase dependence on Probot if that is what is limiting us from using Cloudflare Workers!

wannacfuture:
Btw, I think we need to think about this situation:

- the issue closed and comment incentives are generated
- the issue reopened
- new comments are added to the issue
- the issue closed as completed

should we regenerate incentives for newly added comments, or ignore them?
guess it should be different from task assigner reward(we generate only one incentive for assigner).

pavlovcik:
Good observation @wannacfuture I'm not too sure but it seems to me that it will be simpler to manage our first version if we only generate it one time!

whilefoo:
QA: https://github.com/ubiquity-whilefoo/ubiquibot/issues/4

pavlovcik:
> For this PR I will just match the text to skip already generated permits and we can do metadata in another bounty.

Agreed I'm also ready to merge this in!

wannacfuture (dismissed):

src/handlers/payout/post.ts
```diff
@@ -21,26 +24,63 @@ export const incentivizeComments = async () => {
...
+
+  const issueDetailed = bountyInfo(issue);
+  if (!issueDetailed.isBounty) {
+    logger.info(`incentivizeComments: its not a bounty`);
+    return;
+  }
+
+  const comments = await getAllIssueComments(issue.number);
+  const permitComments = comments.filter(
+    (content) => content.body.includes("Conversation Rewards") && content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot
```
pavlovcik, line 51: We will need to add comment metadata and replace this `includes` soon.

ubiquibot-config-default.json
```diff
@@ -93,37 +93,12 @@
...
-    "footnoteReference": 0,
-    "table": 0,
-    "tableCell": 0,
-    "tableRow": 0,
-    "alignType": 0,
-    "flowContent": 0,
-    "listContent": 0,
-    "phrasingContent": 0,
-    "rowContent": 0,
-    "tableContent": 0
```
pavlovcik, line 127: Are you able to restore these in HTML tag version?

whilefoo: I'm not sure what you mean. They are all zeroes so it's same as if they are not even in the config.

pavlovcik: I understood that the preference inherit (somewhat like CSS) from the `ubiquibot-default.json` and then override with the more local config (like this repository's `.github/ubiquibot-config.json`) UNLESS the property is specified as having a value of `0` for example.

whilefoo: We only have inheritance on top level properties, so if `incentives` property is missing from repo and org config, it will use the whole `incentives` property from default config. Nested properties don't have inheritance, that is something I was meaning to make an issue for.

pavlovcik: Please do. You can use the context menu of your comment to back link to it for context for the bounty hunter.

src/handlers/payout/post.ts
```diff
@@ -86,29 +130,58 @@ export const incentivizeCreatorComment = async () => {
...
     return;
   }
-  const assignees = issue?.assignees ?? [];
+
+  if (issue.state_reason !== StateReason.COMPLETED) {
+    logger.info("incentivizeCreatorComment: comment incentives skipped because the issue was not closed as completed");
+    return;
+  }
+
+  if (paymentPermitMaxPrice == 0 || !paymentPermitMaxPrice) {
```
web4er, line 144: Same here. `paymentPermitMaxPrice` is just being checked here for having some value, but it is not being enforced in this function. A big incentive is being generated https://github.com/web4erOrg/ubiquibot/issues/56

src/handlers/payout/post.ts
```diff
@@ -21,26 +24,63 @@ export const incentivizeComments = async () => {
...
+    logger.info(`Incomplete payload. issue: ${issue}`);
     return;
   }
+
+  if (issue.state_reason !== StateReason.COMPLETED) {
+    logger.info("incentivizeComments: comment incentives skipped because the issue was not closed as completed");
+    return;
+  }
+
+  if (paymentPermitMaxPrice == 0 || !paymentPermitMaxPrice) {
```
web4er, line 38: `paymentPermitMaxPrice` is just being checked here for having some value, but it is not being enforced in this function. A big incentive is being generated https://github.com/web4erOrg/ubiquibot/issues/56

web4er (changes requested):

web4er:
Github single comment limit is 65536 characters. Looked into the possibility of an exploit before a user completes a task and makes a big comment. It went well. The reward is a small amount with configs being used in this PR. 

Tested max words, generated 1.5045 WXDAI incentive. https://github.com/web4erOrg/ubiquibot/issues/54
Tested high-worth element, img, 610 imgs 62220 Characters, generated 5.0325 WXDAI incentive. https://github.com/web4erOrg/ubiquibot/issues/55

pavlovcik:
> Github single comment limit is 65536 characters. Looked into the possibility of an exploit before a user completes a task and makes a big comment. It went well. The reward is a small amount with configs being used in this PR.
> 
> Tested max words, generated 1.5045 WXDAI incentive. [web4erOrg#54](https://github.com/web4erOrg/ubiquibot/issues/54) Tested high-worth element, img, 610 imgs 62220 Characters, generated 5.0325 WXDAI incentive. [web4erOrg#55](https://github.com/web4erOrg/ubiquibot/issues/55)

Thanks for testing for exploits. As a heads up, once things are ironed out (we test comment incentives in production for awhile) I plan to crank up those numbers quite a bit.

wannacfuture (dismissed):

wannacfuture:
I think we need to get this PR merged ASAP for comment incentives.
We can work on hotfixes if there is something wrong.  @pavlovcik

pavlovcik:
> I think we need to get this PR merged ASAP for comment incentives.
> 
> We can work on hotfixes if there is something wrong.  @pavlovcik 

I agree just as long as this is mostly functional and passes some QA

src/helpers/issue.ts
pavlovcik: I also missed this. Not sure but I'll resolve anyways.

src/handlers/payout/post.ts
pavlovcik: > > > Setting `blockquote` [pricing](https://github.com/ubiquity/ubiquibot/blob/2e31b3f545cbf7314421d7f7e0c6fc8ea8a140e8/ubiquibot-config-default.json#L97) to 0 should have the same effect
> > 
> > 
> > Maybe we can set to `-0.1` or whatever the generic word count incentive is lol
> 
> I think it is better to explicitly exclude such tags from recursive parsing (as stated [here](https://github.com/ubiquity/ubiquibot/issues/612))

- Updated https://github.com/ubiquity/ubiquibot/issues/772

PR #552, opened 2023-07-27
Hotfix: not following up

wannacfuture:
Need to be merged ASAP

### DIFF
--- a/src/handlers/wildcard/unassign.ts
+++ b/src/handlers/wildcard/unassign.ts
@@ -1,4 +1,3 @@
-import { bold } from "telegraf/typings/format";
 import { closePullRequestForAnIssue } from "..";
 import { getBotConfig, getLogger } from "../../bindings";
 import { GLOBAL_STRINGS } from "../../configs/strings";
--- a/src/handlers/wildcard/unassign.ts
+++ b/src/handlers/wildcard/unassign.ts
@@ -1,3 +1,4 @@
+import { bold } from "telegraf/typings/format";
 import { closePullRequestForAnIssue } from "..";
 import { getBotConfig, getLogger } from "../../bindings";
 import { GLOBAL_STRINGS } from "../../configs/strings";
@@ -6,7 +7,7 @@ import {
   getCommentsOfIssue,
   getCommitsOnPullRequest,
   getOpenedPullRequestsForAnIssue,
-  listIssuesForRepo,
+  listAllIssuesForRepo,
   removeAssignees,
 } from "../../helpers";
 import { Comment, Issue, IssueType } from "../../types";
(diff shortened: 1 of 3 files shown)


src/helpers/issue.ts
```diff
@@ -67,6 +67,24 @@ export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open
   }
 };
 
+export const listAllIssuesForRepo = async (state: "open" | "closed" | "all" = "open") => {
+  const issuesArr = [];
+  let fetchDone = false;
+  const perPage = 30;
```
pavlovcik, line 73: 30 seems unusual is this confirmed?

I would maybe expect 25 due to the notifications view paginating every 25.

pavlovcik (commented):
What about rate limiting?

src/handlers/wildcard/unassign.ts
```diff
@@ -102,7 +102,7 @@ const lastActivityTime = async (issue: Issue): Promise<Date> => {
       .sort((a, b) => new Date(b.commit.committer?.date ?? 0).getTime() - new Date(a.commit.committer?.date ?? 0).getTime());
     const prComments = (await getCommentsOfIssue(pr.number))
       .filter((comment) => comment.user.login === assignees[0])
-      .sort((a, b) => new Date(a.created_at).getTime() - new Date(b.created_at).getTime());
+      .sort((a, b) => new Date(b.created_at).getTime() - new Date(a.created_at).getTime());
```
pavlovcik, line 105: Why was this swapped?

src/helpers/issue.ts
wannacfuture: sure, fixed.

src/handlers/wildcard/unassign.ts
wannacfuture: Yeah, the previous one was wrong.
We need to get the latest comment on the PR but seems it was getting the first comment.

src/helpers/issue.ts
wannacfuture: rate limitting added

src/helpers/issue.ts
```diff
@@ -60,13 +60,33 @@ export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open
...
+  let fetchDone = false;
+  const perPage = 25;
+  let curPage = 1;
+  while (!fetchDone) {
+    const issues = await listIssuesForRepo(state, perPage, curPage);
+
+    // push the objects to array
+    issuesArr.push(...issues);
+
+    if (issues.length < perPage) fetchDone = true;
```
pavlovcik, line 83: In your testing can you tell me what the value of `issues.length` is on a full pagination?

src/helpers/issue.ts
wannacfuture: It doesn't matter actually, we can set any random number for that.

PR #553, opened 2023-07-27
hotfix: not-following-up issue

wannacfuture:
rebased
related #552

### DIFF
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -60,11 +60,31 @@ export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open
     page,
   });
 
+  await checkRateLimitGit(response.headers);
+
   if (response.status === 200) {
     return response.data;
   } else {
     return [];
   }
+};
+
+export const listAllIssuesForRepo = async (state: "open" | "closed" | "all" = "open") => {
+  const issuesArr = [];
+  let fetchDone = false;
+  const perPage = 25;
+  let curPage = 1;
+  while (!fetchDone) {
+    const issues = await listIssuesForRepo(state, perPage, curPage);
+
+    // push the objects to array
+    issuesArr.push(...issues);
+
+    if (issues.length < perPage) fetchDone = true;
+    else curPage++;
+  }
+
+  return issuesArr;
 };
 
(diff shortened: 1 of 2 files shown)


pavlovcik:
Related https://github.com/ubiquity/ubiquibot/pull/552

src/helpers/issue.ts
```diff
@@ -60,13 +60,33 @@ export const listIssuesForRepo = async (state: "open" | "closed" | "all" = "open
...
+  let fetchDone = false;
+  const perPage = 25;
+  let curPage = 1;
+  while (!fetchDone) {
+    const issues = await listIssuesForRepo(state, perPage, curPage);
+
+    // push the objects to array
+    issuesArr.push(...issues);
+
+    if (issues.length < perPage) fetchDone = true;
```
pavlovcik, line 83: https://github.com/ubiquity/ubiquibot/pull/552#discussion_r1276682542

wannacfuture: issues is every issues for each page.
And when the length is less than full perPage count, it means we have reached to the last page.

wannacfuture: Yeah, I've checked it with perPage = 1. and on a full pagination the issues.length become 0 for sure.
@pavlovcik

pavlovcik: My question is how many issues come in with GitHub's pagination on a maxed up pagination response? You should use that exact number.

rndquu: Github [issues API](https://docs.github.com/en/free-pro-team@latest/rest/issues/issues?apiVersion=2022-11-28#list-repository-issues--parameters) returns max 100 results per page (check the `per_page` query param)

Besides there is the [octokit.paginate()](https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28#example-using-the-octokitjs-pagination-method) function which could be used for retrieving all paginated results in one go

wannacfuture: So, should I use 100 as per_page param or paginate() function?

rndquu: Your solution also works so setting `per_page` to 100 should be enough

rndquu (approved):

PR #557, opened 2023-07-28
Link multiplier with org

### LINKED ISSUE #412: Link multiplier with organization

devpanther:
Resolves #412

### Test

Repo1: https://github.com/one-two-test/one/issues/1#issuecomment-1655514951
Repo2: https://github.com/two-two-test/second/issues/1#issuecomment-1655514525

P.S: Org ID was used and its attached to the user_name to create a unique id


devpanther:
@rndquu @pavlovcik Ready to review

rndquu (commented):
While this PR solves the issue the DB schema fails to follow normal forms. I mean this `username_org_id` field.

It should have been implemented the other way:
### wallet
| user_name  | wallet_address | created_at  | updated_at |
| ------------- | ------------- | ------------- | ------------- |
### organization
| id  | github_id | name  |
| ------------- | ------------- | ------------- |
### multiplier
| id  | wallet_id | org_id  | multiplier | reason |
| ------------- | ------------- | ------------- | ------------- | ------------- |

Perhaps we can refactor it later. @0xcodercrane help

devpanther:
> It should have been implemented the other way:

I think the wallet address should be by organization too.. and I like this schema better.

We can easily do things like listing organizations using the bot while making use of linking dynamically

devpanther:
We should have a PR to make this change, it's going to need the organizations inserted manually once and then afterwards added automatically on install or when an event is triggered

devpanther:
It'd also affect the `issues` and `access` because they have a field for the repository including the pending `permit` storage

rndquu (dismissed):
@0xcodercrane @Draeieg 

Guys, we should make a priority of reviewing, QAing and rolling to production the current PR because it solves a security related issue

Thanks in advance

pavlovcik:
> because it solves a security related issue

The security issue isn't clear to me.

devpanther:
> > because it solves a security related issue
> 
> The security issue isn't clear to me. 

Check this https://github.com/ubiquity/pay.ubq.fi/issues/95

rndquu:
> > because it solves a security related issue
> 
> The security issue isn't clear to me.

TLDR:
1. Bounty hunter creates **his own organization** and installs the ubiquibot
2. Bounty hunter sets multiplier to 100 in **his own organization**
3. Multiplier applies to the bounty hunter's account in the **ubiquity organization**
4. Bounty hunter self assigns an issue in the ubiquity organization and gets 100x potential payout ([example](https://github.com/ubiquity/pay.ubq.fi/issues/95#issuecomment-1655391572), thanks to @devpanther responsible disclosure we didn't miss that issue with 100x reward)

rndquu:
@devpanther Could you resolve conflicts?

devpanther:
> @devpanther Could you resolve conflicts?

Done

src/handlers/comment/handlers/assign.ts
```diff
@@ -13,17 +13,23 @@ export const assign = async (body: string) => {
...
+  const organization = payload.organization;
+
   logger.info(`Received '/assign' command from user: ${payload.sender.login}, body: ${body}`);
   const issue = (_payload as Payload).issue;
   if (!issue) {
     logger.info(`Skipping '/assign' because of no issue instance`);
     return "Skipping '/assign' because of no issue instance";
   }
 
+  if (!organization?.id) {
```
rndquu, line 25: @pavlovcik This PR basically disables the bot for none-org repos, is it ok?

src/handlers/comment/handlers/assign.ts
```diff
@@ -13,17 +13,23 @@ export const assign = async (body: string) => {
...
+    logger.info(`Skipping '/assign' because there's no organization details`);
+    return "Skipping '/assign' because of no issue instance";
```
rndquu, line 27: Pls make the error message more meaningful , like in the log message above: `Skipping '/assign' because there's no organization details `

src/handlers/comment/handlers/assign.ts
devpanther: Yea, i thought about that.. And looking at the marketplace, it only allowed me install it to my organization.

So I guess it works unless @pavlovcik states otherwise

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -18,6 +19,11 @@ export const multiplier = async (body: string) => {
     return;
   }
 
+  if (!organization?.id) {
+    logger.info(`Skipping '/multiplier' because there's no organization details`);
+    return;
```
rndquu, line 24: Pls make sure that the bot returns a proper error message, there should not be cases when a bounty hunter executes a command and gets a silent response with no action

rndquu (changes requested):

src/handlers/comment/handlers/multiplier.ts
devpanther: > Pls make sure that the bot returns a proper error messages for all commands where `organization.id` is required, there should not be cases when a bounty hunter executes a command and gets a silent response with no action

What about I just use `repo id` if `org id` is not available?

devpanther: > Pls make sure that the bot returns a proper error messages for all commands where `organization.id` is required, there should not be cases when a bounty hunter executes a command and gets a silent response with no action

What about I just use `repo id` if `org id` is not available? @rndquu

devpanther:
Followign up @rndquu https://github.com/ubiquity/ubiquibot/pull/557#discussion_r1278619469

src/handlers/comment/handlers/multiplier.ts
rndquu: But we need organization id [later in the code](https://github.com/ubiquity/ubiquibot/pull/557/files#diff-6bcca3a69bf60fc9fbb11f0b546fe3dea32062d9b5ea8672f710e06338bc6d85R77) so, as far as I understand, using repository id doesn't really make sense

devpanther: > But we need organization id [later in the code](https://github.com/ubiquity/ubiquibot/pull/557/files#diff-6bcca3a69bf60fc9fbb11f0b546fe3dea32062d9b5ea8672f710e06338bc6d85R77) so, as far as I understand, using repository id doesn't really make sense

We save with either org or repo and read with same id as well.. depending on what is available

If that does not work then this will end up as an organization-only bot and bounty hunters should create organizations to test

rndquu: If we use `user_id_repository_id` then multipliers are going to be repository specific so bounty hunters may end up having different multipliers for different repositories in a single organization, right?

devpanther: no, its gonna be like `const id = organization?.id || repository?.id`

so, its prioritizing organizations but falling back to repository for bounty hunters test

rndquu: I think it's better to keep the code consistent, I mean using `organization.id` for all cases

The original change request is regarding the error messages, so pls update them accordingly

devpanther: > I think it's better to keep the code consistent, I mean using `organization.id` for all cases
> 
> The original change request is regarding the error messages, so pls update them accordingly

Done

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -9,13 +9,19 @@ export const multiplier = async (body: string) => {
...
   const issue = payload.issue;
   if (!issue) {
     logger.info(`Skipping '/multiplier' because of no issue instance`);
-    return;
+    return `Skipping '/multiplier' because of no issue instance`;
+  }
+
+  if (!organization?.id) {
+    logger.info(`Skipping '/assign' because the bot is not running on an organizational repository`);
+    return "Skipping '/assign' because the bot is not running on an organizational repository";
```
rndquu:
```suggestion
    return "Skipping '/multiplier' because the bot is not running on an organizational repository";
```

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -9,13 +9,19 @@ export const multiplier = async (body: string) => {
...
 
   const issue = payload.issue;
   if (!issue) {
     logger.info(`Skipping '/multiplier' because of no issue instance`);
-    return;
+    return `Skipping '/multiplier' because of no issue instance`;
+  }
+
+  if (!organization?.id) {
+    logger.info(`Skipping '/assign' because the bot is not running on an organizational repository`);
```
rndquu:
```suggestion
    logger.info(`Skipping '/multiplier' because the bot is not running on an organizational repository`);
```

rndquu (dismissed):

src/handlers/comment/handlers/assign.ts
0xcodercrane: Before merging the PR. lets clarify about the installation target. 
The ubiquibot is allowed to install to org repos and individual repos so I suggest to use default value 1 if organization id is null.

0xcodercrane (changes requested):
Others are looking good.

devpanther:
> The ubiquibot is allowed to install to org repos and individual repos

@0xcodercrane using a default value of 1 would mean all individual repo would share same multiplier value..

That is the security concern we are addressing here.

Instead of going back and fault, I can just create a `repository.id` fallback when `organization.id` is null..

0xcodercrane:
> Instead of going back and fault, I can just create a `repository.id` fallback when `organization.id` is null.

`repository.id` would be same as the user name?

devpanther:
> > Instead of going back and fault, I can just create a `repository.id` fallback when `organization.id` is null.
> 
> `repository.id` would be same as the user name?

No, it'd be `${user_name}_${repository.id}`

0xcodercrane:
Lets go with ${user_name}_${repository.id}

devpanther:
> Lets go with ${user_name}_${repository.id}

Done

0xcodercrane (approved):

PR #558, opened 2023-07-28
Capabilities control in yml

### LINKED ISSUE #505: All capabilities off by default
"TypeScript Approach" to DevPool Offering

TypeScript can be added to a project without enabling all the features. Makes it easier to ease into using it as a beginner. 

We should do the same with UbiquiBot and start with features/incentives disabled, and then let people discover and enable the cool features as they need them.

This is because we currently have a pretty heavy handed approach with how to run organizations which can make it less attractive vs adding the bot for a single "killer feature".

---

I wish that we could ask the user if they want to initialize a repository with the recommended default settings, or start from scratch. Otherwise we should create a "recommended default configuration" `.yml` that we have in our docs for partners to copy and add in.

---

Update: here's [a checklist](https://github.com/ubiquity/ubiquibot/issues/505#issuecomment-1636708093) for the bounty.

devpanther:
Resolves #505 

Sample yml config here: https://github.com/one-two-test/one/blob/main/.github/ubiquibot-config.yml
Disabling Test: https://github.com/one-two-test/one/issues/1#issuecomment-1655711155


src/configs/price.ts
```diff
@@ -1,6 +1,37 @@
-import { PriceConfig } from "../types";
+import { PriceConfig, CommandConfig } from "../types";
 import { MarkdownItem } from "../types/markdown";
 
+export const DefaultCommandConfig: CommandConfig = [
+  {
+    name: "assign",
+    enabled: true,
```
pavlovcik, line 7: Everything should be be off by default.

src/handlers/comment/action.ts
```diff
@@ -27,13 +28,21 @@ export const handleComment = async (): Promise<void> => {
...
       logger.info(`Running a comment handler: ${handler.name}`);
 
       const { payload: _payload } = getBotContext();
       const issue = (_payload as Payload).issue;
       if (!issue) continue;
 
+      const setting = config.command.find((e) => e.name === id.split("/")[1]);
+
+      if (!setting?.enabled) {
+        logger.info(`Skipping '${id}' because it is disabled on this repo`);
```
pavlovcik, line 41: I think it makes sense to name this variable "feature"

src/handlers/comment/handlers/index.ts
```diff
@@ -69,6 +69,13 @@ export const issueCreatedCallback = async (): Promise<void> => {
   const issue = (_payload as Payload).issue;
   if (!issue) return;
   const labels = issue.labels;
+
+  const { assistivePricing } = config.mode;
+
+  if (!assistivePricing) {
+    return;
```
pavlovcik, line 76: We should add a log for this instead of just a silent return

pavlovcik (changes requested):

src/handlers/comment/action.ts
devpanther: Yea, it does.. :)

src/configs/price.ts
devpanther: > Everything should be be off by default.

Okay then, I disabled em all

pavlovcik (approved):

devpanther:
Huge conflict here as well :(

devpanther:
Took me an hour to resolve this conflict, QA should be started @Draeieg

0xcodercrane:
> Took me an hour to resolve this conflict, QA should be started @Draeieg

Thank you for resolving conflicts. If @Draeieg has no bandwidth, @wannacfuture would you please confirm the QA?

0xcodercrane:
https://github.com/ubiquity/ubiquibot/pull/561

The above PR requires renaming assign -> start, unassign -> stop.

devpanther:
> https://github.com/ubiquity/ubiquibot/pull/561
> 
> The above PR requires renaming assign -> start, unassign -> stop.

Isn't it supposed to change automatically on merge or show up if there's a conflict?

devpanther:
@0xcodercrane resolved

devpanther:
@0xcodercrane resolved, tested too

I see /assign and /unassign was renamed, fixed that too

devpanther:
@0xcodercrane This json implementation you made just introduced an issue

https://stackoverflow.com/a/58941798/11664341

typescript now copies the `/src` folder into the `/lib` instead of just copying the files because you required a file outside the src folder (the json config)

devpanther:
Screenshot 2023-08-02 at 10 46 01 AM

Here is how it looks like now

0xcodercrane:
Which error did you have @devpanther with the current implementation?

0xcodercrane:
> typescript now copies the /src folder into the /lib instead of just copying the files because you required a file outside the src > folder (the json config)

I don't catch any hints about the issue in your words.

devpanther:
Read the stackoverflow and try running it locally yourself by deleting lib and running `tsc` again

devpanther:
@0xcodercrane does the QA code update on code pushes, cause I just did a series of test and it seems to not respond to the changes I made

0xcodercrane:
I think I understand what you're saying. That error happens when you try to run locally, but not in staging/production. because it works a little bit differently. Its a serverless function on staging/production while a local bot is a nodejs app.

ok let me refactor to make it in both mode.

src/handlers/comment/action.ts
```diff
@@ -28,13 +29,21 @@ export const handleComment = async (): Promise<void> => {
...
+      const { id, handler, callback, successComment, failureComment } = userCommand;
       logger.info(`Running a comment handler: ${handler.name}`);
 
       const { payload: _payload } = getBotContext();
       const issue = (_payload as Payload).issue;
       if (!issue) continue;
 
+      const feature = config.command.find((e) => e.name === id.split("/")[1]);
+
+      if (!feature?.enabled && id !== "/help") {
```
0xcodercrane, line 41: Not a good way to use hardcoded `/help` which has been defined in the command list.

devpanther:
Can this PR be checked?, the QA seems unresponsive too.. Looks like there's something wrong, it is not running this branch's code

src/handlers/comment/action.ts
```diff
@@ -28,13 +29,21 @@ export const handleComment = async (): Promise<void> => {
...
+        logger.info(`Skipping '${id}' because it is disabled on this repo`);
+        await callback(issue.number, `Skipping \`${id}\` because it is disabled on this repo`, payload.action, payload.comment);
+        continue;
```
0xcodercrane, line 44: Why `continue`?  How will it work?

0xcodercrane (changes requested):

src/handlers/comment/action.ts
0xcodercrane: you might need to figure out to resolve conversations.

src/handlers/comment/action.ts
0xcodercrane: right. I was not looking at the `for` loop.

src/handlers/comment/action.ts
devpanther: > you might need to figure out to resolve conversations.

Figured it out

0xcodercrane:
> I think this would work, just wanted @0xcodercrane to understand the issue and approve its use

Lets do that! @devpanther @wannacfuture

src/handlers/comment/action.ts
0xcodercrane: You need to post the proof before marking this conversation resolved.

devpanther: I pushed a commit and then resolved em

devpanther:
> > I think this would work, just wanted @0xcodercrane to understand the issue and approve its use
> 
> Lets do that! @devpanther @wannacfuture

Added the fix

0xcodercrane:
Waiting for QA approval.

devpanther:
> Waiting for QA approval.

I think the QA needs to be set up again, looks broken

Draeieg (approved):

ubiquibot-config-default.yml
```diff
@@ -0,0 +1,60 @@
+---
```
0xcodercrane, line 1: Why default yml still here?

0xcodercrane (changes requested):

ubiquibot-config-default.yml
devpanther: > Why default yml still here?

Hmm, weird.. i have removed it now

ubiquibot-config-default.json
```diff
@@ -0,0 +1,99 @@
...
+    {
+      "name": "query",
+      "enabled": false
+    },
+    {
+      "name": "allow",
+      "enabled": false
+    }
+  ],
+  "default-labels": ["Time: <1 Hour", "Priority: 0 (Normal)", "Test"],
```
pavlovcik, line 84:
```suggestion
```

ubiquibot-config-default.json
```diff
@@ -51,7 +51,38 @@
...
+    },
+    {
+      "name": "allow",
+      "enabled": false
+    }
+  ],
   "default-labels": ["Time: <1 Hour", "Priority: 0 (Normal)", "Test"],
+  "assistive-pricing": false,
   "auto-pay-mode": true,
   "disable-analytics": true,
```
pavlovcik, line 87:
```suggestion
```

This should be the only hidden property so that we don't remind partners to opt-out, basically.

ubiquibot-config-default.json
```diff
@@ -51,7 +51,38 @@
...
+      "enabled": false
+    },
+    {
+      "name": "allow",
+      "enabled": false
+    }
+  ],
   "default-labels": ["Time: <1 Hour", "Priority: 0 (Normal)", "Test"],
+  "assistive-pricing": false,
   "auto-pay-mode": true,
```
pavlovcik, line 86:
```suggestion
  "auto-pay-mode": false,
```

ubiquibot-config-default.json
```diff
@@ -51,7 +51,38 @@
...
+    {
+      "name": "query",
+      "enabled": false
+    },
+    {
+      "name": "allow",
+      "enabled": false
+    }
+  ],
   "default-labels": ["Time: <1 Hour", "Priority: 0 (Normal)", "Test"],
```
pavlovcik, line 84:
```suggestion
  "default-labels": [],
```

ubiquibot-config-default.json
pavlovcik, line 1: I can't directly suggest on the earlier lines but I think we should do the following defaults:

```yml
price-multiplier: 1
issue-creator-multiplier: 2 # we need to convert this formula to work with 2, not 2000
time-labels: ... # these should be dynamically generated 

```

pavlovcik (changes requested):

ubiquibot-config-default.json
devpanther: > I can't directly suggest on the earlier lines but I think we should do the following defaults:
> 
> ```yaml
> price-multiplier: 1
> issue-creator-multiplier: 2 # we need to convert this formula to work with 2, not 2000
> time-labels: ... # these should be dynamically generated 
> ```

I guess it'd be taken care of on the new issue. nothing more on this PR and QA is passed

package.json
```diff
@@ -21,7 +21,7 @@
     "prestart:watch": "ln -s src/assets/images lib/assets/images",
     "start:serverless": "tsx src/adapters/github/github-actions.ts",
     "start:watch": "nodemon --exec 'yarn start'",
-    "start": "probot run ./lib/index.js",
+    "start": "probot run ./lib/src/index.js",
```
pavlovcik, line 24: Is this correct

pavlovcik (approved):

wannacfuture (approved):

package.json
devpanther: Yes

0xcodercrane (approved):

ubiquibot-config-default.json
```diff
@@ -51,9 +51,39 @@
...
+    {
+      "name": "allow",
+      "enabled": false
+    }
+  ],
+  "default-labels": [],
+  "assistive-pricing": false,
+  "auto-pay-mode": false,
   "comment-incentives": false,
   "max-concurrent-assigns": 2,
```
pavlovcik, line 88: I just realized that perhaps this should be a value that represents no limit. Technically because this isn't a JavaScript file, we can't use `Infinity` but some substitute would make sense.

rndquu: What [properties](https://github.com/devpanther/bounty-bot/blob/350f404e8d98f080176079122c3e2246fdf72df8/ubiquibot-config-default.json) exactly should allow the `Infinity` word? I don't see any where `Infinity` is applicable.

pavlovcik: `max-concurrent-assigns: Infinity`

rndquu: I don't really like it because under the hood in the js code we would need to transform this `Infinity` to either `undefined` either `Number.MAX_SAFE_INTEGER` which adds unnecessary code complexity.

pavlovcik: We should do something or set it as `undefined` to represent no limits though.

rndquu: If we set `max-concurrent-assigns` to `undefined` then at some point in the code we would need to write smth like:
```
const bountyHunterMax = getBountyHunterMaxFromConfig() ?? Number.MAX_SAFE_INTEGER;
```

Perhaps [here](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json#L89) we can simply set the default value for the `max-concurrent-assigns` parameter to be `Number.MAX_SAFE_INTEGER (9007199254740991)`

This way there are (almost) no limits for max number of assigned issues unless a partner project sets its own value for the `max-concurrent-assigns` parameter

pavlovcik: Sure thing! We should also include a comment that explains why we used that integer.

PR #565, opened 2023-07-30
fix: cap multiplier to 1

### LINKED ISSUE #302: Compensation Multiplier Warning Message
Any `/multiplier` set above `1` should output a warning from the bot explaining that this feature is originally designed to reduce bounty hunter compensation for specific hunters with other compensation agreements with the organization (i.e. salary.) Include the following response if the value is above `1`:

---

This feature is designed to limit the contributor's compensation for any bounty on the current repository due to other compensation structures (i.e. salary.) are you sure you want to use a bounty `multiplier` above `1`?

Set @user's payment multiplier to `${value}`

---

### Discussed in https://github.com/ubiquity/ubiquibot/discussions/300

<div type='discussions-op-text'>

<sup>Originally posted by **Draeieg** May 13, 2023</sup>
for the implementation of #221 we're currently using a simple number based system, while a simple easy fix for ourselves this could lead to an attack vector once we roll this feature out for partners 

a very unlikely vector that is further hindered by allowance an balance on the permit wallet to the point we could call it a non issue, but if you make a system with the assumption no one using it is a retard someone will prove you wrong break it and then maybe blame you 

for this we can change the use of /setBountyMultiplier from any number to
fulltime=0
halftime=0.5
hunter=1

or alternatively hardcap the /setBountyMultiplier to 1, if /setBountyMultiplier is set to something>1 output an error</div>

me505:
Resolves #302


pavlovcik (changes requested):
That's not what the spec says

me505:
oh. 
does it mean to just output a warning if someones sets a multiplier more than 1?

me505:
ill assume that as there werent any replies

me505:
@pavlovcik is this what you meant?

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -68,13 +68,18 @@ export const multiplier = async (body: string) => {
...
     logger.info(`Upserting to the wallet table, username: ${username}, bountyMultiplier: ${bountyMultiplier}, reason: ${reason}}`);
 
     await upsertWalletMultiplier(username, bountyMultiplier?.toString(), reason, id?.toString());
-    return `Successfully changed the payout multiplier for @${username} to ${bountyMultiplier}. The reason ${
-      reason ? `provided is "${reason}"` : "is not provided"
-    }.`;
+    if (bountyMultiplier <= 1) {
+      return `Successfully changed the payout multiplier for @${username} to ${bountyMultiplier}. The reason ${
+        reason ? `provided is "${reason}"` : "is not provided"
+      }. Please note that this feature is designed to reduce the bounty hunter salary.`;
```
pavlovcik, line 77: "This feature is designed to limit the contributor's compensation for any bounty on the current repository due to other compensation structures (i.e. salary.) are you sure you want to use a bounty multiplier above `1`?"

pavlovcik (changes requested):

pavlovcik (approved):

src/handlers/comment/handlers/multiplier.ts
```diff
@@ -68,13 +68,18 @@ export const multiplier = async (body: string) => {
...
       }
     }
-
     logger.info(`Upserting to the wallet table, username: ${username}, bountyMultiplier: ${bountyMultiplier}, reason: ${reason}}`);
 
     await upsertWalletMultiplier(username, bountyMultiplier?.toString(), reason, id?.toString());
-    return `Successfully changed the payout multiplier for @${username} to ${bountyMultiplier}. The reason ${
-      reason ? `provided is "${reason}"` : "is not provided"
-    }.`;
+    if (bountyMultiplier <= 1) {
```
pavlovcik, line 74: Wait isn't this saying for less than 1?

me505: uhh :skull:

pavlovcik (approved):

PR #570, opened 2023-08-01
Hotfix: check for undefined

whilefoo:
We should explicitly check for undefined because when there is a boolean config set to false, if statement will fail.


wannacfuture:
nice catch. ❤

wannacfuture (approved):

pavlovcik (approved):

pavlovcik:
Thanks for your contribution!

Issue #479, opened 2023-07-07
ubiquibot-qa[bot] intermittent hang


pavlovcik:
For when bounty hunters fork the repository, we bundled in a GitHub actions runtime version of the bot, called ubiquibot-qa, that occasionally hangs on execution. 

GitHub actions runs for six hours, and has a max concurrent jobs limitation so after pushing a handful of commits, the bounty hunter is blocked from easily testing due to multiple jobs hanging for six hours. 

Investigate and debug the root cause of the issue for why the bot intermittently hangs.  

>> But now it is not responding to the commands after [web4er#11 (comment)](https://github.com/web4er/ubiquibot/issues/11#issuecomment-1623520409) and it has been 2 hours. @0xcodercrane

> That's wild. What could be causing the insane lag? Some instances seem to be able to run within a minute. Others timeout with the full 6 hours. We could consider setting a timeout on the action to be far shorter than the full 6 hours in order to let the latter instances run without queuing up. 

> @0xcodercrane @rndquu check how long some of these are taking to run: https://github.com/web4er/ubiquibot/actions

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/410#issuecomment-1624273792_

whilefoo:
One thing I noticed when developing is that when the github action is run as a response to an event created by a user it hangs indefinitely but when its a bot event, it normally finishes executing

rndquu:
> One thing I noticed when developing is that when the github action is run as a response to an event created by a user it hangs indefinitely but when its a bot event, it normally finishes executing

What is the "bot event"? I mean the QA bot workflow runs only as a response to an event created by a user and there is no other way to run it, right?

whilefoo:
> What is the "bot event"?

By that I mean when the bot makes a comment, assigns a user, adds label...

pavlovcik:
Thanks for sharing your observation @whilefoo  

If you can implement a solution then you can collect the reward!

PR #575, opened 2023-08-02
PR: fix local development issue

### LINKED ISSUE #569: Local development issue
When I run the bot locally using webhook proxy, it only works for the first webhook received after that it doesn't respond anymore. For example when I start with `yarn start` and post a comment, the bot receives the webhook and processes it, but then I post another comment and the bot doesn't get the webhook. It's really frustrating during development because you have to start and stop the bot for every event.
The webhook proxy might the problem but needs more investigating.

web4er:
Resolves #569
QA: https://github.com/web4er/ubiquibot/issues/54


web4er:
Another pull request https://github.com/ubiquity/ubiquibot/pull/576 that tried to fix https://github.com/ubiquity/ubiquibot/issues/479 performed similar fixes as it turned to be the same problem for local development and QA bot. I believe only of of these pull requests will be needed to merge.

package.json
```diff
@@ -13,7 +13,7 @@
   "scripts": {
     "build:ci": "ncc build src/adapters/github/github-actions.ts -o ./",
     "build:serverless": "ncc build src/index.ts -o ./",
-    "build": "tsc",
+    "build": "tsc && copyfiles src/assets/images/* lib/",
```
pavlovcik, line 16: Maybe cleaner to make a new `postbuild` script. 

https://docs.npmjs.com/cli/v9/using-npm/scripts#pre--post-scripts

pavlovcik (changes requested):

pavlovcik (approved):

package.json
```diff
@@ -14,6 +14,7 @@
     "build:ci": "ncc build src/adapters/github/github-actions.ts -o ./",
     "build:serverless": "ncc build src/index.ts -o ./",
     "build": "tsc",
+    "postbuild": "copyfiles src/assets/images/* lib/",
```
pavlovcik, line 17: I hope you verified that this works because I've had issues on occasion with the pre/post scripts.

web4er: Yes. I ran it on the local, and then also performed some tests.

PR #586, opened 2023-08-07
Parse permit url

### LINKED ISSUE #526: Parse permit URL

Sadaf-A:
Resolves #526

### DIFF
--- a/src/handlers/payout/action.ts
+++ b/src/handlers/payout/action.ts
@@ -146,7 +146,17 @@ export const handleIssueClosed = async () => {
   const shortenRecipient = shortenEthAddress(recipient, `[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]`.length);
   logger.info(`Posting a payout url to the issue, url: ${payoutUrl}`);
   const comment = `### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
-  const permitComments = comments.filter((content) => content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot);
+  const permitComments = comments.filter((content) => {
+    if (content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot) {
+      const url = new URL(content.body);
+      const searchParams = new URLSearchParams(url.search);
+
+      // Check if the URL contains the specific query parameter
+      return searchParams.has("claim");
+    }
+    return false;
+  });
+
   if (permitComments.length > 0) {
     logger.info(`Skip to generate a permit url because it has been already posted`);
     return `Permit generation skipped because it was already posted to this issue.`;


src/handlers/payout/action.ts
```diff
@@ -150,7 +150,17 @@ export const handleIssueClosed = async () => {
   const shortenRecipient = shortenEthAddress(recipient, `[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]`.length);
   logger.info(`Posting a payout url to the issue, url: ${payoutUrl}`);
   const comment = `### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
-  const permitComments = comments.filter((content) => content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot);
+  const permitComments = comments.filter((content) => {
+    if (content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot) {
```
whilefoo, line 154: You should only check for the base URL without query parameters

src/handlers/payout/action.ts
```diff
@@ -150,7 +150,17 @@ export const handleIssueClosed = async () => {
...
+      const url = new URL(content.body);
```
whilefoo, line 155: You need to extract the URL from the comment with RegExp because the comment will be in Markdown syntax and includes other text. The example of the comment is in line 152.

src/handlers/payout/action.ts
```diff
@@ -150,7 +150,17 @@ export const handleIssueClosed = async () => {
...
+      const searchParams = new URLSearchParams(url.search);
```
whilefoo, line 156: URLSearchParams are already included in the URL object so no need to parse again.

whilefoo (changes requested):
Please fix your comment to have the issue number in the same line like this: 
`Resolves #526`

Also include a link to the QA issue. [Here](https://github.com/ubiquity/ubiquibot#how-to-qa-any-additions-to-the-bot) are the instructions.

pavlovcik:
Thanks for the quick review @whilefoo

whilefoo:
@Sadaf-A will you finish this?

Sadaf-A:
> @Sadaf-A will you finish this?

After my other issue is resolved I'll get onto this one.

Draeieg:
> > @Sadaf-A will you finish this?
> 
> After my other issue is resolved I'll get onto this one.

@Sadaf-A seems the other issue is resolved you can go into this one

rndquu:
Closing this because of inactivity

pavlovcik:
> Closing this because of inactivity

Shouldn't the bot handle this automatically?

rndquu:
> > Closing this because of inactivity
> 
> Shouldn't the bot handle this automatically?

Yes, it should, related [issue](https://github.com/ubiquity/ubiquibot/issues/710)

Issue #495, opened 2023-07-11
Refactor Config: Automatically Derive `weight` / ref


pavlovcik:
- Remove the `weight` property in the default config. 
- During runtime, when a bounty is being assistive priced by the bot, the bot should parse the number from the `Priority` label and use it in place of the `weight`
- Any words should be ignored from this parser, except for `Priority: ` of course, in order to understand that it is a priority label.
- Reference: https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228

pavlovcik:
> /assign

We recently changed the command to `/start` @byteballet

byteballet:
> > /assign
> 
> We recently changed the command to `/start` @byteballet

i forgot...

byteballet:
I think there is a mistake in default weight of time-labels.

  "time-labels": [
    {
      "name": "Time: <1 Hour",
      "weight": 0.125,
      "value": 3600
    },
    {
      "name": "Time: <1 Day",
      "weight": 1,
      "value": 86400
    },
    {
      "name": "Time: <1 Week",
      "weight": 2,
      "value": 604800
    },
    {
      "name": "Time: <2 Weeks",
      "weight": 3,
      "value": 1209600
    },
    {
      "name": "Time: <1 Month",
      "weight": 4,
      "value": 2592000
    }
  ]
I believe weight of "Time: <1 Month" must be 5 as "Time: <3 weeks" should be 4.
@pavlovcik rfc

pavlovcik:
I think I updated the time labels to lower values recently. Check out https://GitHub.com/ubiquity/.GitHub/ubiquibot-config.json for the new values. 
I think don't worry too much about those values for now.

pavlovcik:
@wannacfuture any idea why this isnt generating a payment permit?

wannacfuture:
no, it won't identify as meta issue. it checks checkbox

pavlovcik:
lol the bot is totally bricked on this repo apparently.

byteballet:
> 

It seems it's an issue related to database. https://github.com/ubiquity/ubiquibot/issues/574

EtherealGlow:
how 💀

byteballet:
> how 💀

My GH user name was ByteBallet and the wallet is assigned to ByteBallet on db. I changed my name to byteballet and it seems no wallet address was registered to byteballet before https://github.com/ubiquity/ubiquibot/issues/495#issuecomment-1702760081
[...]

Issue #431, opened 2023-06-24
`value` should be dynamically parsed from the time value using the `ms` library


pavlovcik:
- Parse `value` directly from the time label. 
  - Include the number, and include the unit (e.g. Hours, Days.)
  - Exclude the `<` symbol.
- Remove the `value` associated in the config. 
- Use `ms` library to derive the value at runtime from the time label based on the number and unit. 

> Value should be dynamically parsed from the time value using the `ms` library as we already discussed in the past @0xcodercrane

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1240675183_

b4s36t4:
@pavlovcik is this up for grab?!

b4s36t4:
@pavlovcik `ms` package doesn't have the capability to convert `1 Month` to millseconds? What should I do?!

pavlovcik:
I have a hard time believing that but I can't make the time to verify that other than asking chatgpt

b4s36t4:
<img width="620" alt="Screenshot 2023-07-07 at 2 53 30 AM" src="https://github.com/ubiquity/ubiquibot/assets/59088937/ea16407f-f993-46d3-a7e1-c7720a6d8d7d">

b4s36t4:
Shall I go on and convert that to days and calculate it?

b4s36t4:
@pavlovcik here are the test cases from vercel. even here no mention of `month` parsing. https://github.com/vercel/ms/blob/master/src/index.test.ts

pavlovcik:
@wannacfuture you wanna pick this one up? You can use the linked pull request as reference.

wannacfuture:
> @wannacfuture - Releasing the bounty back to dev pool because the allocated duration already ended! Last activity time: Sat Jun 24 2023 10:04:05 GMT+0000 (Coordinated Universal Time)


Currently the bot checks if there is hunter's comment or pull request(including reviews and comments on it) for last activity time check for an issue.
And when we assign bounty hunter manually there would be no comments or Prs of the hunter so the bot will identify as the allocated duration ended.


Well actually when we assigning bounty hunter manually the bot should add comment like : "@wannacfuture the time range for this issue.... ". And there will be no problem.
It used to do so, but I'm not sure why the bot hasn't added it.
Rfc, @pavlovcik , @rndquu

wannacfuture:
[...]

PR #594, opened 2023-08-09
Update ubiquibot-config-default.json

pavlovcik:
- Reference: https://github.com/ubiquity/ubiquibot/pull/558#discussion_r1286956838


pavlovcik:
I just realized it's a JSON object. Perhaps we can use `.jsonc`?

pavlovcik:
Mixed feelings on adding new dependencies just to add a comment, but I really would prefer to increase code maintainability. I want new developers to understand why we are adding specific integers in as defaults.

ubiquibot-config-default.jsonc
```diff
@@ -1,7 +1,15 @@
...
+  "price-multiplier": 1,
+  "issue-creator-multiplier": 1000,
+  "payment-permit-max-price": 9007199254740991 /* Number.MAX_SAFE_INTEGER */,
+  "max-concurrent-assigns": 9007199254740991 /* Number.MAX_SAFE_INTEGER */,
+  "assistive-pricing": false,
+  "disable-analytics": false,
+  "comment-incentives": false,
+  "register-wallet-with-verification": false,
+  "promotion-comment": "\n<h6>If you enjoy the DevPool experience, please follow <a href='https://github.com/ubiquity'>Ubiquity on GitHub</a> and star <a href='https://github.com/ubiquity/devpool-directory'>this repo</a> to show your support. It helps a lot!</h6>",
+  "default-labels": [],
```
pavlovcik, line 12: It could be nice to have everything labels related to be nested. For example:

```json
{
  "labels": {
    "default": [],
    "time": [
      {
        "name": "Time: <1 Hour",
        "weight": 0.125,
        "value": 3600
      },
      {
        "name": "Time: <1 Day",
        "weight": 1,
        "value": 86400
      },
      {
        "name": "Time: <1 Week",
        "weight": 2,
        "value": 604800
      },
      {
        "name": "Time: <2 Weeks",
        "weight": 3,
        "value": 1209600
      },
      {
        "name": "Time: <1 Month",
        "weight": 4,
        "value": 2592000
      }
    ],
    "priority": [
      {
        "name": "Priority: 0 (Normal)",
        "weight": 1
      },
      {
        "name": "Priority: 1 (Medium)",
        "weight": 2
      },
      {
        "name": "Priority: 2 (High)",
        "weight": 3
      },
      {
        "name": "Priority: 3 (Urgent)",
        "weight": 4
      },
      {
        "name": "Priority: 4 (Emergency)",
        "weight": 5
      }
    ]
  }
}
```

pavlovcik: Is `labels.time.weight` necessary if we already have `labels.time.value`?

pavlovcik: Perhaps we should remove `weight` for both `time` and `priority`, and then only use `value`.

```json
{
  "labels": {
    "default": [],
    "time": [
      {
        "name": "Time: <1 Hour",
        "value": 3600
      },
      {
        "name": "Time: <1 Day",
        "value": 86400
      },
      {
        "name": "Time: <1 Week",
        "value": 604800
      },
      {
        "name": "Time: <2 Weeks",
        "value": 1209600
      },
      {
        "name": "Time: <1 Month",
        "value": 2592000
      }
    ],
    "priority": [
      {
        "name": "Priority: 0 (Normal)",
        "value": 1
      },
      {
        "name": "Priority: 1 (Medium)",
        "value": 2
      },
      {
        "name": "Priority: 2 (High)",
        "value": 3
      },
      {
        "name": "Priority: 3 (Urgent)",
        "value": 4
      },
      {
        "name": "Priority: 4 (Emergency)",
        "value": 5
      }
    ]
  }
}
```

src/utils/private.ts
```diff
@@ -20,7 +20,13 @@ import {
   getRegisterWalletWithVerification,
 } from "./helpers";
 
-import DEFAULT_CONFIG_JSON from "../../ubiquibot-config-default.json";
+import fs from "fs";
+import stripJsonComments from "strip-json-comments";
+
+const configPath = "../../ubiquibot-config-default.jsonc";
+const configFile = fs.readFileSync(configPath, "utf8");
```
rndquu, line 27: As far as I remember if `fs` is used then the config file is not included in the netlify build, @0xcodercrane pls confirm

ubiquibot-config-default.jsonc
rndquu: > Is `labels.time.weight` necessary if we already have `labels.time.value`?

As far as I understand both are necessary. `labels.time.weight` is used for pricing while `labels.time.value` is used for issues' time boundaries

pavlovcik: What are "time boundaries"?

We only had ranges for prices in the past but those have since been deprecated.

src/utils/private.ts
0xcodercrane: We should avoid using `fs.readFileSync`.

`readFileSync` reads the config file in the execution step (means the config file is included when running) but to work as a serverless function, we should have all the necessary files included in the build step (exclude ubiquibot-config.yml -- it reads from the installed repo) because the ubiquibot GitHub app installed on the repo, not the source code being installed.

0xcodercrane (changes requested):
https://github.com/ubiquity/ubiquibot/pull/594/files#r1291011934

ubiquibot-config-default.jsonc
rndquu: > What are "time boundaries"?

"Time boundary" is the amount of seconds for solving an issue. 

Check this config:
```
{
        "name": "Time: <1 Hour",
        "weight": 0.125,
        "value": 3600
}
```

Issue's price label is calculated this way: `Pricing = timeLabel.weight * 100 * priceMultiplier`
When a bounty hunter self assigns an issue then he has `timeLabel.value` seconds to solve an issue

So both `timeLabel.weight` and `timeLabel.value` are necessary

pavlovcik: There's a task open for deriving the amount of time using the Vercel `ms` library.

rndquu: Then we should wait for https://github.com/ubiquity/ubiquibot/issues/431 and https://github.com/ubiquity/ubiquibot/issues/495

Then we could indeed remove `timeLabel.weight` and `timeLabel.value`

pavlovcik:
I just moved the notes to the README.md

rndquu (approved):

Issue #595, opened 2023-08-09
New Default `Time` Labels


pavlovcik:
Longer than 1 week bounties means that more thought needs to be put into it to break it down. So I think they should be removed from the default list to not encourage their use. Also there were some situations where we needed higher intra-day resolution.

Please update the labels in this repository, and the org-wide defaults. 

- `Time: <1 Hour`
- `Time: <2 Hours`
- `Time: <4 Hours`
- `Time: <1 Day`
- `Time: <1 Week`

---

>> @pavlovcik Auto-generation of Time labels will be disabled right? Since we no longer have strictly default time labels
>
> That makes sense. 
>
> To clarify, you mean that when the bot is added to your repository, the time labels are not automatically generated and added to the repository?
>
> However, I think that if they enable `assistive-pricing` it is A LOT easier for them to onboard if we give them some basic time labels to start with. At least `Time: <1 Hour`; `Time: <1 Day`; `Time: <1 Week` labels.
>
> (I kind of want to add `<4 Hours` because in the last week it was needed twice!)
>
_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/584#issuecomment-1671990988_

pavlovcik:
> Skipping `/multiplier` because it is disabled on this repo

I wonder if it makes sense to enable all of the slash commands by default. This is because partners must go out of their way to use those commands, whereas the other capabilities of the bot are more "automatic"

PR #596, opened 2023-08-09
Update ubiquibot-config.yml

pavlovcik:
Resolves https://github.com/ubiquity/ubiquibot/issues/595

### DIFF
--- a/.github/ubiquibot-config.yml
+++ b/.github/ubiquibot-config.yml
@@ -1,22 +1,21 @@
----
 evm-network-id: 100
 price-multiplier: 1.5
 time-labels:
   - name: "Time: <1 Hour"
     weight: 0.125
     value: 3600
+  - name: "Time: <2 Hours"
+    weight: 0.25
+    value: 7200
+  - name: "Time: <4 Hours"
+    weight: 0.5
+    value: 14400
   - name: "Time: <1 Day"
     weight: 1
     value: 86400
   - name: "Time: <1 Week"
     weight: 2
     value: 604800
-  - name: "Time: <2 Weeks"
-    weight: 3
-    value: 1209600
-  - name: "Time: <1 Month"
-    weight: 4
-    value: 2592000
 priority-labels:
   - name: "Priority: 0 (Normal)"
     weight: 1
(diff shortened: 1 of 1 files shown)


.github/ubiquibot-config.yml
```diff
@@ -31,9 +30,9 @@ priority-labels:
 default-labels:
   - "Time: <1 Hour"
   - "Priority: 0 (Normal)"
-  - "Test"
 payment-permit-max-price: 1000
 comment-incentives: true
 max-concurrent-bounties: 2
 promotion-comment: "\n<h6>If you enjoy the DevPool experience, please follow <a href='https://github.com/ubiquity'>Ubiquity on GitHub</a> and star <a href='https://github.com/ubiquity/devpool-directory'>this repo</a> to show your support. It helps a lot!</h6>"
 register-wallet-with-verification: false
+assistive-pricing: true
```
pavlovcik, line 38: It is unclear to me if it inherits from https://github.com/ubiquity/ubiquibot-config/blob/c6b3938f62b9bcf2701e68fc29b9015f7d56da7f/.github/ubiquibot-config.yml 

Or if we must redundantly specify config properties per repostiory?

whilefoo: Repo config properties override org config properties but if a property in repo config is not defined, it will use org config.

rndquu: Yes, as whilefoo stated, if the property is missing in a repo wide config then an org wide property will be used. So there is no need to redundantly specify config properties per repository.

rndquu (approved):

.github/ubiquibot-config.yml
pavlovcik: We should enable all features on our organization wide config then! I assume this is why `multiplier` isnt working?
- https://github.com/ubiquity/ubiquibot/issues/595#issuecomment-1673185590
- https://github.com/ubiquity/ubiquibot-config/blob/development/.github/ubiquibot-config.yml
- https://github.com/ubiquity/ubiquibot/blob/development/.github/ubiquibot-config.yml

Issue #440, opened 2023-06-28
Prevent Double Commenting


pavlovcik:
In the `addComment` function (or whatever the name of the function is that we use everywhere in the code to add a comment) we should check the last comment in the conversation. If the last comment is from the bot, then the bot should edit that comment instead of posting another one. 

> Updated the wallet address for @pavlovcik successfully!	 Your new address: `0x4007CE2083c7F3E18097aeB3A39bb8eC149a341d`

_Originally posted by @ubiquibot[bot] in https://github.com/ubiquity/ubiquibot/issues/410#issuecomment-1610792247_

web4er:
I see double comment by the bot is not happening always. It happens when you edit your comment. 

Is removing the double comment in this case a good idea? Here is my analysis from my limited knowledge of the bot. 

A bot comment is usually just not a comment. It has side effects, e.g. saving wallet to DB. If you remove the double comment, you need to undo its side effects. 

If you change wallet, and then edit the comment to assign, both have side affects. I think double comment in this case should stay.  Here is an example https://github.com/web4er/ubi-iss-rep/issues/4#issuecomment-1612722021

Github already doesn't trigger the change if the comment was editing to same old comment. However, if you add space or something invisible to old comment, a double comment by the bot happens. In this case the double comment should be prevented. But instead of removing the double comment, a better approach is to improve the comment parser.  I see two approaches to improve the parser. 
1. Make it very strict e.g. if there are extra spaces, just ignore it.
2. Use regex and allow multiple white characters

pavlovcik:
We can see the entire edit history and cross reference timestamps if necessary to track things down. Ultimately, when reviewing a conversation I think that the only information thats relevant is the final state of e.g. the registered wallet. 

Furthermore this proposal also solves for when a new contributor comments on a repository and the bot will reply with a help menu. If that new user edits their first comment, the bot will continue to post those tutorial comments again and again.

pavlovcik:
Intuitively, the bot should adapt its response based on what it was responding to. 

In practice, this means that we should detect which comment was edited, and then edit the response of the bot in the immediately following comment.

web4er:
I have started work on this. Based on what is available through GitHub API, this is my approach. 

Let's assume what the user is commenting are commands
- We already have information in the bot if it is `edited` comment or newly `created` comment by the user. 
- If it is an edited comment, we retrieve 2 comments since the `created_at` time of this edited comment.
[...]

PR #597, opened 2023-08-10
feat: fixed minor bugs for parent issue fix

### LINKED ISSUE #268: Disable Bounties for "Parent Issues"
### Context

It's fast and easy to write todo lists in the original comment of a GitHub issue, and then immediately convert them to "child issues" directly from the GitHub user interface.

The problem I am seeing is that the bounty system breaks down where contributors can "double dip" on rewards easily by claiming bounties for the actual bounties (child issues) and then the "meta" or "parent" issue.

The argument possibly could be made for project management. However in the future all of the **issue creators**, and **commenters** (both are how the project manager would "manage") are to receive compensation for their contributions upon the completion of the bounty.

### Proposal

- When changing a label, the bot should check if there are linked issues in the original comment and disable setting a price (and most importantly, as a consequence, payment permit generation.)

- The bot should comment explaining why it refuses to set a price.

---

              Parent issues should not have bounties. Instead, the child issues should have bounties.

_Originally posted by @pavlovcik in https://github.com/ubiquity/grants/issues/5#issuecomment-1519067539_

wannacfuture:
closes #268 
fixed some concerns
https://github.com/wannacfuture/Battleship/issues/77


wannacfuture:
hey, @web4er 
Just need QA be done for this PR.

src/configs/strings.ts
```diff
@@ -2,4 +2,6 @@ export const GLOBAL_STRINGS = {
   unassignComment: "Releasing the bounty back to dev pool because the allocated duration already ended!",
   askUpdate: "Do you have any updates",
   assignCommandDisabledComment: "The `/assign` command is disabled for this repository.",
+  skipPriceLabelGenerationComment: "Price label generation skipped since it is identified as parent issue.",
```
pavlovcik, line 5:
```suggestion
  skipPriceLabelGenerationComment: "Pricing is disabled on parent issues.",
```

src/configs/strings.ts
```diff
@@ -2,4 +2,6 @@ export const GLOBAL_STRINGS = {
...
+  ignorStartCommandForParentIssuesComment: "The `/start` command is disabled for parent issues.",
```
pavlovcik, line 6:
```suggestion
  ignoreStartCommandForParentIssueComment: "Please select a child issue from the specification checklist to work on. The `/start` command is disabled on parent issues.",
```

pavlovcik (changes requested):
Minor adjustments needed.

wannacfuture:
REady for QA

pavlovcik (approved):
Code looks good.

pavlovcik:
> Ready for QA

I accidentally merged it but lets just handle QA retroactively then (you can do a hot fix if QA finds an issue)

web4er:
> hey, @web4er Just need QA be done for this PR.

QA: https://github.com/web4erOrg/ubiquibot/issues/20

Failures:
- if user opens a PR, and PR is merged, permnit is being generated. Permit should not be generated. https://github.com/web4erOrg/ubiquibot/issues/20#issuecomment-1672879797
- bot is adding `pricing is disabled...` message multiple times at start. Can be seen both in https://github.com/web4erOrg/ubiquibot/issues/20 and also was reproduced in https://github.com/web4erOrg/ubiquibot/issues/24
- some price labels are not being removed by the bot https://github.com/web4erOrg/ubiquibot/issues/20#issuecomment-1672897384

wannacfuture:
I fixed the other things. But multiple message issue:

You added both Time label and priority label at once so the pricingLabelLogic is called twice.
that's the reason why the warning message shows twice.

Rfc, @pavlovcik , @web4er

wannacfuture:
> I thought we programmed the comment function to edit the last reply to prevent double posts anyways?

Not sure about this. Could you provide the link or something?

pavlovcik:
- https://github.com/ubiquity/ubiquibot/issues/440

web4er:
> > I thought we programmed the comment function to edit the last reply to prevent double posts anyways?
> 
> Not sure about this. Could you provide the link or something?

https://github.com/ubiquity/ubiquibot/blob/2abac46f6ecba351677f528f2fa3b29e200029b3/src/helpers/issue.ts#L149
There is `upsertComment` function instead of `addComment` function. 

But it has a condition
https://github.com/ubiquity/ubiquibot/blob/2abac46f6ecba351677f528f2fa3b29e200029b3/src/helpers/issue.ts#L150

pavlovcik:
> https://github.com/ubiquity/ubiquibot/blob/2abac46f6ecba351677f528f2fa3b29e200029b3/src/helpers/issue.ts#L150


Okay so if I am understanding this correctly, this means that if the human edits their comment, that the bot will edit its comment. 

How about a hot fix so that the bot will never double post? Does that make sense? It will band-aid a lot of the small edge cases like when it aggressively follows up a ton of times, or what we were discussing earlier in this conversation.

web4er:
> Okay so if I am understanding this correctly, this means that if the human edits their comment, that the bot will edit its comment.

correct

For this PR, In a new hotfix https://github.com/ubiquity/ubiquibot/pull/598 the initial double comment about pricing has been changed into no comment at all. And I have already approved it. I think in most cases user already knows that pricing can't be set in parent issue, and it is like noise even as a single comment. In case user doesn't know, the `Pricing is disabled on parent issues.` will show up if user tries to set price label. 

I think those follow-ups have different causes and the root cause should be fixed. Editing the bot comment will be like sweeping them under the rug.

PR #602, opened 2023-08-10
Resolves #488 Ai_duplicate_check

### LINKED ISSUE #488: AI: Check for Duplicate Specs
# Simplified Specification

Derived from the conversation comment [here](https://github.com/ubiquity/ubiquibot/issues/488#issuecomment-1668927166). 

We could do things without a vector database in order to simplify things, here's how: 

1. A new issue is posted
2. We ask ChatGPT to extract a word list of the most "important" (i.e. unique adjectives?) words (for example, when I wanted to find this issue, I searched for the term "duplicate")
3. We search the repository for all issues with the important words
4. We go from highest issue number (most recent) and read the specification. 
5. If >80% confidence, stop the search and link back to it with a warning saying that it's likely to be a duplicate. 

This approach might be a little more brittle (rate limits) but we won't have to worry about the database which should make implementation and maintenance much easier.

# Original Specification (for reference only, do not use)

### Overview

As part of our AI roadmap, we want to [prevent duplicate issues](https://github.com/ubiquity/ubiquibot/discussions/107#discussioncomment-4945998).

The idea is to have:
1. an event handler for creating a new issue 
2. that passes the issue specification to ChatGPT (asynchronously/slowly is fine)
3. have a cache "vector database" of issue similarity within the same repository only
4. have a `ubiquibot-config` property with a `issue-similarity-confidence-threshold: float`

Then with an issue similarity confidence threshold of e.g. `0.8` that means that if the bot is 80% confident it's a redundant issue. It will then post a comment that will explain it is X% confident that this is a redundant issue and backlink the redundant issue in question.

### Remarks

- This should be broken down further but will put a broad `Time: <1 Week` for anybody particularly motivated to get started with it. 
- I'm not sure about the relationship with ChatGPT and the vector database, but from what I understand, vector database is good for evaluating similarity between things and ChatGPT is a good interpretation engine. 

### Context

              Duplicate 

- https://github.com/ubiquity/ubiquibot/issues/477

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/406#issuecomment-1630097779_

ByteBallet:
Resolves #488
QA for byteballet/santa-bringyouwishes#39


pavlovcik:
What happened here

ByteBallet:
> What happened here

Oh, I had some issues on my commit message. I will update it and will send pr again.

PR #604, opened 2023-08-10
Logs to supabase

### LINKED ISSUE #592: We should save all of our logs to Supabase.
- Every log should associate a comment (or event handler) ID. 
- Using the IDs we should aim to be able to backtrace the exact conversation/context from the GitHub conversation.
- Clean up all the old code related to mezmo as that will be deprecated after this implementation. 
- If there is no GitHub mutation associated with the logging activity (a comment on GitHub) then there is no way to link back a "node id" in which case, the "node id" can be null.

devpanther:
Resolves #592 

View of `log_entries` on my DB
Screenshot 2023-08-11 at 12 42 44 PM

- It was made to work like the previous memzo logger


pavlovcik:
Could you add us to your supabase as a contributor so we can try manipulating your bot and seeing the logs? 

Your screenshot looks promising!

devpanther:
> Could you add us to your supabase as a contributor so we can try manipulating your bot and seeing the logs?
> 
> Your screenshot looks promising!

Which email should I use to invite you?, i'd have leave my bot running locally

pavlovcik:
bounties@ubq.fi

devpanther:
> [bounties@ubq.fi](mailto:bounties@ubq.fi)

Sent a request, you should let me know when you'd be available to test this

whilefoo (changes requested):
Actually I'd prefer if we use an existing logging library instead of reinventing the wheel. For example [winston](https://github.com/winstonjs/winston) supports custom [Transports](https://github.com/winstonjs/winston#adding-custom-transports), which we can use to write logs to database. We can still have a simple wrapper for adding custom fields from bot's context but we don't need to implement concurrency.
Concurrency is hard to reason about and can introduce race conditions and bugs so by using a battle tested library we can avoid that.

@pavlovcik thoughts?

pavlovcik:
I'm not concerned about race conditions because logs print in series. 

Not reinventing libraries is only valid before the reinvention of libraries. It looks like this pull request is already finished?

supabase/migrations/20230814093932_log_entries_rls.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE log_entries ENABLE ROW LEVEL SECURITY
```
rndquu, line 1: @devpanther Did you test how the bot works with this "restrict all" policy? As far as I understand this policy does not allow writing to a DB to any of the users.

devpanther: It allows writing if you're using a service role api key which the bot is using..

src/adapters/supabase/helpers/log.ts
```diff
@@ -0,0 +1,196 @@
...
+      const log = this.logQueue.shift();
+      if (!log) {
+        continue;
+      }
+      await this.processLogs(log);
+    }
+  }
+
+  async throttle(fn: () => void) {
+    if (this.throttleCount >= this.maxConcurrency) {
```
rndquu, line 102: @devpanther So if there are 6 concurrent "save log" requests while `this. maxConcurrency === 5` then the 6th "save log" request is no saved to a DB, right?

devpanther: > @devpanther So if there are 6 concurrent "save log" requests while `this. maxConcurrency === 5` then the 6th "save log" request is not saved to a DB, right?

It remains in the queue and will be sent once the rest are resolved... so it does not overwhelm the supabase database

src/adapters/supabase/helpers/log.ts
```diff
@@ -0,0 +1,196 @@
...
+    try {
+      await this.sendLogsToSupabase(log);
+    } catch (error) {
+      console.error("Error sending log (after retry):", error);
+      await this.retryLog(log, retryCount + 1);
+    }
+  }
+
+  async processLogQueue() {
+    while (this.logQueue.length > 0) {
```
rndquu, line 92: @devpanther Why do we need this loq queue? Why can't we simply send logs to supabase every time there is a new log and let the node's event loop handle everything?

devpanther: > @devpanther Why do we need this loq queue? Why can't we simply send logs to supabase every time there is a new log and let the node's event loop handle everything?

I just don't want to overwhelm the db at any point.. right now it logs go in one at a time, but its to handle cases where they come in bulk.

I've seen supabase db lose connection when too many requests is sent, so I am handling it in queue and retrying on error to make sure every single log get in and in their right order

supabase/migrations/20230814093932_log_entries_rls.sql
devpanther: > @devpanther Did you test how the bot works with this "restrict all" policy? As far as I understand this policy does not allow writing to a DB to any of the users.

It allows writing if you're using a service role api key which the bot is using..

rndquu:
Although in most cases it is better to use a "battle tested" logging library (like [winston](https://www.npmjs.com/package/winston)) in this particular case (when we simply need to store a row in a DB) we don't need anything fancy so keeping it simple is also the solution.

Regarding the [queue feature](https://github.com/ubiquity/ubiquibot/blob/a5664d32d8ca7af10c03b775d5b10039d302a540/src/adapters/supabase/helpers/log.ts#L145), I anticipate hard to catch bugs in the future so I would remove the queue logic from [log.ts](https://github.com/ubiquity/ubiquibot/blob/a5664d32d8ca7af10c03b775d5b10039d302a540/src/adapters/supabase/helpers/log.ts) and apply it (or check for other solution) only when we spot issues with supabase API.

@whilefoo @0xcodercrane help

supabase/migrations/20230814093932_log_entries_rls.sql
0xcodercrane: Yeah, @devpanther 's right here. we already switched using service role key in backend.

supabase/migrations/20230811105133_log_entries.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE log_entries (
```
0xcodercrane, line 1:
```suggestion
CREATE TABLE logs (
```

devpanther:
> only when we spot issues with supabase API.

I've tried something like this on a production environment, where we didn't need something fancy but the queue was a fix to the supabase api not responding sometimes.. 

That's why I anticipated it here, we won't know when logs go missing without a proper look.. That's why there's a queue and retry

supabase/migrations/20230811105133_log_entries.sql
```diff
@@ -0,0 +1,12 @@
...
+  id SERIAL PRIMARY KEY,
+  repo_name TEXT,
+  org_name TEXT,
+  comment_id INT,
+  issue_number INT,
+  log_message TEXT,
+  level INT,
```
0xcodercrane, line 8: Why is the data type of `level` INT? Log level should be "debug", "info", "warn", "error"... so I prefer making it string.

whilefoo:
> Regarding the [queue feature](https://github.com/ubiquity/ubiquibot/blob/a5664d32d8ca7af10c03b775d5b10039d302a540/src/adapters/supabase/helpers/log.ts#L145), I anticipate hard to catch bugs in the future

It is because of the queue feature I recommended a battle tested library because I'm worried there may be race conditions and bugs that are hard to spot on PR review and can later cause issues that we can't pinpoint.

src/bindings/config.ts
```diff
@@ -32,8 +32,8 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
 
   const botConfig: BotConfig = {
     log: {
+      logEnvironment: process.env.LOG_ENVIRONMENT || "production",
```
0xcodercrane, line 35: What does `logEnvironment` do?

src/bindings/event.ts
```diff
@@ -17,14 +17,8 @@ export const getBotConfig = () => botConfig;
 let adapters: Adapters = {} as Adapters;
 export const getAdapters = () => adapters;
 
-export type Logger = {
-  info: (msg: string | object, options?: JSON) => void;
-  debug: (msg: string | object, options?: JSON) => void;
-  warn: (msg: string | object, options?: JSON) => void;
-  error: (msg: string | object, options?: JSON) => void;
-};
```
0xcodercrane, line 25: This is an interface that we can integrate any kind of logging libraries. 
Even if you create a GithubLogger, it should be inherited from `Logger`.

src/bindings/event.ts
```diff
@@ -17,14 +17,8 @@ export const getBotConfig = () => botConfig;
...
-let logger: Logger;
-export const getLogger = (): Logger => logger;
+let logger: GitHubLogger;
+export const getLogger = (): GitHubLogger => logger;
```
0xcodercrane, line 21: Revert this and inherit GithubLogger from Logger.

0xcodercrane (changes requested):
1/ Please resolve conversations first.
2/ Create 1 migration file. I don't see any reason we have to have 2 migrations for simple queries.
3/ Lets resolve the lint stuff

supabase/migrations/20230811105133_log_entries.sql
whilefoo: I recommended that because it's better for database index. We can convert to string in the UI

devpanther:
> 1/ Please resolve conversations first.
> 2/ Create 1 migration file. I don't see any reason we have to have 2 migrations for simple queries.
> 3/ Lets resolve the lint stuff 

The two queries didn't work together, so I had to separate it.. So you create the table first and enable RLS

src/bindings/config.ts
devpanther: If its development.. It logs on the console for easy log tracking.. Would come in handy, I've always wanted to see logs without going to memzo dashboard or using console.log and the forgetting to remove it

src/bindings/event.ts
devpanther: Some libraries like `wiston` does not do the log.info  or debug thing though

But it'd bring that back and inherit from it

0xcodercrane:
In terms of the queue feature, I like this idea but if supabase API works properly, we don't need to add them to the queue right?

In terms of the processing performance (ubiquibot is a lambda function that is executed from the github events), we should consider the function timeout (< 10s). In case we have tons of retries, it will definitely affect the performance.

so I think you can add a flag to enable/disable or make `retryLimit` configurable. Totally we should get the control of the factors that affect the performance.

0xcodercrane:
> The two queries didn't work together, so I had to separate it.. So you create the table first and enable RLS

Interesting, but I still think its possible. would you post the error message here so I can help with that?

devpanther:
> > The two queries didn't work together, so I had to separate it.. So you create the table first and enable RLS
> 
> Interesting, but I still think its possible. would you post the error message here so I can help with that? 

Its just an invalid sql query, it's expecting the table to be created already

rndquu:
> In terms of the queue feature, I like this idea but if supabase API works properly, we don't need to add them to the queue right?
> 
> In terms of the processing performance (ubiquibot is a lambda function that is executed from the github events), we should consider the function timeout (< 10s). In case we have tons of retries, it will definitely affect the performance.
> 
> so I think you can add a flag to enable/disable or make `retryLimit` configurable. Totally we should get the control of the factors that affect the performance.
> 
> @devpanther @rndquu

So you're talking about:
1. Add a flag to enable/disable the logs queue
2. Disable the queue by default

Right?

src/bindings/event.ts
devpanther: > Revert this and inherit GithubLogger from Logger.

I'd have to make `Logger` an interface so I can use `implements`.. you cannot extend types

devpanther:
> Right?

I think he means a flag to disable/enable `retryLimit` or control the number of times it can retry

0xcodercrane:
> So you're talking about:
> 
> 1. Add a flag to enable/disable the logs queue
> 2. Disable the queue by default
> 
> Right?

Right

src/adapters/supabase/helpers/log.ts
pavlovcik: http/s max concurrency by default in Chrome is 6. Perhaps that is the optimal amount. Why did you choose 5?

pavlovcik:
Looks like pretty solid progress so far.

devpanther:
@0xcodercrane should I add the retry Limit configuration as an env?

0xcodercrane:
> @0xcodercrane should I add the retry Limit configuration as an env?

Please configure it in ubiquibot-config.yml/ubiquibot-config-default.json/env.

pavlovcik:
> > @0xcodercrane should I add the retry Limit configuration as an env?
> 
> Please configure it in ubiquibot-config.yml/ubiquibot-config-default.json/env.

Shouldn't this be some type of environment variable inside of Netlify that we control internally? I don't think we should let our partners configure settings related to logging! Other than the secret option of fully disabling logging.

devpanther:
> > > @0xcodercrane should I add the retry Limit configuration as an env?
> > 
> > Please configure it in ubiquibot-config.yml/ubiquibot-config-default.json/env.
> 
> Shouldn't this be some type of environment variable inside of Netlify that we control internally? I don't think we should let our partners configure settings related to logging! Other than the secret option of fully disabling logging. 

That's why I asked, it's supposed to be an internal variable..

I'd make it an env and push now

devpanther:
last commit added `retryLimit` configurable using env.. so it can be controlled internally

`0` for retry disabled
`> 0` for number of retries

@pavlovcik @0xcodercrane

src/adapters/supabase/helpers/log.ts
devpanther: > http/s max concurrency by default in Chrome is 6. Perhaps that is the optimal amount. Why did you choose 5?

Just a choice, but I made it 6 now

src/adapters/supabase/helpers/log.ts
```diff
@@ -27,11 +27,12 @@ export class GitHubLogger implements Logger {
   private app;
   private logEnvironment;
   private logQueue: Log[] = []; // Your log queue
-  private maxConcurrency = 5; // Maximum concurrent requests
-  private retryLimit = 3; // Maximum number of retries
+  private maxConcurrency = 6; // Maximum concurrent requests
   private retryDelay = 1000; // Delay between retries in milliseconds
   private throttleCount = 0;
 
+  private retryLimit = process.env.LOG_RETRY || 0; // Retries disabled by default
```
0xcodercrane, line 34: We shouldn't read .env variable here. It should be read [here](https://github.com/devpanther/bounty-bot/blob/logs-to-supabase/src/bindings/config.ts).

You should also set the value to retryLimit in the constructor function.

0xcodercrane (changes requested):

0xcodercrane:
https://github.com/ubiquity/ubiquibot/commit/f6e99dfb406559a044969bb82aff478daae172cd

Create a migration for you @devpanther. please use this.

src/adapters/supabase/helpers/log.ts
devpanther: > We shouldn't read .env variable here. It should be read [here](https://github.com/devpanther/bounty-bot/blob/logs-to-supabase/src/bindings/config.ts).
> 
> You should also set the value to retryLimit in the constructor function.

Done

devpanther:
> [f6e99df](https://github.com/ubiquity/ubiquibot/commit/f6e99dfb406559a044969bb82aff478daae172cd)
> 
> Create a migration for you @devpanther. please use this.

Thank you, used that in the latest commit

pavlovcik:
@0xcodercrane @rndquu @whilefoo I guess this is ready to merge?

src/types/config.ts
```diff
@@ -61,8 +61,9 @@ export const AssignSchema = Type.Object({
 });
 
 export const LogConfigSchema = Type.Object({
+  logEnvironment: Type.String(),
   level: Type.String(),
```
whilefoo, line 65:
```suggestion
  maxLevel: Type.Enum(Level),
```

whilefoo (changes requested):
just some minor changes for naming and typing

src/adapters/supabase/helpers/log.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  async throttle(fn: () => void) {
+    if (this.throttleCount >= this.maxConcurrency) {
+      return;
+    }
+
+    this.throttleCount++;
+    try {
+      await fn();
```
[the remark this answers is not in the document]
devpanther: Both same thing, calling fn() was just shorter while preserving `this` at the same time

devpanther:
> just some minor changes for naming and typing

Should be good now

src/adapters/supabase/helpers/log.ts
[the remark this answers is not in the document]
whilefoo: If you call `processLogQueue` directly, it still preserves `this`, isn't it?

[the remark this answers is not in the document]
devpanther: Yes, but then the first fn call gives way for it to be called the next time with the preserved this..

If there was no `fn` param, it can no longer be passed in the `finally` block to preserve the this

So let's leave that the way it is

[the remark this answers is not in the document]
whilefoo: I just tried directly calling it and it works the same...I still don't understand why does this make any difference? 
`this` is just a reference to the current instance of the class and it will be available to the function either way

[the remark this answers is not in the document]
devpanther: > I just tried directly calling it and it works the same...I still don't understand why does this make any difference? `this` is just a reference to the current instance of the class and it will be available to the function either way

Alright, I've removed it

whilefoo:
@devpanther I just tested this locally and it appears some logs are missing.

stdout output:
```
INFO (server): Running Probot v12.3.1 (Node.js: v18.16.0)
INFO (server): Forwarding https://smee.io/dFBMktaX9Gr0bZd8 to http://localhost:3000/
INFO (server): Listening on http://localhost:3000
INFO (server): Connected
DEBUG (octokit): GitHub request: GET https://api.github.com/repos/whilefoo/ubiquibot/contents/.github%2Fubiquibot-config.yml - 200
UbiquiBot Config loaded! config: {"price":{"baseMultiplier":1.5,"issueCreatorMultiplier":1000,"timeLabels":[{"name":"Time: <1 Hour","weight":0.125,"value":3600},{"name":"Time: <2 Hours","weight":0.25,"value":7200},{"name":"Time: <4 Hours","weight":0.5,"value":14400},{"name":"Time: <1 Day","weight":1,"value":86400},{"name":"Time: <1 Week","weight":2,"value":604800}],"priorityLabels":[{"name":"Priority: 0 (Normal)","weight":1},{"name":"Priority: 1 (Medium)","weight":2},{"name":"Priority: 2 (High)","weight":3},{"name":"Priority: 3 (Urgent)","weight":4},{"name":"Priority: 4 (Emergency)","weight":5}],"commentElementPricing":{"blockquote":0,"break":0,"code":0,"definition":0,"emphasis":0,"heading":0,"html":0,"image":0,"imageReference":0,"inlineCode":0,"link":0,"linkReference":0,"list":0,"listItem":0,"paragraph":0,"root":0,"strong":0,"text":0,"thematicBreak":0,"delete":0,"footnoteDefinition":0,"footnoteReference":0,"table":0,"tableCell":0,"tableRow":0,"alignType":0,"flowContent":0,"listContent":0,"phrasingContent":0,"rowContent":0,"tableContent":0},"defaultLabels":["Time: <1 Hour","Priority: 0 (Normal)"]},"unassign":{"followUpTime":345600000,"disqualifyTime":604800000},"mode":{"paymentPermitMaxPrice":0,"disableAnalytics":false,"incentiveMode":true,"assistivePricing":true},"log":{"logEnvironment":"development","level":"debug","retryLimit":0},"wallet":{"registerWalletWithVerification":false}} undefined
UbiquiBot Started binding events... id: 5e9e73e0-3dbd-11ee-838e-48c9cf1f38c0, name: issue_comment.created, allowedEvents: issues.labeled,issues.unlabeled,issues.assigned,issues.unassigned,issues.closed,issues.opened,issues.reopened,issue_comment.created,issue_comment.edited,pull_request.opened,installation_repositories.added,push undefined
UbiquiBot Creating adapters for supabase, telegram, twitter, etc... undefined
UbiquiBot Running pre handlers: nullHandler, event: issue_comment.created undefined
UbiquiBot Running handler, name: nullHandler undefined
UbiquiBot Running main handlers: handleComment, event: issue_comment.created undefined
UbiquiBot Handling an issue comment on issue 21 undefined
UbiquiBot Running a comment handler: query undefined
UbiquiBot Skipping '/query' because it is disabled on this repo undefined
DEBUG (octokit): GitHub request: POST https://api.github.com/repos/whilefoo/ubiquibot/issues/21/comments - 201
    body: "Skipping `/query` because it is disabled on this repo"
UbiquiBot Running post handlers: nullHandler, event: issue_comment.created undefined
UbiquiBot Running handler, name: nullHandler undefined
UbiquiBot Running wildcard handlers: checkBountiesToUnassign,collectAnalytics,checkWeeklyUpdate undefined
UbiquiBot Getting all the issues... undefined
DEBUG (octokit): GitHub request: GET https://api.github.com/repos/whilefoo/ubiquibot/issues?state=open&per_page=100&page=1 - 200
UbiquiBot Checking expired bounties done! total: 0, unassigned: 0 undefined
UbiquiBot Collecting analytics information... undefined
DEBUG (octokit): GitHub request: GET https://api.github.com/repos/whilefoo/ubiquibot/contents/.github%2Fubiquibot-config.yml - 200
UbiquiBot Config loaded! config: {"price":{"baseMultiplier":1.5,"issueCreatorMultiplier":1000,"timeLabels":[{"name":"Time: <1 Hour","weight":0.125,"value":3600},{"name":"Time: <2 Hours","weight":0.25,"value":7200},{"name":"Time: <4 Hours","weight":0.5,"value":14400},{"name":"Time: <1 Day","weight":1,"value":86400},{"name":"Time: <1 Week","weight":2,"value":604800}],"priorityLabels":[{"name":"Priority: 0 (Normal)","weight":1},{"name":"Priority: 1 (Medium)","weight":2},{"name":"Priority: 2 (High)","weight":3},{"name":"Priority: 3 (Urgent)","weight":4},{"name":"Priority: 4 (Emergency)","weight":5}],"commentElementPricing":{"blockquote":0,"break":0,"code":0,"definition":0,"emphasis":0,"heading":0,"html":0,"image":0,"imageReference":0,"inlineCode":0,"link":0,"linkReference":0,"list":0,"listItem":0,"paragraph":0,"root":0,"strong":0,"text":0,"thematicBreak":0,"delete":0,"footnoteDefinition":0,"footnoteReference":0,"table":0,"tableCell":0,"tableRow":0,"alignType":0,"flowContent":0,"listContent":0,"phrasingContent":0,"rowContent":0,"tableContent":0},"defaultLabels":["Time: <1 Hour","Priority: 0 (Normal)"]},"unassign":{"followUpTime":345600000,"disqualifyTime":604800000},"mode":{"paymentPermitMaxPrice":0,"disableAnalytics":false,"incentiveMode":true,"assistivePricing":true},"log":{"logEnvironment":"development","level":"debug","retryLimit":0},"wallet":{"registerWalletWithVerification":false}} undefined
UbiquiBot Started binding events... id: 5fffffb0-3dbd-11ee-83f8-b19997698e79, name: issue_comment.created, allowedEvents: issues.labeled,issues.unlabeled,issues.assigned,issues.unassigned,issues.closed,issues.opened,issues.reopened,issue_comment.created,issue_comment.edited,pull_request.opened,installation_repositories.added,push undefined
UbiquiBot Creating adapters for supabase, telegram, twitter, etc... undefined
UbiquiBot Skipping the event. reason: sender is a bot undefined
INFO (http): POST / 200 - 370ms
INFO (server): POST http://localhost:3000/ - 200
DEBUG (octokit): GitHub request: GET https://api.github.com/repos/whilefoo/ubiquibot/issues?state=all&per_page=30&page=1 - 200
DEBUG (octokit): GitHub request: GET https://api.github.com/users/whilefoo - 200
DEBUG (octokit): GitHub request: GET https://api.github.com/users/whilefoo2 - 200
UbiquiBot Upserting users: whilefoo2,whilefoo undefined
UbiquiBot Creating a new user done", { data: null, error: [object Object] } undefined
UbiquiBot Creating a new user done", { data: null, error: [object Object] } undefined
UbiquiBot Upserting bounties: QA for Html comments,Bounty Proposal: testing default labels undefined
UbiquiBot Upserting an issue done, data: [object Object], error: null undefined
UbiquiBot Upserting an issue done, data: [object Object], error: null undefined
INFO (http): POST / 200 - 3535ms
INFO (server): POST http://localhost:3000/ - 200
```

Database:
image

src/adapters/supabase/helpers/log.ts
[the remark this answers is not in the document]
devpanther: > I just tried directly calling it and it works the same...I still don't understand why does this make any difference? `this` is just a reference to the current instance of the class and it will be available to the function either way

Show me how you called it directly here, the last commit i made broke the log queue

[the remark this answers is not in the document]
devpanther: > > I just tried directly calling it and it works the same...I still don't understand why does this make any difference? `this` is just a reference to the current instance of the class and it will be available to the function either way
> 
> Show me how you called it directly here, the last commit i made (calling it directly) broke the log queue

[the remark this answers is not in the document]
whilefoo:
```js
async throttle() {
    if (this.throttleCount >= this.maxConcurrency) {
      return;
    }

    this.throttleCount++;
    try {
      await this.processLogQueue();
    } finally {
      this.throttleCount--;
      if (this.logQueue.length > 0) {
        await this.throttle();
      }
    }
  }

  async addToQueue(log: Log) {
    this.logQueue.push(log);
    if (this.throttleCount < this.maxConcurrency) {
      await this.throttle();
    }
  }
```

devpanther:
> @devpanther I just tested this locally and it appears some logs are missing.

I think at that point the supabase adapter is not set, from the second event.. it sends all of the logs

src/adapters/supabase/helpers/log.ts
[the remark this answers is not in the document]
devpanther: Thanks, fixed it

devpanther:
> @devpanther I just tested this locally and it appears some logs are missing.

I've moved the creation of adapters before the log engine and it works now

src/adapters/supabase/helpers/log.ts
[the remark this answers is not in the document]
pavlovcik: > I just tried directly calling it and it works the same...I still don't understand why does this make any difference? `this` is just a reference to the current instance of the class and it will be available to the function either way

This is getting spicy

.env.example
```diff
@@ -10,6 +10,7 @@ SUPABASE_KEY=
 AUTO_PAY_MODE=
 ANALYTICS_MODE=
 
-# Use `trace` to get verbose logging or `info` to show less
+# Log environment
+LOG_ENVIRONMENT=production # development to see logs in console
 LOG_LEVEL=debug
```
pavlovcik, line 15: What are all the options perhaps you can include in a comment? I would imagine that the default level would not be `debug` (for production use)

pavlovcik:
```suggestion
LOG_LEVEL=debug # error, warn, info, http, verbose, debug, silly
```

What is silly? 👀

src/adapters/supabase/helpers/log.ts
pavlovcik: I think this is a more robust approach, good job!

.env.example
pavlovcik:
```suggestion
LOG_LEVEL=debug # 0: error 1: warn 2: info 3: http 4: verbose 5: debug 6: silly
```

pavlovcik (approved):

whilefoo (approved):

src/adapters/supabase/helpers/log.ts
0xcodercrane: @devpanther If you can link the fix commit/PR before you mark the conversation resolved so it would make reviewers easy to confirm.

0xcodercrane (approved):

PR #608, opened 2023-08-11
AI: Check for Duplicate Specs

### LINKED ISSUE #488: AI: Check for Duplicate Specs

byteballet:
Resolves #488
QA for byteballet/santa-bringyouwishes#39


src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (issue?.number) {
+          await upsertCommentToIssue(issue?.number, `Similar issue (${iss.title}) found at ${iss.html_url}.\nSimilarity is about ${probability}%`, "created");
+          return;
+        }
+      }
+    }
+  }
+};
+
+const countIncludedWords = (inputString: string, words: string[]): number => {
```
whilefoo, line 26: The input string is in Markdown syntax so you should preprocess the text by removing URLs, images, ... and put all the words in an array/set, so you can compare with important words.

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+};
+
+export const extractImportantWords = async (): Promise<string[]> => {
+  const { payload: _payload } = getBotContext();
+  const issue = (_payload as Payload).issue;
+  const { result: res } = await getAnswerFromChatGPT(`'Issue title: "${issue?.title}"\nIssue body: "${issue?.body}"'`);
+  return res.split("#");
+};
+
+export const getAnswerFromChatGPT = async (prompt: string): Promise<{ result: string }> => {
```
whilefoo, line 46: why return an object with string instead of just a string?

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+    data: body,
+  };
+  try {
+    const response = await axios(config);
+    const data: Choices = response.data;
+    const { choices: choice } = data;
+    const answer = choice[0].message.content;
+    return { result: answer };
+  } catch (error) {
+    return { result: "" };
```
whilefoo, line 84: log the error

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+      {
+        role: "user",
+        content:
+          (process.env.CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS ||
+            "I need your help to find duplicate issues on my GitHub repository. For context, the entire strategy is the following:\n\n1. A new issue is posted\n2. We ask you to extract a word list of the most \"important\" (i.e. unique adjectives?) words.\n3. We search the repository for all issues with the important words.\n4. We go from highest issue number (most recent) and read the issue description.\n5. If >80% confidence that it's a redundant issue, stop the search and link back to it with a warning saying that it's likely to be a duplicate.\nRight now, we are on step 2.\n") +
+          prompt,
+      },
+    ],
+    max_tokens: 1500,
+    temperature: 0,
```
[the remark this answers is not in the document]
ByteBallet: Yes, temperature is related to randomness of the answer. I set it to 0 because we don't need random answer...

src/helpers/similarity.ts
ByteBallet: Do you mean by console.log or current logger?

src/helpers/similarity.ts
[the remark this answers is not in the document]
ByteBallet: Would it be good to move temperature to botconfig?

src/helpers/similarity.ts
whilefoo: But this function only counts words, ChatGPT is not involved here

src/helpers/similarity.ts
whilefoo: current logger

src/helpers/similarity.ts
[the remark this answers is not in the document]
whilefoo: I don't think we need temperature (randomness) to be customisable. @pavlovcik rfc

src/helpers/similarity.ts
ByteBallet: Okay. Is it okay too use commentParser?

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+  const issue = (_payload as Payload).issue;
+  const issues = await listAllIssuesForRepo("all");
+  issues.shift();
+  const importantWords = await extractImportantWords();
+  const wordCount = importantWords.length;
+  for (const iss of issues) {
+    if (iss.body) {
+      const probability = (countIncludedWords(iss.body, importantWords) * 100.0) / wordCount;
+      if (probability > parseInt(process.env.SIMILARITY_THRESHOLD || "80")) {
+        if (issue?.number) {
```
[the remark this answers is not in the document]
whilefoo: I don't understand your answer. 
Let me rephrase, `if (issue?.number)` is not needed if you check if issue exists at the top of the function like this:
```js
if (!issue) {
  return;
}
```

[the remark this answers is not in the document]
ByteBallet: Yes, my bad.
Forgot to remove the comment.

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
+import { getBotContext } from "../bindings";
+import { Payload, Choices } from "../types";
+import { listAllIssuesForRepo, upsertCommentToIssue } from "../helpers/issue";
+import axios from "axios";
+
+export const findDuplicateOne = async () => {
+  const { payload: _payload } = getBotContext();
+  const issue = (_payload as Payload).issue;
+  const issues = await listAllIssuesForRepo("all");
+  issues.shift();
```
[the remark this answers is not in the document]
ByteBallet: It seemed `listAllIssuesForRepo` returned me all the issues including current one. So I needed `shift`. Isn't it?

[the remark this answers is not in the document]
whilefoo: I see, I think it would be better if we use `filter` function to explicitly remove the current issue.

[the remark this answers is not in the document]
ByteBallet: Sure, will do that.

src/helpers/similarity.ts
whilefoo: Yes, you can use [parseComments](https://github.com/ubiquity/ubiquibot/blob/dfb48b101d50c8134a51221384968f39fb7930b9/src/helpers/comment.ts#L22C23-L22C23)

src/helpers/similarity.ts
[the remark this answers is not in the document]
web4er: temperature and similarity threshold can be feature configs and not openai config. They can go in repo config. Eventually, I think openai here will be used for other purposes too. 

`duplicate-issue-ai-temperature`
`duplicate-issue-similarity-threshold`

What do you think @pavlovcik

web4er (changes requested):

src/helpers/similarity.ts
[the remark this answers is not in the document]
web4er: Or feature config stored as repo config
```
duplicate-issue-ai-system-prompt
duplicate-issue-ai-user-prompt
duplicate-issue-ai-temperature
duplicate-issue-similarity-threshold
```
env var
```
OPENAI_API_HOST
OPENAI_API_KEY=
```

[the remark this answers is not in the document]
pavlovcik: > Eventually, openai here will be used for other purposes too. The following can be feature-specific config stored as repo config
> 
> ```
> 
> duplicate-issue-ai-system-prompt
> 
> duplicate-issue-ai-user-prompt
> 
> duplicate-issue-ai-temperature
> 
> duplicate-issue-similarity-threshold
> 
> ```
> 
> 
> 
> What do you think @pavlovcik 

I think we can consider specific settings (prompt, temperature) for specific capabilities. Like the ones that it makes sense to customize like this one proposal to give the bot a "personality" 

However for this duplicates one I agree no need to expose properties for customization.

.env.example
```diff
@@ -13,3 +13,8 @@ ANALYTICS_MODE=
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=debug
 LOGDNA_INGESTION_KEY=
+OPENAI_API_HOST=https://api.openai.com
+OPENAI_API_KEY=
+CHATGPT_SYSTEM_PROMPT_FOR_IMPORTANT_WORDS="You are an 'important words finder'. You need to find important words from given context. You only have to give important words from given context and you have to separate the words by #"
+CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS="I need your help to find duplicate issues on my GitHub repository. For context, the entire strategy is the following:\n\n1. A new issue is posted\n2. We ask you to extract a word list of the most \"important\" (i.e. unique adjectives?) words.\n3. We search the repository for all issues with the important words.\n4. We go from highest issue number (most recent) and read the issue description.\n5. If >80% confidence that it's a redundant issue, stop the search and link back to it with a warning saying that it's likely to be a duplicate.\nRight now, we are on step 2.\nPlease separate the words by '#' so I can parse them easily. Please answer simply as I only need the important words."
+SIMILARITY_THRESHOLD=80
```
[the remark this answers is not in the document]
pavlovcik: I don't think prompts need to be customizable for this feature.

src/helpers/similarity.ts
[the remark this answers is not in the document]
pavlovcik: > I see, I think it would be better if we use `filter` function to explicitly remove the current issue.

Why is that? Isn't shift the most efficient?

src/helpers/similarity.ts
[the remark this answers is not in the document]
pavlovcik: > This checks for all past issues...

It seems like you implemented a different strategy than what is described in the specification. I believe it makes sense to stop the search as soon as a similar issue is found, which will save on API calls?

pavlovcik:
Great reviews, team!

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
+import { getBotContext } from "../bindings";
+import { Payload, Choices } from "../types";
+import { listAllIssuesForRepo, upsertCommentToIssue } from "../helpers/issue";
+import axios from "axios";
+
+export const findDuplicateOne = async () => {
+  const { payload: _payload } = getBotContext();
+  const issue = (_payload as Payload).issue;
+  const issues = await listAllIssuesForRepo("all");
```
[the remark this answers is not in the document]
ByteBallet: RFC @pavlovcik

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+};
+
+const countIncludedWords = (inputString: string, words: string[]): number => {
+  const lowerCaseString = inputString.toLowerCase(); // Convert the string to lowercase for case-insensitive matching
+
+  let count = 0;
+  for (const word of words) {
+    if (lowerCaseString.includes(word.toLowerCase())) {
```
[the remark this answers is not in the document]
ByteBallet: Yes, indexOf is a a tiny, tiny amount faster than includes but in reality this will make no difference. Anyway, I will change this.

.env.example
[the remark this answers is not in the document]
ByteBallet: > I don't think prompts need to be customizable for this feature.

So shall I remove from env and leave them just as constants? @pavlovcik

src/helpers/similarity.ts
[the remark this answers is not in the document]
whilefoo: In that case `listAllIssuesForRepo` should be fixed to filter out PRs because the name implies only issues.

Currently we already store issues as a part of analytics so we can change to store it when the issue is opened and add description to the schema. Then we can query all of the issues in an efficient manner.

src/helpers/similarity.ts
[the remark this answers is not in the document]
whilefoo: > Why is that? Isn't shift the most efficient?

That's true, but we can't know for certain the current issue will be first in the array - what if two issues get created at the same time?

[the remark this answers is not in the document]
ByteBallet: > > Why is that? Isn't shift the most efficient?
> 
> That's true, but we can't know for certain the current issue will be first in the array - what if two issues get created at the same time?

Oh, I didn't get you. Is that possible that two same issues are created at the same time? Even though in that case, isn't it normal to alert about the second issue?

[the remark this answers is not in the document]
whilefoo: Yes, but we don't know if they are the same, maybe just created at the same time but different content. So by using `filter` we exclude the current issue and compare with all other issues.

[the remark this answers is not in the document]
ByteBallet: I am changing the strategy as @pavlovcik mentioned so this will not be a problem.

src/helpers/similarity.ts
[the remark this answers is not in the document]
ByteBallet: > I think we can consider specific settings (prompt, temperature) for specific capabilities. Like the ones that it makes sense to customize like [this one proposal](https://github.com/ubiquity/research/issues/58) to give the bot a "personality"
> 
> (It's in the research repository because I want to discuss the concept on a call with the research team before we make it official.)
> 
> Perhaps it could be useful to expose properties for customization in case repository maintainers want to add additional context to enhance duplicate issue finding accuracy on their specific repositories (e.g. explicitly listing out the "important" words.)

I am a bit confused now about where to store each keys. So should OPENAI_API_KEY and OPENAI_API_HOST be stored in .env or /config/shared.ts ?
Other similarity configs including duplicate-issue-ai-system-prompt, duplicate-issue-ai-user-prompt, duplicate-issue-ai-temperature, duplicate-issue-similarity-threshold
should be stored in repo config?
@pavlovcik RFC

.env.example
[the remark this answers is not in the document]
pavlovcik: > > I don't think prompts need to be customizable for this feature.
> 
> So shall I remove from env and leave them just as constants? @pavlovcik

I changed my mind because I realized that repository maintainers can increase the accuracy of this feature by providing additional context on what their project is about and perhaps even specify specific words to look out for lol

src/helpers/similarity.ts
```diff
@@ -0,0 +1,86 @@
...
+      Authorization: `Bearer ${process.env.OPENAI_API_KEY}`,
+      "Content-Type": "application/json",
+    },
+    data: body,
+  };
+  try {
+    const response = await axios(config);
+    const data: Choices = response.data;
+    const { choices: choice } = data;
+    const answer = choice[0].message.content;
```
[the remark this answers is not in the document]
pavlovcik:
```suggestion
    const answer = choice[0]?.message.content;
```

Like that/

src/helpers/similarity.ts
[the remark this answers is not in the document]
pavlovcik: @0xcodercrane knows best

src/helpers/similarity.ts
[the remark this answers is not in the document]
pavlovcik: Perhaps it makes sense to focus on code readability vs optimizing for marginal performance benefits (I think this would make more sense if we were looping in the 100,000+ times range. Perhaps we can focus on performance optimizations as needed. In the meantime lets maintain `includes` for readability reasons.

pavlovcik:
@ByteBallet please be sure to "Resolve conversation" for each after you implement requested changes, and have all of your questions answered.

ByteBallet:
> @ByteBallet please be sure to "Resolve conversation" for each after you implement requested changes, and have all of your questions answered.

Yes, sure.

src/helpers/similarity.ts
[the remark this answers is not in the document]
ByteBallet: @0xcodercrane rfc

src/helpers/similarity.ts
ByteBallet: I think parsing every issues will consume some more time. @whilefoo

src/handlers/issue/pre.ts
```diff
@@ -0,0 +1,33 @@
...
+  const perPage = 10;
+  let curPage = 1;
+  issue;
+
+  while (!fetchDone) {
+    const issues = await listIssuesForRepo("all", perPage, curPage, "created", "asc");
+    for (const iss of issues) {
+      if (iss.body && iss.number != issue.number && !iss.pull_request) {
+        const probability = (countIncludedWords(iss.body, importantWords) * 100.0) / wordCount;
+        if (probability > parseInt(process.env.SIMILARITY_THRESHOLD || "80")) {
```
pavlovcik, line 22: I'm concerned about the probability calculation strategy. 

We should be leveraging ChatGPT's ability to interpret synonyms. It appears that you're manually calculating this based on the frequency of specific words being used.

ByteBallet: I am not sure about this. Do you want ChatGPT to find all similar words from word list in the issues? Assuming there are 1000 past issues, then we need to send 1001 requests to OpenAI. Or do you want ChatGPT to make all possible words from important words list?

pavlovcik: Well, my original vision is something more like getting a batch of the issue specifications (let's say whatever is close to maxing out the context window of the LLM, starting from latest issue number) and asking GPT to compare the newly created issue with every other issue in that batch. Then if it thinks there is a duplicate, to post a comment back to the duplicate issue.

ByteBallet: > We could do things without a vector database in order to simplify things, here's how:

> A new issue is posted
> We ask ChatGPT to extract a word list of the most "important" (i.e. unique adjectives?) words (for example, when I wanted to find this issue, I searched for the term "duplicate")
> We search the repository for all issues with the important words
> We go from highest issue number (most recent) and read the specification.
> If >80% confidence, stop the search and link back to it with a warning saying that it's likely to be a duplicate.
> This approach might be a little more brittle (rate limits) but we won't have to worry about the database which should make implementation and maintenance much easier.

But I think we decided to measure similarity by rate of important words's appearance. Not asking ChatGPT to compare the newly created issue and other issues.

pavlovcik: Perhaps there was a misunderstanding but refer to [this explanation](https://github.com/ubiquity/ubiquibot/issues/488#issuecomment-1668970677) of my perspective.

1. I parse out the "important" words of the newly created issue e.g. "duplicate"
2. Then I search those important words on GitHub issues search
3. Which will return me a list of all the specifications of somewhat similar issues (i believe this returned seven issues, which all have the keyword "duplicate")
4. Use ChatGPT to interpret the meaning of all seven specifications and compare to the current created one to see if it is describing the same thing.

ByteBallet: Then what is the mechanism to measure similarity? @pavlovcik 
Search issues for the top three important words and we can just only ask ChatGPT if the searched issues are similar to the new one, not similarity but yes or no.

pavlovcik: I think I'm okay with boolean over percentage IF 
1. We can not figure out how to prompt it to return confidence percentage and 
2. Are leveraging chatGPT's ability to understand meaning vs trying to manually calculate a percentage, throwing out all the advanced interpretation capabilities of ChatGPT

ByteBallet: I will try to make prompts to figure out percentage.

byteballet: I tried this kind of prompt: "I have two github issues.\n%first%\nand\n%second%\nPlease give me the possibility of the 2 issues are the same content.\n Give me in number format and add % after the number.\nDo not tell other things since I only need the number."
It seems GPT returns me the possibility in %.
But the problem here is we are comparing found issues with newly created one and the OpenAI returns me 429 (request rate limit) error.
Are we going to use paid OpenAI API key?

src/helpers/similarity.ts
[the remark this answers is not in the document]
0xcodercrane: you should store them in `.env`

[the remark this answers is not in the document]
byteballet: Okay, got it.

src/helpers/similarity.ts
byteballet: Changed strategy and measure similarity using ChatGPT. I think this is resolved.

.env.example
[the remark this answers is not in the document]
byteballet: Stored configs to .env according to @0xcodercrane

src/handlers/issue/pre.ts
byteballet: QA for byteballet/santa-bringyouwishes#75

pavlovcik: > Are we going to use paid OpenAI API key? 

We're going to try to get as far as we can without paying. 

When I searched in my previous example I shared above, there were only seven issues to compare, which is really not a lot

whilefoo: I think API has only paid tier, no free.

byteballet: OpenAI Api has free plan.
image

whilefoo: Those are rate limits. I think for [free trial](https://help.openai.com/en/articles/4936830-what-happens-after-i-use-my-free-tokens-or-the-3-months-is-up-in-the-free-trial) you get 5$ and that's all

pavlovcik:
I was thinking about the rate limit problem. What if we "preserve state" by passing in the progress to the database. For example:

```js
{ 

checking: { url: "ubiquity/ubiquibot/520", startTime: "1234123412" },

queue: [

{pending: false, similarity: 0.7, issueUrl: "ubiquity/ubiquibot/510"}

{pending: true, issueUrl: "ubiquity/ubiquibot/511"},

{pending: true, issueUrl: "ubiquity/ubiquibot/512"}

] }
``` 

If we hit rate limits, this job can continue the next time an issue is posted?

(Sorry for the sloppy example I wrote this on my phone.)

byteballet:
>this job can continue the next time an issue is posted?

But the problem is we can also hit rate limit while checking the next issue. If this process persists, I think we will have a lot of delay.

pavlovcik:
> But the problem is we can also hit rate limit while checking the next issue. If this process is repeated, I think we will have a lot of delay.

In the original specification I explicitly stated that slow is fine. Even if it takes a day its still immensely helpful for repositories with hundreds of open issues and community contributors opening new ones.
> that passes the issue specification to ChatGPT (asynchronously/slowly is fine)

byteballet:
> In the original specification I explicitly stated that slow is fine. Even if it takes a day its still immensely helpful for repositories with hundreds of open issues and community contributors opening new ones.

https://github.com/ayaka14732/ChatGPTAPIFree
I was looking for a new method and found this project on github.

pavlovcik:
Exponential backoff is pretty standard. We can use it.

byteballet:
> Exponential backoff is pretty standard. We can use it.

QA for https://github.com/byteballet/santa-bringyouwishes/issues#86
Used exponential-backoff

pavlovcik:
How long for the bot to react? I posted a real world example of a duplicate at https://github.com/byteballet/santa-bringyouwishes/issues/89 and https://github.com/byteballet/santa-bringyouwishes/issues/91

byteballet:
> How long for the bot to react? I posted a real world example of a duplicate at [byteballet/santa-bringyouwishes#89](https://github.com/byteballet/santa-bringyouwishes/issues/89) and [byteballet/santa-bringyouwishes#91](https://github.com/byteballet/santa-bringyouwishes/issues/91)

QA for https://github.com/byteballet/santa-bringyouwishes/issues/107
Could you check please? @pavlovcik

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+        .replace("%first%", first)
+        .replace("%second%", second)}`,
+    parseFloat(process.env.MEASURE_SIMILARITY_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
+  const percent = res.split("%")[0].split(" ").pop();
+  if (!percent) {
+    return 0;
+  } else {
+    return parseFloat(percent);
```
whilefoo, line 37: `parseFloat` can return `NaN` if something is wrong, so you should handle that

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+      ${(
+        process.env.CHATGPT_USER_PROMPT_FOR_MEASURE_SIMILARITY ||
+        'I have two github issues and I need to measure the possibility of the 2 issues are the same content (I need to parse the % so other contents are not needed and give me only the number in %).\n Give me in number format and add % after the number.\nDo not tell other things since I only need the number (e.g. 85%). Here are two issues:\n 1. "%first%"\n2. "%second%"'
+      )
+        .replace("%first%", first)
+        .replace("%second%", second)}`,
+    parseFloat(process.env.MEASURE_SIMILARITY_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
+  const percent = res.split("%")[0].split(" ").pop();
```
whilefoo, line 33: why is `.split(" ")` needed?

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+  const res = await getAnswerFromChatGPT(
+    '',
+    `${
+      process.env.CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS ||
+      "I need your help to find important words (e.g. unique adjectives) from github issue below and I want to parse them easily so please separate them using #(No other contexts needed). Please separate the words by # so I can parse them easily. Please answer simply as I only need the important words. Here is the issue content.\n"
+    } '${content}'`,
+    parseFloat(process.env.IMPORTANT_WORDS_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
+  return res.split(/[,# ]/);
```
whilefoo, line 17: if `res` is `""` then you should return empty array because split on empty string will be `[""]`

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+    const data: Choices = response.data;
+    const validate = ajv.compile(ChoicesSchema);
+    const valid = validate(data);
+    if (!valid) {
+      throw new Error("Error occured from OpenAI");
+    }
+    const { choices: choice } = data;
+    if (choice.length <= 0) {
+      return "";
```
whilefoo, line 98: log an error so we know something is wrong because this should not happen normally

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+      retry: (e: AxiosError) => {
+        if (e.response && e.response.status === 429) return true;
+        return false;
+      },
+    });
+    const data: Choices = response.data;
+    const validate = ajv.compile(ChoicesSchema);
+    const valid = validate(data);
+    if (!valid) {
+      throw new Error("Error occured from OpenAI");
```
whilefoo, line 94: also log an error

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+      "I need your help to find important words (e.g. unique adjectives) from github issue below and I want to parse them easily so please separate them using #(No other contexts needed). Please separate the words by # so I can parse them easily. Please answer simply as I only need the important words. Here is the issue content.\n"
+    } '${content}'`,
+    parseFloat(process.env.IMPORTANT_WORDS_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
+  return res.split(/[,# ]/);
+};
+
+export const measureSimilarity = async (first: string, second: string): Promise<number> => {
+  const res = await getAnswerFromChatGPT(
```
whilefoo, line 21: `getAnswerFromChatGPT` can throw an error so either don't throw an error and return `""` like in other cases or handle the error here

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+      process.env.CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS ||
+      "I need your help to find important words (e.g. unique adjectives) from github issue below and I want to parse them easily so please separate them using #(No other contexts needed). Please separate the words by # so I can parse them easily. Please answer simply as I only need the important words. Here is the issue content.\n"
+    } '${content}'`,
+    parseFloat(process.env.IMPORTANT_WORDS_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
+  return res.split(/[,# ]/);
+};
+
+export const measureSimilarity = async (first: string, second: string): Promise<number> => {
```
whilefoo, line 20:
```suggestion
export const measureSimilarity = async (first: Issue, second: Issue): Promise<number> => {
```

I think it's better to let the function extract whatever it needs from the issue

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
+import { getLogger } from "../bindings";
+import axios, { AxiosError } from "axios";
+import { ajv } from "../utils";
+import { Static, Type } from "@sinclair/typebox";
+import { backOff } from "exponential-backoff";
+
+export const extractImportantWords = async (content: string): Promise<string[]> => {
```
whilefoo, line 7:
```suggestion
export const extractImportantWords = async (issue: Issue): Promise<string[]> => {
```

Let the function extract what it needs

src/helpers/similarity.ts
```diff
@@ -0,0 +1,106 @@
...
+  const res = await getAnswerFromChatGPT(
+    '',
+    `${
+      process.env.CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS ||
+      "I need your help to find important words (e.g. unique adjectives) from github issue below and I want to parse them easily so please separate them using #(No other contexts needed). Please separate the words by # so I can parse them easily. Please answer simply as I only need the important words. Here is the issue content.\n"
+    } '${content}'`,
+    parseFloat(process.env.IMPORTANT_WORDS_AI_TEMPERATURE || "0")
+  );
+  console.log(res);
```
whilefoo, line 16:
```suggestion
```
this is probably accidentally commited?

whilefoo (changes requested):

src/helpers/similarity.ts
byteballet: Changed to regular expression.

src/helpers/similarity.ts
```diff
@@ -0,0 +1,109 @@
...
+        .replace("%second%", `Issue title: ${second.title}\nIssue content: ${second.body}`)}`,
+    parseFloat(process.env.MEASURE_SIMILARITY_AI_TEMPERATURE || "0")
+  );
+  const matches = res.match(/\d+/);
+  const percent = matches && matches.length > 0 ? parseInt(matches[0]) : 0;
+  if (!percent) {
+    return 0;
+  } else {
+    return percent;
+  }
```
whilefoo, line 39: this if/else statement can be simplified to `return percent || 0`

src/handlers/issue/pre.ts
```diff
@@ -0,0 +1,49 @@
...
+              );
+              return;
+            }
+          }
+        }
+        if (response.data.items.length < perPage) fetchDone = true;
+        else curPage++;
+      }
+    } catch (e: unknown) {
+      logger.debug(`Could not find any issues, reason: ${e}`);
```
whilefoo, line 46:
```suggestion
      logger.error(`Could not find any issues, reason: ${e}`);
```

whilefoo (changes requested):

whilefoo (changes requested):

src/helpers/similarity.ts
```diff
@@ -0,0 +1,105 @@
...
+      ${(
+        process.env.CHATGPT_USER_PROMPT_FOR_MEASURE_SIMILARITY ||
+        'I have two github issues and I need to measure the possibility of the 2 issues are the same content (I need to parse the % so other contents are not needed and give me only the number in %).\n Give me in number format and add % after the number.\nDo not tell other things since I only need the number (e.g. 85%). Here are two issues:\n 1. "%first%"\n2. "%second%"'
+      )
+        .replace("%first%", `Issue title: ${first.title}\nIssue content: ${first.body}`)
+        .replace("%second%", `Issue title: ${second.title}\nIssue content: ${second.body}`)}`,
+    parseFloat(process.env.MEASURE_SIMILARITY_AI_TEMPERATURE || "0")
+  );
+  const matches = res.match(/\d+/);
+  const percent = matches && matches.length > 0 ? parseInt(matches[0]) : 0;
```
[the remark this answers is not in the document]
byteballet: Sorry...

byteballet:
QA for https://github.com/byteballet/santa-bringyouwishes/issues/114

src/helpers/similarity.ts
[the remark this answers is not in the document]
pavlovcik: > `parseInt` can return `NaN`

Wouldn't TypeScript catch that? Perhaps your suggestion is a non-issue?

.env.example
```diff
@@ -13,3 +13,10 @@ ANALYTICS_MODE=
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=debug
 LOGDNA_INGESTION_KEY=
+OPENAI_API_HOST=https://api.openai.com
+OPENAI_API_KEY=
+CHATGPT_USER_PROMPT_FOR_IMPORTANT_WORDS="I need your help to find important words (e.g. unique adjectives) from github issue below and I want to parse them easily so please separate them using #(No other contexts needed). Please separate the words by # so I can parse them easily. Please answer simply as I only need the important words. Here is the issue content.\n"
+CHATGPT_USER_PROMPT_FOR_MEASURE_SIMILARITY='I have two github issues and I need to measure the possibility of the 2 issues are the same content (No other contents needed and give me only the number in %).\n Give me in number format and add % after the number.\nDo not tell other things since I only need the number (e.g. 85%). Here are two issues:\n 1. "%first%"\n2. "%second%"'
+SIMILARITY_THRESHOLD=80
```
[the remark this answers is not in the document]
pavlovcik: - To make it clear, it should be stored in `ubiquibot-config.json` because our partners should be able to configure these to accommodate per repository. This is because certain partners might be able to tweak the config to increase similarity accuracy. https://github.com/ubiquity/ubiquibot/pull/608/files#r1292178783
- If it is in `.env` that implies that we must change it in our production instance Netlify environment config, which affects every partner.

.github/workflows/bot.yml
pavlovcik, line 1: This should probably be reverted before merging, right?

src/helpers/similarity.ts
```diff
@@ -0,0 +1,105 @@
...
+export const measureSimilarity = async (first: Issue, second: Issue): Promise<number> => {
+  const res = await getAnswerFromChatGPT(
+    "",
+    `
+      ${(
+        process.env.CHATGPT_USER_PROMPT_FOR_MEASURE_SIMILARITY ||
+        'I have two github issues and I need to measure the possibility of the 2 issues are the same content (I need to parse the % so other contents are not needed and give me only the number in %).\n Give me in number format and add % after the number.\nDo not tell other things since I only need the number (e.g. 85%). Here are two issues:\n 1. "%first%"\n2. "%second%"'
+      )
+        .replace("%first%", `Issue title: ${first.title}\nIssue content: ${first.body}`)
+        .replace("%second%", `Issue title: ${second.title}\nIssue content: ${second.body}`)}`,
```
pavlovcik, line 30: This is an unusual, but acceptable, replace mechanism.

Why not use string literals?

```typescript
const myVariable = "testing";
const myStringLiteral = `my string literal ${myVariable}`;
// "my string literal testing"
```

pavlovcik (approved):
If there are any more small requested changes, assuming that this current commit is stable, let's create new bounties for them. I am ready to merge this in!

src/helpers/similarity.ts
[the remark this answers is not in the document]
whilefoo: Actually no, `NaN` is not a separate type, but is of type `Number` and you can still do operations like `NaN * 5` is perfectly valid

src/helpers/similarity.ts
whilefoo: I agree, it would be better if it's done the same way as [here](https://github.com/ubiquity/ubiquibot/pull/608/files#diff-289ca7c6ad34c7929d5662f47d65edd1bbd87b256689fd2cd475d61153f10cb1R14)

src/helpers/similarity.ts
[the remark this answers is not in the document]
byteballet: Actually, I got matches using regex and I don't think there is a chance for parseInt to be NaN but I changed it for readability.

pavlovcik (approved):

Draeieg:
haven't been able to duplicate the QA on [staging](https://github.com/ubiquibot/staging/issues/127)

I assume we're missing the API key as and Environment variable in our preview @pavlovcik @byteballet

byteballet:
> haven't been able to duplicate the QA on [staging](https://github.com/ubiquibot/staging/issues/127)
> 
> I assume we're missing the API key as and Environment variable in our preview @pavlovcik @byteballet

Yes, maybe it will be missing as .env is in gitignore.

pavlovcik:
> haven't been able to duplicate the QA on [staging](https://github.com/ubiquibot/staging/issues/127)
> 
> 
> 
> I assume we're missing the API key as and Environment variable in our preview @pavlovcik @byteballet 

Are you able to make one on open ai? Otherwise you will need to wait until one of the others can do that and add it. I can't right now.

byteballet:
> 
> Are you able to make one on open ai? Otherwise you will need to wait until one of the others can do that and add it. I can't right now.
> 
> @byteballet I believe you also have access (@ubiquity-bounties) so please feel free to adjust the secrets on the QA instance.

I added some env variables for only previews. Maybe need to change webhook URL or create a new app to test it? @Draeieg

Draeieg:
> > haven't been able to duplicate the QA on [staging](https://github.com/ubiquibot/staging/issues/127)
> > 
> > 
> > 
> > I assume we're missing the API key as and Environment variable in our preview @pavlovcik @byteballet 
> 
> Are you able to make one on open ai? Otherwise you will need to wait until one of the others can do that and add it. I can't right now. 
> 
> @byteballet I believe you also have access (@ubiquity-bounties) so please feel free to adjust the secrets on the QA instance. 

OpenAI has my country blocked so I'm unable to make an account unless someone lends me their info lol

  @byteballet  plz gib key

byteballet:
> > > haven't been able to duplicate the QA on [staging](https://github.com/ubiquibot/staging/issues/127)
> > > I assume we're missing the API key as and Environment variable in our preview @pavlovcik @byteballet
> > 
> > 
> > Are you able to make one on open ai? Otherwise you will need to wait until one of the others can do that and add it. I can't right now.
> > @byteballet I believe you also have access (@ubiquity-bounties) so please feel free to adjust the secrets on the QA instance.
> 
> OpenAI has my country blocked so I'm unable to make an account unless someone lends me their info lol
> 
> @byteballet plz gib key

On github issue???
I added here: https://app.netlify.com/sites/ubiquibot-staging/configuration/env#OPENAI_API_KEY

byteballet:
> OpenAI has my country blocked so I'm unable to make an account unless someone lends me their info lol
> 
> @byteballet plz gib key

Please check all the keys added.
I added about 5 keys...

Draeieg:
> > OpenAI has my country blocked so I'm unable to make an account unless someone lends me their info lol
> > @byteballet plz gib key
> 
> Please check all the keys added. I added 7 keys...

still [nothing ](https://github.com/ubiquibot/staging/issues/132)

Draeieg:
> > > > OpenAI has my country blocked so I'm unable to make an account unless someone lends me their info lol
> > > > @byteballet plz gib key
> > > 
> > > 
> > > Please check all the keys added. I added 7 keys...
> > 
> > 
> > still [nothing ](https://github.com/ubiquibot/staging/issues/132)
> 
> As I know, time limit for serverless functions are set to 10 for netlify, vercel and for others. We can increase it. image image image

@pavlovcik @0xcodercrane

web4er:
I have tried to QA this PR. 

In this repo, I created 5 issues. 1st and last ones are duplicates. Rest are not. https://github.com/web4erOrg/AiDupCheckRepo/issues

I modified the bot to also comment important words it picked, and show a similarity score if there is any. 

It didn't find a duplicate and the similarity score was 1 for all. This needs improvement in prompt. The prompt is forcing chatGPT to reply with number, but chatGPT is not replying with a number, it is replying with this : 

> 
> To measure the possibility of the two GitHub issues being the same content, we can use a text similarity metric such as Levenshtein distance. This metric calculates the minimum number of single-character edits (insertions, deletions, or substitutions) required to change one string into another.
> 
> Using this approach, we can calculate the similarity percentage between the two issues. Here is the calculation:
> 
> 1. Calculate the Levenshtein distance between the two issues.
> 2. Divide the Levenshtein distance by the length of the longer issue.
> 3. Subtract the result from 1 to get the similarity percentage.
> 4. Multiply the similarity percentage by 100 to get the final result.
> 
> Please provide the actual content of the two issues ("%first%" and "%second%") so that we can calculate the similarity percentage for you.
> 

That is why all the similarity score are 1, because ` res.match(/\d+/);` is returning `[1]` for all.

ChatGPT also consumed 1 cent for these answers using API.

However, I also tried all these 5 issues directly with chatGPT and got interesting results. Even though it gave a different similarity score on different tries, but similarity score of the duplicate was higher. 

https://chat.openai.com/share/8a97e9da-a8ce-45e9-8bf0-9bdd5b45f643

pavlovcik:
Oh man we have to start charging for these features. I'll DM you some card info @Draeieg

byteballet:
@web4er 
Did you test with the latest changes?
As I modified a bit on prompt for measuring similarity.
https://github.com/ubiquity/ubiquibot/pull/608/commits/6f86e2411c041b7fc9a28f77ef0441a625858c2b
It might be working fine on local but it will cause error on netlify as it limits time of serverless function.

byteballet:
https://github.com/ubiquibot/staging/issues/141
QA for deployed branch on Netlify.
It worked but I think increasing time limit will be necessary further...
@pavlovcik

byteballet:
I am not sure if it is necessary to increase time limit for more than 10 seconds as it seems working fine...

web4er:
> Did you test with the latest changes?

I have tested it at https://github.com/ubiquity/ubiquibot/pull/608/commits/93818b42556d1f3add070a1ecbd63b85707bd40b that was latest when I was doing it.

I think the point of concern in the QA was that you are always expecting a number from AI, and if it responds with some text, you are parsing any number from the answer as a similarity.

byteballet:
> 
> I have tested it at [93818b4](https://github.com/ubiquity/ubiquibot/commit/93818b42556d1f3add070a1ecbd63b85707bd40b) that was latest when I was doing it.
> 
> I think the point of concern in the QA was that you are always expecting a number from AI, and if it responds with some text, you are parsing any number from the answer as a similarity.

Since we don't want 100% correctness from AI, I think it's reasonable to do like that. It's just the problem that we concern more on number of % or bot's answering rate. And in my case, I didn't experience the bot answering with number that is irrelevant to the similarity.

pavlovcik:
> @web4er 
> 
> Did you test with the latest changes?
> 
> As I modified a bit on prompt for measuring similarity.
> 
> https://github.com/ubiquity/ubiquibot/pull/608/commits/6f86e2411c041b7fc9a28f77ef0441a625858c2b
> 
> It might be working fine on local but it will cause error on netlify as it limits time of serverless function.
> 
> 

You shouldn't have undone the string literal. Also consider the prompt I wrote because I got really good results in my limited testing:

byteballet:
Please check the code carefully. I couldn't use string literal as I used the prompt from .env file. First, I used exactly same prompt from yours but gpt did not give me satisfactory result. It always gave me words from the prompt, not content. (e.g. It always gave me "important" as important word)

byteballet:
> Also consider the prompt I wrote because I got really good results for the keywords in my limited testing: [#488 (comment)](https://github.com/ubiquity/ubiquibot/issues/488#issuecomment-1668940692)

If we want, we can just change prompts from env variable.

byteballet:
What do you think about it? @pavlovcik

.env.example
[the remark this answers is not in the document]
pavlovcik: No they should be in `ubiquibot-config.json` so that each partner can adjust on their own instance.

Draeieg:
as I understand it currently we have 2 options for QA on  previews

1) get rugged by temporarily pay for the premium servers 
2) wait until #652 

or 3rd option do local testing like @web4er and trust me bro  it will run fine once we figure a solution

byteballet:
> as I understand it currently we have 2 options for QA on previews
> 
> 1. get rugged by temporarily pay for the premium servers
> 2. wait until [Similar Issue Detection: Minimize Runtime #652](https://github.com/ubiquity/ubiquibot/issues/652)
> 
> or 3rd option do local testing like @web4er and trust me bro it will run fine once we figure a solution

Or just test again on staging, I think it will be working. ^^
Although the time limit of functions will be problem further, I guess no problem in QA (https://github.com/ubiquibot/staging/issues/141).

byteballet:
Or qa for https://github.com/ubiquibot/staging/issues/158

pavlovcik:
> 1. get rugged by temporarily pay for the premium servers

I gave you card info to set it up its fine for now.

src/handlers/processors.ts
```diff
@@ -8,10 +8,11 @@ import { checkPullRequests } from "./assign/auto";
 import { createDevPoolPR } from "./pull-request";
 import { runOnPush } from "./push";
 import { incentivizeComments, incentivizeCreatorComment } from "./payout";
+import { findDuplicateOne } from "./issue";
 
 export const processors: Record<string, Handler> = {
   [GithubEvent.ISSUES_OPENED]: {
-    pre: [nullHandler],
+    pre: [findDuplicateOne],
```
pavlovcik, line 15: @0xcodercrane do you think it makes sense that it is in the `GithubEvent.ISSUES_OPENED[pre]` hook instead of adding it into the `GithubEvent.ISSUES_OPENED[action]` array?

byteballet: @0xcodercrane rfc

0xcodercrane: @pavlovcik is right. Let me push and merge the PR

src/handlers/processors.ts
```diff
@@ -8,10 +8,11 @@ import { checkPullRequests } from "./assign/auto";
...
     action: [nullHandler], // SHOULD not set `issueCreatedCallback` until the exploit issue resolved.  https://github.com/ubiquity/ubiquibot/issues/535
```
0xcodercrane, line 16:
```suggestion
    pre: [nullHandler],
    action: [findDuplicateOne], // SHOULD not set `issueCreatedCallback` until the exploit issue resolved.  https://github.com/ubiquity/ubiquibot/issues/535
```

0xcodercrane (changes requested):

src/handlers/processors.ts
0xcodercrane: https://github.com/ubiquity/ubiquibot/pull/608/commits/89e670562e89856c6cb8607cd6a39696f8bfeffa

0xcodercrane (approved):

PR #611, opened 2023-08-12
fix: query regex

### LINKED ISSUE #590: Invalid syntax for query command
### Describe the background or context

I experimented today with the 'query' command after enabling in my fork always received 

`Invalid syntax for query command` error, see https://github.com/gitcoindev/ubiquibot/pull/3

I was able to run this command only because I analysed the regex in https://github.com/ubiquity/ubiquibot/blob/development/src/handlers/comment/handlers/query.ts

By default GitHub adds an empty space after username after typing @username and this is rejected as the regex does not allow any white space after user name.

### Describe the solution

Proposal to fix this: Update the regex to allow to add an optional empty white space after user name.

EtherealGlow:
Resolves #590

### DIFF
--- a/src/handlers/comment/handlers/query.ts
+++ b/src/handlers/comment/handlers/query.ts
@@ -19,7 +19,7 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /(?<=\/query @)\w+/g;
+  const regex = /(?<=\/query @)\w+/g
   const matches = body.match(regex);
   const user = matches?.[1];
 
--- a/src/handlers/comment/handlers/query.ts
+++ b/src/handlers/comment/handlers/query.ts
@@ -19,9 +19,9 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /(?<=\/query @)\w+/g
+  const regex = /^\/query\s+(@\w+)\s?$/;
   const matches = body.match(regex);
-  const user = matches?.[1];
+  const user = matches?.[1].substring(1);
 
   if (user) {
     const walletInfo = await getWalletInfo(user, id?.toString());
--- a/src/handlers/comment/handlers/query.ts
+++ b/src/handlers/comment/handlers/query.ts
@@ -19,7 +19,7 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /\/query @(\w+)/gm;
+  const regex = /(?<=\/query @)\w+/g;
   const matches = body.match(regex);
   const user = matches?.[1];
 
--- a/src/handlers/comment/handlers/query.ts
+++ b/src/handlers/comment/handlers/query.ts
@@ -19,7 +19,7 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /^\/query\s@(\w+)$/;
+  const regex = /\/query @(\w+)/gm;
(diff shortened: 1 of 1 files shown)


EtherealGlow:
i tested this using a ts debugger but i cannot qa this unfortunately  as i am getting a module not found error and idk why

EtherealGlow:
here is the result from the [debugger](https://regex101.com/r/tajdDM/3)

src/handlers/comment/handlers/query.ts
```diff
@@ -19,7 +19,7 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /^\/query\s@(\w+)$/;
+  const regex = /(?<=\/query @)\w+/g
```
whilefoo, line 22: this regex is different than your debugger and the positive lookbehind seems rather unnecessary, this can be done with simple capturing groups.

also please make sure it still matches if you put multiple spaces between like `/query    @test`

src/handlers/comment/handlers/query.ts
```diff
@@ -19,7 +19,7 @@ export const query = async (body: string) => {
...
   const matches = body.match(regex);
```
whilefoo, line 23: `match` function behaves [differently](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value) when a regex is global like yours `/g`. I think we don't need global regex because we are just searching for one instance of query

whilefoo (changes requested):

src/handlers/comment/handlers/query.ts
```diff
@@ -19,9 +19,9 @@ export const query = async (body: string) => {
     return `Skipping '/query' because of no issue instance`;
   }
 
-  const regex = /^\/query\s@(\w+)$/;
+  const regex = /^\/query\s+(@\w+)\s?$/;
```
whilefoo, line 22:
```suggestion
  const regex = /^\/query\s+@([\w-]+)\s*$/;
```

if you put `@` out of the capture group you don't need substring, username can contain `-` and we can allow for zero or multiple whitespace chars at the end

src/handlers/comment/handlers/query.ts
```diff
@@ -19,9 +19,9 @@ export const query = async (body: string) => {
...
   const matches = body.match(regex);
-  const user = matches?.[1];
+  const user = matches?.[1].substring(1);
```
whilefoo, line 24:
```suggestion
  const user = matches?.[1];
```

whilefoo (changes requested):

web4er (changes requested):
> i tested this using a ts regex debugger but i cannot qa this unfortunately as i am getting a module not found error and idk why

This should have a QA. Maybe try the following if you haven't already. 
```sh
yarn clean
yarn 
yarn build 
yarn start
```

EtherealGlow:
@web4er it is finding the user now but i cant set the address

EtherealGlow:
[qa](https://github.com/EtherealGlow/ubiquibot/issues/10)

src/handlers/comment/handlers/query.ts
pavlovcik: We can simplify everything by just taking the comment corpus and trimming it `String.trim();` when it enters this function?

whilefoo (approved):

pavlovcik:
@web4er just needs your approval for merge

web4er (approved):
> @web4er just needs your approval for merge

QA: https://github.com/web4erOrg/ubiquibot/issues/32

Working ok.

PR #615, opened 2023-08-14
Edit instead of double post (initial pull request)

### LINKED ISSUE #601: Edit Instead of Double Post

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/597#issuecomment-1673497830_

byteballet:
Resolves #601 
QA (I am not sure how I can make the bot 'double post')
I basically implemented logic for "edit instead of double post".
The logic is to check if last comment is bot and if so just update comment instead of adding.
image

### DIFF
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -598,6 +598,6 @@ export const getAvailableOpenedPullRequests = async (username: string) => {
 export const checkIfLastCommentIsBot = async (issue_number: number) => {
   const comments = await getAllIssueComments(issue_number);
   const lastComment = comments.pop();
-  if(!lastComment) return false;
-  return lastComment.user.type === UserType.Bot
+  if (!lastComment) return false;
+  return lastComment.user.type === UserType.Bot;
 };
--- a/src/helpers/issue.ts
+++ b/src/helpers/issue.ts
@@ -1,6 +1,6 @@
 import { Context } from "probot";
 import { getBotContext, getLogger } from "../bindings";
-import { AssignEvent, Comment, IssueType, Payload } from "../types";
+import { AssignEvent, Comment, IssueType, Payload, UserType } from "../types";
 import { checkRateLimitGit } from "../utils";
 import { DEFAULT_TIME_RANGE_FOR_MAX_ISSUE, DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED } from "../configs";
 
@@ -147,7 +147,8 @@ export const updateCommentOfIssue = async (msg: string, issue_number: number, re
 };
 
 export const upsertCommentToIssue = async (issue_number: number, comment: string, action: string, reply_to?: Comment) => {
-  if (action == "edited" && reply_to) {
+  const checkLastComment = await checkIfLastCommentIsBot(issue_number);
+  if (action == "edited" && reply_to && !checkLastComment) {
     await updateCommentOfIssue(comment, issue_number, reply_to);
   } else {
     await addCommentToIssue(comment, issue_number);
@@ -592,4 +593,11 @@ export const getAvailableOpenedPullRequests = async (username: string) => {
     }
   }
   return result;
+};
+
+export const checkIfLastCommentIsBot = async (issue_number: number) => {
+  const comments = await getAllIssueComments(issue_number);
+  const lastComment = comments.pop();
+  if(!lastComment) return false;
+  return lastComment.user.type === UserType.Bot
 };


pavlovcik:
Are you able to post proof as an example of it editing the last comment?

src/helpers/issue.ts
```diff
@@ -147,7 +147,8 @@ export const updateCommentOfIssue = async (msg: string, issue_number: number, re
 };
 
 export const upsertCommentToIssue = async (issue_number: number, comment: string, action: string, reply_to?: Comment) => {
-  if (action == "edited" && reply_to) {
+  const checkLastComment = await checkIfLastCommentIsBot(issue_number);
+  if (action == "edited" && reply_to && !checkLastComment) {
     await updateCommentOfIssue(comment, issue_number, reply_to);
```
web4er, line 152: It looks like this will break the existing workings of `updateCommentOfIssue`. This function is supposed to update a bot's response to a command irrelevant of who commented at the end of an issue. For example, you edited `/help` to `/wallet` in 3rd comment of the issue, bot is supposed to edit its response which is the 4th comment. If there are 20 total comments in the issue, you are checking 20th comment, and preventing update to it's response to 3rd comment.
This really needs a QA.

web4er (changes requested):
One possible way to generate a double comment is send two commands very fast. Here I have generated double comment https://github.com/web4erOrg/ubiquibot/issues/34

This can also be two different users sending two different commands.

src/helpers/issue.ts
byteballet: Yes, I got it. Let me change it.

web4er: I don't like the idea of delaying the bot.

byteballet: What do you suggest? @web4er

byteballet: My basic idea was to make a config variable that stores the bot's writing state. If the bot is writing now, the bot checks its writing state every second and once writing is all done, it reads the last comment and checks.

web4er: I think it should not break, or downgrade some existing functionality. That is why my approach would be to start by reproducing the issue. But let's see what @pavlovcik says

byteballet: > That is why my approach would be to start by reproducing the issue.

In my understanding, it will also downgrade the bot's functionality since we need to get all the comments. Am I wrong?

byteballet: > My basic idea was to make a config variable that stores the bot's writing state. If the bot is writing now, the bot checks its writing state every second and once writing is all done, it reads the last comment and checks.

I think it would be good to use exponential-backoff here too.
Basically we can add delayFirstAttempt and can make a small delay before the bot completes writing.
@pavlovcik rfc

pavlovcik: Cant we check the node id of the edited comment and then edit the following comment by the bot? Then the delays or order don't matter.

byteballet: > Cant we check the node id of the edited comment and then edit the following comment by the bot? Then the delays or order don't matter.

I apologize if I misunderstood you, but the problem does not occur during the editing process. The issue arises when a user quickly comments two commands (e.g. "/start" and "/stop") in succession. In this situation, the bot identifies the last comment as "/stop" both times.

To clarify, if a user comments "/stop" while the bot is in the process of adding a comment assigned to the command "/start" but the state is still pending, the bot gets the last comment as "/stop" instead of the bot's comment designated as "/start".

byteballet: @pavlovcik
Waiting for you comment.

pavlovcik: Netlify generally has a pretty quick response time of ~500ms so let's let this edge case slide because I can't think of a solution now. 

@rndquu any ideas?

byteballet: IMO, we can ignore that kind of double post.
Actually, I tried to trigger that kind of double post but I couldn't...
https://github.com/byteballet/santa-bringyouwishes/issues/121

rndquu: > Netlify generally has a pretty quick response time of ~500ms so let's let this edge case slide because I can't think of a solution now.
> 
> @rndquu any ideas?

I agree with web4er that the [root cause](https://github.com/ubiquity/ubiquibot/pull/597#issuecomment-1673631086) should be fixed. Editing comments will simply make a debug harder.

rndquu:
Closing this one because the original [issue](https://github.com/ubiquity/ubiquibot/issues/601) is no longer relevant because editing bot comments [breaks](https://github.com/ubiquity/ubiquibot/issues/601#issuecomment-1690784123) other bot's functionality

PR #383, opened 2023-06-08
feat: default config

Venoox:
Resolves https://github.com/ubiquity/ubiquibot/issues/327

@rndquu I think even relative path is not gonna work because the @vercel/ncc bundler doesn't put it in the build process.

PR #627, opened 2023-08-17
fix: derive weight automatically

### LINKED ISSUE #495: Refactor Config: Automatically Derive `weight` / ref

byteballet:
Resolves #495


pavlovcik (changes requested):
Priority weight was not implemented according to the spec:

> During runtime, when a bounty is being assistive priced by the bot, the bot should parse the number from the Priority label and use it in place of the weight

Did you accommodate the following?

> Any words should be ignored from this parser, except for Priority:  of course, in order to understand that it is a priority label.

---

- I'm reviewing the referenced comment https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228 and realized that I might not have been clear that I was only referring to `weight` in the context of the `Priority` label.

src/helpers/shared.ts
```diff
@@ -19,3 +19,15 @@ export const shouldSkip = (): { skip: boolean; reason: string } => {
 };
 
 export const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
+
+export const calculateWeight = (label: LabelItem | undefined): number => {
+  if (!label) return 0;
+  const matches = label.name.match(/\d+/);
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) : 0;
+  if (label.name.split(":")[0] === "Priority") return number + 1;
```
pavlovcik, line 27:
```suggestion
  if (label.name.split(":")[0] === "Priority") return number;
```

I think this might be correct for the spec.

src/helpers/shared.ts
```diff
@@ -19,3 +19,15 @@ export const shouldSkip = (): { skip: boolean; reason: string } => {
...
+  if (label.name.includes("Hour")) return number * 0.125;
+  if (label.name.includes("Day")) return 1 + (number - 1) * 0.25;
+  if (label.name.includes("Week")) return number + 1;
+  if (label.name.includes("Month")) return 5 + (number - 1) * 8;
```
pavlovcik, line 31: Can you explain the math formulas?

byteballet:
> Priority weight was not implemented according to the spec:
> 
> > During runtime, when a bounty is being assistive priced by the bot, the bot should parse the number from the Priority label and use it in place of the weight
> 
> Did you accommodate the following?
> 
> > Any words should be ignored from this parser, except for Priority:  of course, in order to understand that it is a priority label.
> 
> * I'm reviewing the referenced comment [feat: default config #383 (comment)](https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228) and realized that I might not have been clear that I was only referring to `weight` in the context of the `Priority` label.
> * What is the `weight` representing for the `Time` labels? Does anybody know?
> * @wannacfuture is working on calculating the `Time` values in [`value` should be dynamically parsed from the time value using the `ms` library #431](https://github.com/ubiquity/ubiquibot/issues/431)



> Priority weight was not implemented according to the spec:
> 
> > During runtime, when a bounty is being assistive priced by the bot, the bot should parse the number from the Priority label and use it in place of the weight
> 
> Did you accommodate the following?
> 
> > Any words should be ignored from this parser, except for Priority:  of course, in order to understand that it is a priority label.
> 
> * I'm reviewing the referenced comment [feat: default config #383 (comment)](https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228) and realized that I might not have been clear that I was only referring to `weight` in the context of the `Priority` label.
> * What is the `weight` representing for the `Time` labels? Does anybody know?
> * @wannacfuture is working on calculating the `Time` values in [`value` should be dynamically parsed from the time value using the `ms` library #431](https://github.com/ubiquity/ubiquibot/issues/431)

Bounty price is depend on both weight from priority and time label.
`const price = 1000 * base * timeValue * priority;`

src/helpers/shared.ts
byteballet: I made the formulas to calculate weight from time-label name.
1Hour: 0.125
1Day: 1
1Week: 2
https://github.com/ubiquity/ubiquibot/blob/59825e4d7289a8da87bb3f568d9e69b183c821cf/ubiquibot-config-default.json#L16C8-L16C8

src/helpers/shared.ts
```diff
@@ -19,3 +19,15 @@ export const shouldSkip = (): { skip: boolean; reason: string } => {
 };
 
 export const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
+
+export const calculateWeight = (label: LabelItem | undefined): number => {
+  if (!label) return 0;
+  const matches = label.name.match(/\d+/);
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) : 0;
+  if (label.name.split(":")[0] === "Priority") return number;
```
whilefoo, line 27:
```suggestion
  if (label.name.toLowerCase().includes("priority")) return number;
```
Maybe it's more flexible to check if it includes the word `priority`

whilefoo (changes requested):

whilefoo:
https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/handlers/shared/pricing.ts#L11C1-L11C1

@pavlovcik it appears we only use `weight` from priority and time labels to calculate the price. We only use the `value` to calculate end date.

src/helpers/shared.ts
byteballet: I am not sure what I have to do with time label weight...
@pavlovcik rfc.

pavlovcik:
> https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/handlers/shared/pricing.ts#L11C1-L11C1
> 
> 
> 
> @pavlovcik it appears we only use `weight` from priority and time labels to calculate the price. We only use the `value` to calculate end date.

This looks like a mistake. One is to set the price, one is to get the price. Both should be using one or the other.

byteballet:
> > https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/handlers/shared/pricing.ts#L11C1-L11C1
> > @pavlovcik it appears we only use `weight` from priority and time labels to calculate the price. We only use the `value` to calculate end date.
> 
> This looks like a mistake. One function is to set the price, the other function is to get the price. Both should be using `value` or `weight`.

But now it seems weight and value are set default...

whilefoo:
Both are using weight. The `timeValue` parameter in `calculateBountyPrice` is actually passed weight in https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/helpers/label.ts#L74

pavlovcik:
> Both are using weight

Then let's get rid of `value`

byteballet:
> 
> Then let's get rid of `value`

`

> > Both are using weight
> 
> We should rename it to `value` and then fully get rid of `weight`

For both time and priority?

byteballet:
> We should rename it to `value` and then fully get rid of `weight`

Please note that weight of both labels are used to calculate price and value in time label is used to calculate deadline.

pavlovcik:
> > We should rename it to `value` and then fully get rid of `weight`
> 
> 
> 
> Please note that `weight` of both labels are used to calculate price and `value` in time label is used to calculate deadline. We can remove both `weight` and `value` and make a simple helper function that calculates them from label `name`. @pavlovcik 

Yes that is what the specification says. The only adjustment I have is to rename the property name from `weight` to `value`

byteballet:
I removed both `value` and `weight` and made simple function that calculates them.

pavlovcik:
@whilefoo you think you can re-review?

byteballet:
> @whilefoo you think you can re-review?

Please note that I made a simple function that calculates value from time label name not using `ms` library to make clearer. It will conflict with https://github.com/ubiquity/ubiquibot/pull/638 . One of https://github.com/ubiquity/ubiquibot/pull/627 and https://github.com/ubiquity/ubiquibot/pull/638 might need modifications after another one is merged.

pavlovcik (approved):
This implementation looks pretty clean. Just needs QA.

src/helpers/shared.ts
```diff
@@ -19,3 +19,27 @@ export const shouldSkip = (): { skip: boolean; reason: string } => {
...
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) || 0 : 0;
+  if (label.name.toLowerCase().includes("priority")) return number;
+  if (label.name.toLowerCase().includes("hour")) return number * 0.125;
+  if (label.name.toLowerCase().includes("day")) return 1 + (number - 1) * 0.25;
+  if (label.name.toLowerCase().includes("week")) return number + 1;
+  if (label.name.toLowerCase().includes("month")) return 5 + (number - 1) * 8;
+  return 0;
+};
+
+export const calculateDuration = (label: LabelItem): number => {
```
whilefoo, line 35: duration calculation will be implemented in #638 so I'm not sure why did you also implement it

src/helpers/shared.ts
```diff
@@ -19,3 +19,27 @@ export const shouldSkip = (): { skip: boolean; reason: string } => {
...
+
+export const calculateWeight = (label: LabelItem | undefined): number => {
+  if (!label) return 0;
+  const matches = label.name.match(/\d+/);
+  const number = matches && matches.length > 0 ? parseInt(matches[0]) || 0 : 0;
+  if (label.name.toLowerCase().includes("priority")) return number;
+  if (label.name.toLowerCase().includes("hour")) return number * 0.125;
+  if (label.name.toLowerCase().includes("day")) return 1 + (number - 1) * 0.25;
+  if (label.name.toLowerCase().includes("week")) return number + 1;
+  if (label.name.toLowerCase().includes("month")) return 5 + (number - 1) * 8;
```
whilefoo, line 31: I don't understand this calculation, how is weight related to time?

byteballet: I calculated it to adjust the value with the defaults.
1hour: 0.125
1day: 1
1 week: 2
2 weeks: 3
3 weeks: 4
1 month: 5
and made a rule for multiplier like 0.125, 0.25, 1, 8

src/helpers/shared.ts
byteballet: Original spec is to remove both weight and value.

Draeieg:
https://github.com/ubiquibot/staging/issues/144 tags haven't been changed in accordance to the new weights which causes the (old) 0 normal tag to not price the issue @byteballet

byteballet:
> [ubiquibot/staging#144](https://github.com/ubiquibot/staging/issues/144) tags haven't been changed in accordance to the new weights which causes the (old) 0 normal tag to not price the issue @byteballet

IMO, we are going to shift the priority tag.
https://github.com/ubiquity/ubiquibot/pull/627#discussion_r1297758520

>So in this case we would shift them all up by one, meaning, Priority: 1 (Normal) would be the first value, and we can delete the weight property.

https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228

@Draeieg

Draeieg:
> > [ubiquibot/staging#144](https://github.com/ubiquibot/staging/issues/144) tags haven't been changed in accordance to the new weights which causes the (old) 0 normal tag to not price the issue @byteballet
> 
> I thought, we are going to shift the priority tag. [#627 (comment)](https://github.com/ubiquity/ubiquibot/pull/627#discussion_r1297758520)
> 
> > So in this case we would shift them all up by one, meaning, Priority: 1 (Normal) would be the first value, and we can delete the weight property.
> 
> [#383 (comment)](https://github.com/ubiquity/ubiquibot/pull/383#discussion_r1254078228)
> 
> @Draeieg

@pavlovcik if this is OK then QA is approved

pavlovcik:
> [ubiquibot/staging#144](https://github.com/ubiquibot/staging/issues/144) tags haven't been changed in accordance to the new weights which causes the (old) 0 normal tag to not price the issue @byteballet

Yeah we can update the priority numbers.

Draeieg (approved):

whilefoo (approved):

PR #638, opened 2023-08-18
feat: extract duration dynamically from time label name

### LINKED ISSUE #431: `value` should be dynamically parsed from the time value using the `ms` library

wannacfuture:
Resolves #431 

QA: https://github.com/wannacfuture/Battleship/issues/85

Used `timestring` library due to `ms` doesn't support months and years

### DIFF
--- a/.github/ubiquibot-config.yml
+++ b/.github/ubiquibot-config.yml
@@ -3,19 +3,14 @@ price-multiplier: 1.5
 time-labels:
   - name: "Time: <1 Hour"
     weight: 0.125
-    value: 3600
   - name: "Time: <2 Hours"
     weight: 0.25
-    value: 7200
   - name: "Time: <4 Hours"
     weight: 0.5
-    value: 14400
   - name: "Time: <1 Day"
     weight: 1
-    value: 86400
   - name: "Time: <1 Week"
     weight: 2
-    value: 604800
 priority-labels:
   - name: "Priority: 0 (Normal)"
     weight: 1
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@supabase/supabase-js": "^2.4.0",
     "@types/mdast": "^3.0.11",
     "@types/ms": "^0.7.31",
+    "@types/timestring": "^6.0.2",
     "@typescript-eslint/eslint-plugin": "^5.59.11",
     "@typescript-eslint/parser": "^5.59.11",
     "@uniswap/permit2-sdk": "^1.2.0",
@@ -61,6 +62,7 @@
     "prettier": "^2.7.1",
     "probot": "^12.2.4",
     "telegraf": "^4.11.2",
+    "timestring": "^7.0.0",
     "tsx": "^3.12.7",
     "yaml": "^2.2.2"
   },
--- a/src/handlers/assign/action.ts
+++ b/src/handlers/assign/action.ts
@@ -1,3 +1,5 @@
+import timestring from "timestring";
+
 import { getBotConfig, getBotContext, getLogger } from "../../bindings";
 import { addCommentToIssue, closePullRequest, getOpenedPullRequestsForAnIssue } from "../../helpers";
 import { Payload, LabelItem } from "../../types";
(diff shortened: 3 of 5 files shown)


web4er:
I see this PR and another PR https://github.com/ubiquity/ubiquibot/pull/627 both changing this
```typescript
const LabelItemSchema = Type.Object({
  name: Type.String(),
  weight: Type.Number(),
  value: Type.Optional(Type.Number()),
});
```
What I understand is that the configs are being changed to make it easier to specify priority weight, and time value in a single line but the concept of priority weight and time value both still exist. The problem with removing them here is, now wherever time label is going `timestring()` is going with it, and wherever the priority label is going `calculateWeight()` is going with it. 

I would suggest keeping the above schema and this function which appears to be a good place to transform and fill priorityLabels, and time values all at once, and you probably will need to use `timestring()`  and `calculateWeight()` only once.
https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/bindings/config.ts#L10

whilefoo:
I think you missed other parts of code that still use the config time tabels to calculate pricing... like in https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/handlers/shared/pricing.ts#L11

Example of price label not working: https://github.com/ubiquity-whilefoo/ubiquibot/issues/6

whilefoo:
> I would suggest keeping the above schema and this function which appears to be a good place to transform and fill priorityLabels, and time values all at once, and you probably will need to use `timestring()` and `calculateWeight()` only once.

Are you suggesting that when `loadConfig` is called, the bot checks for all labels defined in the repo and fills the labels config, and then it can be normally used without other changes in the code?

web4er:
> Are you suggesting that when loadConfig is called, the bot checks for all labels defined in the repo and fills the labels config

I think the bot is doing that already. Just need to transform the labels.

>  and then it can be normally used without other changes in the code?

I think so. One case that comes to mind is when some labels are loaded directly by some function and the function needs its weight/value and wants to call transform functions directly, but that can also be overcome by comparing it with existing labels from config instead of calling transformer.

whilefoo:
> I think the bot is doing that already. Just need to transform the labels.

I mean now it's reading them from the config, but we would need to read them with [Github API](https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#list-labels-for-a-repository)

wannacfuture:
> I think you missed other parts of code that still use the config time tabels to calculate pricing... like in

I guess we can only set the time labels which are in the config.
at least that's what needed to be done for this issue.

whilefoo:
I thought this issue about removing hardcoded time labels from the config and parsing them dynamically without the config. @pavlovcik can you clarify?

pavlovcik:
> I thought this issue about removing hardcoded time labels from the config and parsing them dynamically without the config. @pavlovcik can you clarify?

That sounds nice but technically the task is only for removing the value labels. You can see the original linked context from the pull request review months ago in the spec.

whilefoo:
Then I suggest we use @web4er idea to we remove `value` from `ubiquibot-config.yaml` but keep it in the internal config and parse it once in `loadConfig` so it can be used anywhere in the project without parsing it every time.

Draeieg:
is this ready to QA?

pavlovcik:
Well it looks like the consensus between at least 0xcodercrane and I is that the e should be using vercel's ms library because it's more used and from a top notch team. It's a higher quality dependency.

wannacfuture:
> Well it looks like the consensus between at least 0xcodercrane and I is that the we should be using vercel's ms library because it's more used and from a top notch team. It's a higher quality dependency.

Okay. Let me use ms.

0xcodercrane:
any updates here? @wannacfuture

wannacfuture:
let me handle both this and https://github.com/ubiquity/ubiquibot/issues/685

wannacfuture:
tbh, I also agree with it - 

https://github.com/ubiquity/ubiquibot/pull/638#issuecomment-1687889371

We will just need to parse the time label name once in loadConfig().

@0xcodercrane , @pavlovcik

wannacfuture:
@rndquu This pr and both linked issue can be closed. Seems it is already fixed.

PR #639, opened 2023-08-19
feat: read from env

### LINKED ISSUE #619: Read `DEFAULT_TIME_RANGE_FOR_MAX_ISSUE` from `env`
Original [discussion](https://github.com/ubiquibot/ubiquibot-config/pull/2#discussion_r1293969395)

For QA purposes we need to read [DEFAULT_TIME_RANGE_FOR_MAX_ISSUE](https://github.com/ubiquity/ubiquibot/blob/d56a523c81829e17513f7da1f1ccb1bbdca8b180/src/configs/shared.ts#L6) and [DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED](https://github.com/ubiquity/ubiquibot/blob/d56a523c81829e17513f7da1f1ccb1bbdca8b180/src/configs/shared.ts#L7C14-L7C54) from env variables so that @Draeieg could set them in the netlify dashboard for testing purposes

What should be done:
1. Read `DEFAULT_TIME_RANGE_FOR_MAX_ISSUE` and `DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED` from `process.env`
2. By default set `DEFAULT_TIME_RANGE_FOR_MAX_ISSUE` to `24` (if env variable is empty)
3. By default set `DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED` to `true` (if env variable is empty)

EtherealGlow:
Resolves #619


whilefoo (changes requested):
Because we are reading from env I think it's better to put these variables in the [loadConfig](https://github.com/ubiquity/ubiquibot/blob/f6e99dfb406559a044969bb82aff478daae172cd/src/bindings/config.ts#L75C6-L75C6) but leave the default values in this file

EtherealGlow:
Doesn't load config function return object of type bot config or am I getting confused

whilefoo:
yes but this is internal bot config, not like ubiquibot-config.yaml

EtherealGlow:
i didnt exactly understand the spec

src/bindings/config.ts
```diff
@@ -84,6 +91,14 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     wallet: {
       registerWalletWithVerification: registerWalletWithVerification,
     },
+    time: {
+      DEFAULT_TIME_RANGE_FOR_MAX_ISSUE: process.env.DEFAULT_TIME_RANGE_FOR_MAX_ISSUE
```
whilefoo, line 95: the variable names should be in snakeCase and I think it's better to put them in unassign object because it's related to unassigning

src/helpers/issue.ts
```diff
@@ -631,8 +632,10 @@ export const getCommitsOnPullRequest = async (pullNumber: number) => {
 };
 
 export const getAvailableOpenedPullRequests = async (username: string) => {
-  if (!DEFAULT_TIME_RANGE_FOR_MAX_ISSUE_ENABLED) return [];
   const context = getBotContext();
+  const botConfig = await loadConfig(context);
```
whilefoo, line 636: this is not how you load the config, you should use `getBotConfig`

whilefoo (changes requested):

src/types/config.ts
```diff
@@ -47,6 +47,8 @@ export const PayoutConfigSchema = Type.Object({
 export const UnassignConfigSchema = Type.Object({
   followUpTime: Type.Number(),
   disqualifyTime: Type.Number(),
+  defaultTimeRangeForMaxIssue: Type.Number(),
+  defaultTimeRangeForMaxIssueEnabled: Type.Boolean(),
```
whilefoo, line 51: why is it called default?

whilefoo (changes requested):

src/types/config.ts
whilefoo, line 51:
```suggestion
  timeRangeForMaxIssueEnabled: Type.Boolean(),
```

src/types/config.ts
```diff
@@ -47,6 +47,8 @@ export const PayoutConfigSchema = Type.Object({
 export const UnassignConfigSchema = Type.Object({
   followUpTime: Type.Number(),
   disqualifyTime: Type.Number(),
+  defaultTimeRangeForMaxIssue: Type.Number(),
```
whilefoo, line 50:
```suggestion
  timeRangeForMaxIssue: Type.Number(),
```

whilefoo (changes requested):

whilefoo (approved):

PR #640, opened 2023-08-20
E2E tests

### LINKED ISSUE #633: Automated QA: End-to-end tests that run all commands and check their responses + `yarn test`

whilefoo:
Resolves #633 

Example of E2E test: https://github.com/ubiquibot-whilefoo-testing/ubiquibot/issues/24


gitcoindev:
This is cool. Not mocked but connecting to an actual staging github application using a personal access token and verifying the results. At some point would be useful to add coverage output.

tests/commands.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    await waitForNWebhooks(2);
+
+    await checkLastComment(issue.number, `Successfully changed the payout multiplier for @${username} to 1. The reason is not provided.`);
+
+    await createComment(issue.number, `/multiplier @${username} 2`);
+    await waitForNWebhooks(2);
+
+    await checkLastComment(
+      issue.number,
+      `Successfully changed the payout multiplier for @${username} to 2. The reason is not provided. This feature is designed to limit the contributor's compensation for any bounty on the current repository due to other compensation structures (i.e. salary.) are you sure you want to use a bounty multiplier above 1?`
```
pavlovcik, line 115: Do you think it's feasible that we export and use the exact same strings from the code base?

whilefoo: That would be ideal, but I'm not quite sure how to do it elegantly, because we need to replace username in the string, so we would need to do something like `Successfully changed the payout multiplier for @%username%` and then use `.replace("%username%, username)`

cuddle-dev (changes requested):
I can see @whilefoo you did a lot of works to setup unit tests + e2e. 

but I feel like its a little bit away from the original goal.

In terms of unit tests, we should have coverage % in branch and function.
In terms of e2e, need to create a scenario and verify each step -- those need to be in workflow.

src/handlers/payout/post.ts
```diff
@@ -55,6 +56,20 @@ export const incentivizeComments = async () => {
...
+      const res = botCommentBody.match(pattern);
+      if (res) {
+        if (res[1] === "false") {
+          return;
+        }
+        break;
+      }
+    }
+  }
+
```
rndquu, line 72: What is this whole loop used for?

0xcodercrane:
I am gonna contribute to this task as an codeowner because its an very important task for now and future. 
No split in payment even if I contribute. The whole amount would get paid to @whilefoo .

To contribute the task(To be honest, I wanna say this is far away from the completion),

- merge @whilefoo 's branch to `feat/e2e`
- I will push my work on `feat/e2e`. @whilefoo you can continue working on `feat/e2e` branch cause you're already a member. right?

What do you think admins?

rndquu (dismissed):
Just to clear the terminology, by "e2e testing" we mean running tests against a staging organization in a real github environment.

The original issue states `the test script should file a new issue on a locally configurable repository (.env) that tests all the bot commands`. Covering all cases is a huge task so this PR tries to cover at least some cases. 

I actually have hard time running the tests getting `thrown: "Exceeded timeout of 10000 ms for a hook` error all the time:
```
  commmands test
    ✕ /wallet correct address (5 ms)
    ✕ /wallet wrong address
    ✕ /multiplier
    ✕ /query
    ✕ /query wrong username
    ✕ /help
    ✕ /allow
    ✕ /start and /stop

  ● commmands test › /wallet correct address

    thrown: "Exceeded timeout of 10000 ms for a hook.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      150 |   let issue: Issue;
      151 |
    > 152 |   beforeAll(async () => {
          |   ^
      153 |     const res = await octokitAdmin.rest.issues.create({
      154 |       repo,
      155 |       owner,

      at beforeAll (../tests/commands.test.ts:152:3)
      at Object.<anonymous> (../tests/commands.test.ts:149:9)
```

Nevertheless I think that the current PR is good to merge and we should run e2e tests introduced in the current PR in a github CI ASAP because every new feature makes the e2e tests implemented in the current PR a bit obsolete.

I would:
1. Merge the current PR
2. Add a new CI workflow that runs e2e tests in a staging org on every push to a newly created PR by a bounty hunter (+fix the timeout errors)
3. Create a new issue to refactor e2e tests step by step to cover all basic cases (i.e. commands)
4. Start implementing unit tests (which is a totally different epic)

P.S. E2E tests on a staging org are gonna be really brittle but it is better than nothing

0xcodercrane:
If the task is a minimal goal of overall E2E, we can merge the PR. but we should add `yarn test` to workflwo file on the current PR.
So we can quickly review the hunters don't break the commands in review step.

0xcodercrane:
> So we keep the current e2e tests and add a new "test workflow" as a part of the current issue, right?

Right. In that way, we can make the PR valuable after the merge.

src/handlers/payout/post.ts
whilefoo: it's checking if autopay was disabled

whilefoo:
I got rate limited one time during development, so I'm worried if we enable this for every commit, we will get rate limited quickly.

pavlovcik:
> I got rate limited one time during development, so I'm worried if we enable this for every commit, we will get rate limited quickly.

Different GitHub CI events to consider, including only running on merge to `development` if rate limiting is a concern.

rndquu:
> I got rate limited one time during development, so I'm worried if we enable this for every commit, we will get rate limited quickly.

Yes, perhaps we could run e2e tests only on push to the `development` branch

0xcodercrane:
> > I got rate limited one time during development, so I'm worried if we enable this for every commit, we will get rate limited quickly.
> 
> Yes, perhaps we could run e2e tests only on push to the `development` branch

The problem I am seeing if we enable e2e on push event in the `development` branch is that we (as a reviewer) couldn't verify the hunters don't break before merging the PR right?

Even if we have rate limit, we should allow e2e tests on every pull requests to `development` branch.

rndquu:
> The problem I am seeing if we enable e2e on push event in the development branch is that we (as a reviewer) couldn't verify the hunters don't break before merging the PR right?

Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.

pavlovcik:
> > The problem I am seeing if we enable e2e on push event in the development branch is that we (as a reviewer) couldn't verify the hunters don't break before merging the PR right?
> 
> Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.

Great idea!

package.json
```diff
@@ -56,6 +56,7 @@
     "node-html-to-image": "^3.3.0",
     "nodemon": "^2.0.19",
     "octokit": "^3.1.0",
+    "parse5": "^7.1.2",
```
pavlovcik, line 59: Is this necessary?

pavlovcik:
> > > The problem I am seeing if we enable e2e on push event in the development branch is that we (as a reviewer) couldn't verify the hunters don't break before merging the PR right?
> > 
> > 
> > Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.
> 
> Great idea!



> I actually have hard time running the tests getting `thrown: "Exceeded timeout of 10000 ms for a hook` error all the time:

We have six hours to run tests. I don't see how timeout is an issue. Just increase the time?

package.json
whilefoo: you mean one line above `octokit`? I think technically it's not needed because it's imported by probot but usually it's not a good idea to import indirect dependencies

whilefoo:
I will fix the tests because there were some changes in between, make timeout 6 hours and add workflow

whilefoo:
> Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.

It seems you can only manually run workflows on a branch and not PR

rndquu:
> > > The problem I am seeing if we enable e2e on push event in the development branch is that we (as a reviewer) couldn't verify the hunters don't break before merging the PR right?
> > 
> > 
> > Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.
> 
> Great idea! Any scope creep (this) can be spun off into a separate task though, so that we can merge asap?

As far as I understand the [plan](https://github.com/ubiquity/ubiquibot/pull/640#issuecomment-1708517166) is to fix current e2e tests and add a workflow that runs them as a part of the current PR

rndquu:
> > Yes. Perhaps we could add the `workflow_dispatch` handler for the e2e workflow and make it a part of the QA process so that QA engineer could run e2e tests manually before approving.
> 
> It seems you can only manually run workflows on a branch and not PR

Let's keep it simple, I think it's enough (for start) to have an e2e workflow with the `workflow_dispatch` handler. At least we would be able to fetch bounty hunter's PRs and run e2e tests against our own test orgs. We can modify the e2e workflow later to run on forked repos.

package.json
pavlovcik: I don't see parse5 being used in your code.

whilefoo: If I go to `Files changed` tab, I don't see parse5 as being added, only octokit 
image

pavlovcik: Now me too. Something is up with your git because last time you somehow nuked my work on the config without displaying the deletion on your pull request.

pavlovcik:
Hey @whilefoo its been over a month. Why don't we leave it with a simple implementation so that we can merge asap.

whilefoo:
Successful workflow run: https://github.com/ubiquity-whilefoo/ubiquibot/actions/runs/6290628334

@0xcodercrane if you don't have any comments, let's merge this

0x4007:
Next we should definitely test every incentive cause those are the most important features.

0x4007:
> Successful workflow run: https://github.com/ubiquity-whilefoo/ubiquibot/actions/runs/6290628334
> 
> @0xcodercrane if you don't have any comments, let's merge this

Why do some take so long to run? Like start and stop taking 98 seconds?

.github/workflows/e2e-test.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Build
+        run: yarn build
+
+      - name: Test
+        env:
+          LOG_ENVIRONMENT: "production"
+          SUPABASE_URL: ${{ secrets.SUPABASE_URL }}
+          SUPABASE_KEY: ${{ secrets.SUPABASE_KEY }}
+          X25519_PRIVATE_KEY: "QCDb30UHUkwJAGhLWC-R2N0PiEbd4vQY6qH2Wloybyo"
```
0x4007, line 34: Should this be exposed? Not sure exactly what it is used for.

whilefoo: it's okay, it must be the same because the test's config wallet private key in encrypted with this private key

whilefoo:
> Why do some take so long to run? Like start and stop taking 98 seconds?

It's a long test. Also there is some delay between because Github sometimes doesn't include latest comments right away so sometimes even with the delay it fails

PR #649, opened 2023-08-20
Realtime logger app

### LINKED ISSUE #593: We should have a simple webapp to stream realtime logs from the database
- We must read the logs from the Supabase and display them on a simple and public web app. 
  - Create a directory for the logs web app, and deploy to Cloudflare Pages via Github Actions.
- Ideally, the logs should stream realtime with a web socket.

devpanther:
Resolves #593


devpanther:
Supabase realtime-js library not working at this point, apparently it needs a native Nodejs compatible platform to run..

So I am going to copy the websocket script and make a custom class for the websocket, mimic the realtime-js messages to the was server and it should work

pavlovcik:
> apparently it needs a native Nodejs compatible platform to run..

What does that mean? Netlify and the other major serverless providers run Node.

devpanther:
Screenshot 2023-08-22 at 7 49 08 AM

`.gitignore` should be allowed and `_worker` is the only way for cloudflare to understand worker files.. so they both should be allowed past this check @0xcodercrane @pavlovcik

0xcodercrane:
The overall architecture doesn't make sense to me. you wanted to create another typescript project inside ubiquibot. 

I recommend to setup a new repo separate from ubiquibot if it should be another nodejs app as you say.
If it can be a part of serverless function, it should be one of the adapters.

devpanther:
I think it makes sense that way as opposed to creating a new repo and I've fixed the issue with node compatibility by creating a custom websocket connection.

What I'm having issues with now is the failing CI, the namings it's rejecting are important for this to work

whilefoo:
> Screenshot 2023-08-22 at 7 49 08 AM
> `.gitignore` should be allowed and `_worker` is the only way for cloudflare to understand worker files.. so they both should be allowed past this check @0xcodercrane @pavlovcik

https://github.com/ubiquity/ubiquibot/blob/fee0dd90ad0bc19fcbe31643e5b150fddd81c764/.github/workflows/kebab-case.yml#L15

You can add `log-app` folder to ignoreList in the workflow

devpanther:
> I'm very skeptical of this. You can always downgrade to the last stable version if the latest is "Not working"

It is not a latest or older version problem, it just does not work on platforms that are not native node compatible..

You can try the out previous commits I struggled to fix, and we are using Cloudflare pages not Netlify

devpanther:
The custom websocket it an exact replica with Websocker implementation that works for the Cloudflare pages

whilefoo:
Currently you're using Cloudflare workers to forward Websocket messages from Supabase to client, but that is unnecessary because we can use the anonymous token. Supabase JS library is isomorphic so it can be used in NodeJS or in browser, so we don't need Cloudflare workers and everything can be done on the client-side. We have row-level security enabled so you just need to create a policy and then you can use the anonymous/public JWT token and listen to changes. Check the [guide](https://supabase.com/docs/guides/realtime/postgres-changes).

devpanther:
> Currently you're using Cloudflare workers to forward Websocket messages from Supabase to client, but that is unnecessary because we can use the anonymous token. Supabase JS library is isomorphic so it can be used in NodeJS or in browser, so we don't need Cloudflare workers and everything can be done on the client-side. We have row-level security enabled so you just need to create a policy and then you can use the anonymous/public JWT token and listen to changes. Check the [guide](https://supabase.com/docs/guides/realtime/postgres-changes).

I know that, Cloudflare does not share env variables with the frontend.. that is why there is a worker consuming the env and forwarding the socket messages, if we are using frontend directly. The tokens will have to be hard-coded

devpanther:
> we don't need Cloudflare workers

The `_worker` is a part of the cloudflare pages, for the sake of having environment variables.. we can have it forward, we might need an auth at some point so anyone with URL cannot view logs

whilefoo:
If we later need auth, we can just have a login system that verifies username/password pair and [signs a JWT token](https://supabase.com/docs/guides/realtime/postgres-changes#custom-tokens) that allows access to the logs table. Or even better [Supabase Auth](https://supabase.com/docs/guides/auth) which automates everything. 
I don't think it makes sense to forward Websocket messages.

devpanther:
So we just stick to hardcoding environment variables on the frontend?

whilefoo:
Yes, that's fine for now

pavlovcik:
Yes I also think it's fine to make logs public. There isn't any sensitive data that the runtime has access to, other than possibly the signing key to the permit wallet I think? @rndquu is this where the homomorphic encryption comes in, to be able to protect against the key being logged out?

whilefoo:
Technically private key can be logged if we are not careful, because it's decrypted when we [load the config](https://github.com/ubiquity/ubiquibot/blob/fee0dd90ad0bc19fcbe31643e5b150fddd81c764/src/utils/private.ts#L156). Maybe it would be better to keep it in the encrypted form and only decrypt it when we [sign permits](https://github.com/ubiquity/ubiquibot/blob/fee0dd90ad0bc19fcbe31643e5b150fddd81c764/src/helpers/permit.ts#L54C14-L54C38)?

devpanther:
Screenshot 2023-08-22 at 11 20 20 AM

Everything in the config is logged, so if the key is decrypted.. it is going to be logged too

devpanther:
> Yes, that's fine for now

💀, 2 days tryna fix the supabase and websocket issue so I can hide env variables.. i could have just used the frontend cdn

rndquu:
> Yes I also think it's fine to make logs public. There isn't any sensitive data that the runtime has access to, other than possibly the signing key to the permit wallet I think? @rndquu is this where the homomorphic encryption comes in, to be able to protect against the key being logged out?

Why do we need a separate web page with all the logs public when the logs are stored in supabase and any explicitly authorized core team member can access them?

rndquu:
> Technically wallet private key can be logged if we are not careful, because it's decrypted when we [load the config](https://github.com/ubiquity/ubiquibot/blob/fee0dd90ad0bc19fcbe31643e5b150fddd81c764/src/utils/private.ts#L156). Maybe it would be better to keep it in the encrypted form and only decrypt it when we [sign permits](https://github.com/ubiquity/ubiquibot/blob/fee0dd90ad0bc19fcbe31643e5b150fddd81c764/src/helpers/permit.ts#L54C14-L54C38)?

Yes, private key is decrypted at the beginning of github event handling so private key can easily be leaked (especially with public logs). We could decrypt it only on permit sign but in the end this is not a 100% guarantee from a PK leak.

pavlovcik:
> and any explicitly authorized core team member can access them from a supabase UI?

The original idea was for any bounty hunter to be able to help with debugging. Issues sometimes are noticed on production and I will file an issue for it. Here is a very [recent example](https://github.com/ubiquity/ubiquibot/issues/654) of this.

whilefoo:
One simple way would be to make use of Typescript classes which support private variables so it makes it an error if someone tries to access it. For example `PermitSigner` class which has a constructor that accepts the private key and has one function which signs permits. This would limit private key leakage only when loading the config.

We can achieve even better isolation by creating a separate Netlify/Cloudflare serverless function that has access to the public/private key for encrypting/decrypting Ethereum wallet private key. Ubiquibot sends encrypted wallet private key along with the permit to the function which then returns signed data. But I'm not sure if it's possible to limit who can access the function and this would make it harder to setup local development.

pavlovcik:
> creating a separate Netlify/Cloudflare serverless function that has access to the public/private key for encrypting/decrypting Ethereum wallet private key.

I think this is a good idea, perhaps as a submodule/seperate repository. Then we can limit push access on that separate repository to only administrators. That way the secure/isolated module is unlikely to be exploited by bounty hunters opening a pull request and accidentally getting it merged in.

rndquu:
> One simple way would be to make use of Typescript classes which support private variables so it makes it an error if someone tries to access it. For example `PermitSigner` class which has a constructor that accepts the private key and has one function which signs permits. This would limit private key leakage only when loading the config.
> 
> We can achieve even better isolation by creating a separate Netlify/Cloudflare serverless function that has access to the public/private key for encrypting/decrypting Ethereum wallet private key. Ubiquibot sends encrypted wallet private key along with the permit to the function which then returns signed data. But I'm not sure if it's possible to limit who can access the function and this would make it harder to setup local development.

Even if we move permit signing to a separate service there are still critical env variables that can also be accidentally exposed (`SUPABASE_KEY`, bot's app private key)

I still think it's better not to expose the logs to the public

pavlovcik:
What do you think is the simplest way for us to authorize a bounty hunter to debug production without depending on a core contributor to monitor real-time logs?

I feel like making a whole oauth flow for accessing the logs page is a bit overkill, but if we can somehow enable access based on GitHub user name that would be most convenient to delegate access.

0xcodercrane:
> I still think it's better not to expose the logs to the public

Addictions for this: I agree not to expose the production logs to the public because an unexpected security incident may happen.

> What do you all think is the simplest way for us to authorize a bounty hunter to investigate production logs without depending on a core contributor to monitor and relay the real-time logs?

Ideally, the bounty masters pick up the error message from log and a bounty hunter should be able to reproduce it once the master gives him an error message.

pavlovcik:
> without depending on a core contributor to monitor and relay the real-time logs?

> Ideally, the bounty masters should pick up the error message from log and a bounty hunter should be able to reproduce it once the master gives him an error message.

This will unnecessarily slow down development

0xcodercrane:
What you mean by UI is supabase UI or realtime-logger-app?

pavlovcik:
realtime-logger-app

pavlovcik:
I tried going to https://deploy-preview-649--ubiquibot-staging.netlify.app/log-app/index.html @0xcodercrane please confirm how to load this with netlify preview

0xcodercrane:
> I tried going to https://deploy-preview-649--ubiquibot-staging.netlify.app/log-app/index.html @0xcodercrane please confirm how to load this with netlify preview
> 
> @devpanther please let me know if you expect the path to the app to be different?

would you refresh the page? I have to change the publish directory which was incorrect.

pavlovcik:
Okay so as a new task: lets build a "login" that asks the user to pass in a supabase key, and saves to localStorage?

Core team will manually give keys out that expire in a short time e.g. one week?

devpanther:
> @devpanther please let me know if you expect the path to the app to be different?

This is meant to run on cloudflare pages, because it builds the ts files to js before uploading.. we don't have js files on the repo

devpanther:
> * I'm not sure if the current production instance supports this logging capability, but I dont see the logs being output from my comment submission now.

It uses my keys right now, we'd have to substitute that with the production URL and Anon key, you can try [my pages perview](realtime-logger-app.logger-web-apps.pages.dev) if you want to see how it works..

I'd spin up my bot and you can interact [here](https://github.com/one-two-test/one/issues/1) to see the logs in realtime

pavlovcik:
> > * I'm not sure if the current production instance supports this logging capability, but I dont see the logs being output from my comment submission now.
> 
> It uses my keys right now, we'd have to substitute that with the production URL and Anon key, you can try [my pages perview](https://realtime-logger-app.logger-web-apps.pages.dev/) if you want to see how it works..
> 
> I'd spin up my bot and you can interact [here](https://github.com/one-two-test/one/issues/1) to see the logs in realtime

This is looking pretty good thanks for sharing!

rndquu:
> > I still think it's better not to expose the logs to the public
> 
> Addictions for this: I agree not to expose the production logs to the public because an unexpected security incident may happen.
> 
> > What do you all think is the simplest way for us to authorize a bounty hunter to investigate production logs without depending on a core contributor to monitor and relay the real-time logs?
> 
> Ideally, the bounty masters should pick up the error message from log and a bounty hunter should be able to reproduce it once the master gives him an error message.
> 
> @pavlovcik @rndquu

We could also implement an encryption similar to how github encrypts organization and repo secrets. So all env variables should be encrypted by default. On bot start we load encrypted env variables and decrypt them right before the actual usage. The same rule applies for partners' wallets private keys as @whilefoo already [mentioned](https://github.com/ubiquity/ubiquibot/pull/649#issuecomment-1687899481).

Draeieg:
so, how do I QA this lol

pavlovcik:
> > so, how do I QA this lol
> 
> What'd be best, me doing a short video or you give me a time you'd be available..
> 
> I spin up my local ubiquibot and you visit https://realtime-logger-app.logger-web-apps.pages.dev/ and interact with my github repo.. Then you can QA what you see

Both of those options don't seem great. Can you set up an instance on your own Netlify and let us QA on your fork?

devpanther:
> Both of those options don't seem great. Can you set up an instance on your own Netlify and let us QA on your fork?

An instance of the bot running on my repo?

devpanther:
> looks like my deployment is not working

Okay, I needed to change the webhook URL

@pavlovcik @Draeieg here's the [Log Viewer](https://realtime-logger-app.logger-web-apps.pages.dev/) and here's the [Test Repo](https://github.com/one-two-test/one/issues/1)

pavlovcik:
Looks good to me 

image

pavlovcik (approved):

pavlovcik:
@rndquu merge queue no longer works. I had this error on another issue, so I deleted the `preview` and `production` deploy environments under the repository settings, which didn't do anything. 

Screenshot 2023-08-25 at 18 35 20

Would you mind addressing the issue?

Specifically: 
1. determine if we benefit from using a merge queue (I originally enabled it for experimental purposes but not sure if we really leveraged its capabilities)
2. fix or remove merge queue feature

log-app/package.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "cf-pages-functions",
+  "version": "0.0.0",
+  "scripts": {
+    "build": "tsc && tsx ./scripts/esbuild.ts"
```
pavlovcik, line 5: I just realized that `tsc` is not necessary here. Just do `tsx`

pavlovcik:
> @rndquu merge queue no longer works. I had this error on another issue, so I deleted the `preview` and `production` deploy environments under the repository settings, which didn't do anything.
> 
> Screenshot 2023-08-25 at 18 35 20
> Would you mind addressing the issue?
> 
> Specifically:
> 
> 1. determine if we benefit from using a merge queue (I originally enabled it for experimental purposes but not sure if we really leveraged its capabilities)
> 2. fix or remove merge queue feature

By the way I was able to fix this by removing some more branch protection features.

rndquu:
> > @rndquu merge queue no longer works. I had this error on another issue, so I deleted the `preview` and `production` deploy environments under the repository settings, which didn't do anything.
> > Screenshot 2023-08-25 at 18 35 20
> > Would you mind addressing the issue?
> > Specifically:
> > 
> > 1. determine if we benefit from using a merge queue (I originally enabled it for experimental purposes but not sure if we really leveraged its capabilities)
> > 2. fix or remove merge queue feature
> 
> By the way I was able to fix this by removing some more branch protection features but I'm unsure if we should try and rectify merge queue or not. Seems like it could be a useful feature.

**TLDR;**

Simply check the `Require branches to be up to date before merging` option in the branch protection rules which has the same effect as enabling the "merge queue" feature.

**Full answer**  

[Here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#successful-ci) is how the merge queue feature works. Basically it (applying to the `ubiquibot` repo):
1. Creates a temporary branch from a PR
2. Merges the **latest** `development` branch in that temporary branch from step 1
3. Runs CI on all workflows that have the `merge_group` handler
4. If CI passes then the PR is merged else the PR is removed from a merge queue

So simply put the "merge queue" feature simply runs the CI again with the latest changes from the `development` branch thus allowing a bounty hunter not to merge the `development` branch manually. The thing is that if the merge queue CI fails then a bounty hunter will have to merge the `development` branch anyway (to debug why CI fails). Keeping in mind that merging the latest changes from the `development` branch can be done via github UI makes the "merge queue" feature not really useful.

Screenshot 2023-08-28 at 18 54 06
 
So in my opinion the simplest solution is to check the `Require branches to be up to date before merging` option in the branch protection rules and force bounty hunters to click the "Update branch" button in the github UI.

P.S. The `Require branches to be up to date before merging` is already checked for the https://github.com/ubiquity/ubiquibot repo

PR #656, opened 2023-08-22
Db check duplicates

### LINKED ISSUE #574: Use Github IDs instead of names in DB
Currently user, repository, organization are all stored by their name in the database. User, repo or organization can change their name and I guess this is not that common but I still think it would be better to store IDs that we get from the Github API instead of names. 

---

- [ ] Store and use user ID. 
- [ ] Store and save username. 
- [ ] Ensure that user name is updated every time user ID is accessed, in case a username is updated.

kamaalsultan:
Resolves #574


0x4007 (commented):
I just updated the specification, but I don't see a mechanism to sync the username. Do you have that here?

byteballet:
> I just updated the specification, but I don't see a mechanism to sync the username. Do you have that here?

Sorry, I thought it will be a new issue. I will update.

pavlovcik:
> > I just updated the specification, but I don't see a mechanism to sync the username. Do you have that here?
> 
> 
> 
> Sorry, I thought it will be a new issue. I will update.

That might have made sense actually. Hopefully it's fine I updated the existing one!

pavlovcik:
@rndquu rfc

I figured caching the username for reads is useful to save on network requests. It seems useful to be able to directly refer to the username from the database.

byteballet:
> I figured caching the username for reads is useful to save on unnecessary network requests to the GitHub API. Intuitively to me it seems useful to be able to directly refer to the username from the database. But at the same time, perhaps it's a waste of storage space.

Also I think there is no big difference between reading database and using github api.

whilefoo:
I took a quick look and we never get username from database. Most of the time all user data is available in the webhook payload.

package-lock.json
pavlovcik, line 1: This should not be here. We use yarn. Please remove.

src/handlers/comment/handlers/query.ts
```diff
@@ -22,9 +23,11 @@ export const query = async (body: string) => {
   const regex = /^\/query\s+@([\w-]+)\s*$/;
   const matches = body.match(regex);
   const user = matches?.[1];
+  getUser;
```
pavlovcik, line 26:
```suggestion
```

pavlovcik:
> I took a quick look and we never query username from database. Most of the time all user data is available in the webhook payload.

Looks like we should get rid of the username entirely then.

byteballet:
> Looks like we should get rid of the username entirely then.

Yes, that must be the solution.

Draeieg:
let me know when this is ready to go into QA

byteballet:
> let me know when this is ready to go into QA

I guess it's now.

rndquu:
@byteballet Could you resolve the conflicts one more time?

@whilefoo @wannacfuture Could you help with the review?

wannacfuture (commented):
@byteballet  Could you provide QA? Just wanted to make sure if its working as normal.

src/handlers/comment/handlers/query.ts
```diff
@@ -24,7 +25,8 @@ export const query = async (body: string) => {
   const user = matches?.[1];
```
whilefoo, line 25:
```suggestion
  const username = matches?.[1];
```

src/handlers/comment/handlers/query.ts
```diff
@@ -24,7 +25,8 @@ export const query = async (body: string) => {
...
 
   if (user) {
-    const walletInfo = await getWalletInfo(user, id?.toString());
+    const { id: user_id } = (await getUser(user)) ?? {};
```
whilefoo, line 28:
```suggestion
    const user = await getUser(user);
    if (!user) {
      logger.info(`Error retrieving data about user @${username}`);
      return `Error retrieving data about user @${username}`;
    }
```

src/handlers/comment/handlers/query.ts
```diff
@@ -24,7 +25,8 @@ export const query = async (body: string) => {
...
+    const walletInfo = await getWalletInfo(user, id?.toString(), user_id || 0);
```
whilefoo, line 29:
```suggestion
    const walletInfo = await getWalletInfo(user, id.toString(), user.id);
```

whilefoo (changes requested):
Code looks mostly good. Please post QA.

cuddle-dev (changes requested):

supabase/migrations/20230821160747_add_id.sql
```diff
@@ -0,0 +1,27 @@
+ALTER TABLE wallets
```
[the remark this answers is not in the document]
byteballet: Okay, I will have a look.
Thanks

PR #662, opened 2023-08-23
August 28 Update - Comment Incentives

pavlovcik:
This one has full comment incentive support and must be merged in by [28th August (Monday)](https://t.me/c/1588400061/3250/4373)


ubiquibot-config-default.json
```diff
@@ -93,37 +93,12 @@
...
-    "footnoteReference": 0,
-    "table": 0,
-    "tableCell": 0,
-    "tableRow": 0,
-    "alignType": 0,
-    "flowContent": 0,
-    "listContent": 0,
-    "phrasingContent": 0,
-    "rowContent": 0,
-    "tableContent": 0
```
pavlovcik, line 127: These should all be initialized to `0` why was this removed?

whilefoo: Because they every element is 0 by default

src/handlers/payout/post.ts
```diff
@@ -1,18 +1,21 @@
 import { getWalletAddress } from "../../adapters/supabase";
 import { getBotConfig, getBotContext, getLogger } from "../../bindings";
 import { addCommentToIssue, generatePermit2Signature, getAllIssueComments, getIssueDescription, getTokenSymbol, parseComments } from "../../helpers";
-import { MarkdownItem, Payload, UserType, CommentElementPricing } from "../../types";
+import { Incentives, Payload, StateReason, UserType } from "../../types";
+import { commentParser } from "../comment";
+import Decimal from "decimal.js";
+import { bountyInfo } from "../wildcard";
 
-const ItemsToExclude: string[] = [MarkdownItem.BlockQuote];
```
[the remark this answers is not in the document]
whilefoo: we don't use Markdown anymore, so I didn't want it to be implying it's markdown

ubiquibot-config-default.json
pavlovcik: This is the `default` configuration so I thought it makes more sense to explicitly set defaults here instead of inferring them inside of the loader?

rndquu: I agree that the default incentives config should be explicit because right now it is not clear what parameters are available

This [one](https://github.com/inikulin/parse5/blob/7f8c43c957ea60461d3afd55ca1905c095452f40/packages/parse5/lib/common/html.ts#L33) seems to be a comprehensive list of all available tags (supported by [parse5](https://parse5.js.org/index.html))

Adding all of the tags to the config is redundant but we can add at least most commonly used like `h1`, `img`, `label`, etc...

@whilefoo What do you think?

whilefoo: Ok I agree, we can have the most common tags so it will be easier for partners to understand

.github/ubiquibot-config.yml
```diff
@@ -33,6 +33,17 @@ default-labels:
...
+  comment:
+    elements:
+      code: 5
+      img: 5
+      h1: 1
+      li: 0.5
+      a: 0.5
+      blockquote: 0
+    totals:
+      word: 0.1
```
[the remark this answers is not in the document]
whilefoo: Sure, shall I create a PR?

ubiquibot-config-default.json
rndquu: > Ok I agree, we can have the most common tags so it will be easier for partners to understand. Shall I create a PR?

This would be awesome, thank you

pavlovcik: > I agree that the default incentives config should be explicit because right now it is not clear what parameters are available
> 
> 
> 
> This [one](https://github.com/inikulin/parse5/blob/7f8c43c957ea60461d3afd55ca1905c095452f40/packages/parse5/lib/common/html.ts#L33) seems to be a comprehensive list of all available tags (supported by [parse5](https://parse5.js.org/index.html))
> 
> 
> 
> Adding all of the tags to the config is redundant but we can add at least most commonly used like `h1`, `img`, `label`, etc...
> 
> 
> 
> @whilefoo What do you think?

Wait I already fixed this in my pull request. I need to find it which is tedious from my phone, but if you read this comment before I find it, please check the open pull requests from four days ago. Check my newest one that is not a draft.

src/handlers/payout/post.ts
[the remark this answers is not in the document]
0xcodercrane: https://github.com/ubiquity/ubiquibot/commit/66ff9d36fa51b270dc11eb7a0c3146c1bf2cfce8

0xcodercrane (changes requested):

src/handlers/payout/post.ts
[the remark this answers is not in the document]
whilefoo: Actually we can use [tag enum](https://github.com/inikulin/parse5/blob/7f8c43c957ea60461d3afd55ca1905c095452f40/packages/parse5/lib/common/html.ts#L33) directly from parse5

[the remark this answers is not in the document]
pavlovcik: Most of those tags are not supported by GitHub though, I would prefer to do only what is compatible with GitHub comments.

0xcodercrane (approved):

PR #664, opened 2023-08-23
fix: no double posts

### LINKED ISSUE #601: Edit Instead of Double Post

Keyrxng:
added check to edit previous post vs post new one.

Resolves #601


rndquu:
Closing this one because the original [issue](https://github.com/ubiquity/ubiquibot/issues/601) is no longer relevant because editing bot comments [breaks](https://github.com/ubiquity/ubiquibot/issues/601#issuecomment-1690784123) other bot's functionality

PR #668, opened 2023-08-25
Default config and config change

whilefoo:
Related to:
https://github.com/ubiquity/ubiquibot/pull/662#discussion_r1303166625
https://github.com/ubiquity/ubiquibot/pull/662#discussion_r1303198292


0x4007 (changes requested):
There's way more elements supported. Did you check my draft? 

I also added a link to the readme in the last section related to the default config. 

https://github.com/syntax-tree/mdast#nodes

whilefoo:
I looked at [Markdown-to-HTML](https://github.com/syntax-tree/mdast-util-to-hast) converter and added all tags that are supported.

0x4007 (changes requested):
This still seems like way too little. I'm not sure what happened to my changes I made to `ubiquibot-config-default.json` but perhaps I never merged them in.

https://github.com/pavlovcik/ubiquibot/blob/refactor/general/ubiquibot-config-default.json#L96-L128

---

You can also see some comments I left when I briefly had it as a `.jsonc` file.

https://github.com/pavlovcik/ubiquibot/blob/691e79df852d718001b58993791eb9aaa5b07223/ubiquibot-config-default.jsonc#L92-L126

--- 

According to my vscode: you clobbered my full list addition in https://github.com/ubiquity/ubiquibot/blob/a67607009a2d50fc0d1282ec289478db191ad053/ubiquibot-config-default.json

Last existed in https://github.com/ubiquity/ubiquibot/blob/fc0516e248cbaee4fc3b7c871a4e652d5e836914/ubiquibot-config-default.json

No idea why it doesnt show you deleting it in your commit. Seems that it magically disappeared. Even when directly comparing the history on github https://github.com/ubiquity/ubiquibot/commits/a67607009a2d50fc0d1282ec289478db191ad053/ubiquibot-config-default.json

Comparing bases (`...`) its not there https://github.com/ubiquity/ubiquibot/compare/61b4dd09b0a91b6d4d763a080b3a616e4a3e5bf4...7c6f399b039a874e93ed99c56925d3f7d998de75
[Three-dot Git diff comparison](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#three-dot-git-diff-comparison)
The three-dot comparison shows the difference between the latest common commit of both branches (merge base) and the most recent version of the topic branch.

Comparing commits directly (`..`) you can see the removal 
https://github.com/ubiquity/ubiquibot/compare/61b4dd09b0a91b6d4d763a080b3a616e4a3e5bf4..7c6f399b039a874e93ed99c56925d3f7d998de75
[Two-dot Git diff comparison](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-comparing-branches-in-pull-requests#two-dot-git-diff-comparison)
The two-dot comparison shows the difference between the latest state of the base branch (for example, main) and the most recent version of the topic branch.

I feel like this is a huge problem if code disappears without a trace.

rndquu (dismissed):
So the current PR adds html config elements for the comment incentives feature.

When this [PR](https://github.com/ubiquity/ubiquibot/pull/688) is merged then comment incentives reward calculation is going to work this way (@whilefoo pls correct me if I'm wrong):
1. The bot gets and parses comments in `md` format
2. For each `md` comment node (i.e. `Heading` or `ListItem`) the bot finds the corresponding html config [here](https://github.com/ubiquity/ubiquibot/blob/9d0e958c9fd7e5bde6418090b2bc82b42e6285e7/ubiquibot-config-default.json#L98) and applies price accordingly

Overall most of the `md` elements are supported (i.e. have their corresponding html tags) which is enough for tweaking comment incentive rewards

whilefoo:
Yes that's correct

pavlovcik:
@whilefoo can you answer if you included everything? It seems like there could be many more tags supported.

wannacfuture:
lets add `pre` and `ol` @whilefoo

whilefoo:
now it should support all tags

ubiquibot-config-default.json
```diff
@@ -81,7 +81,31 @@
...
+        "tr": 0,
+        "br": 0,
+        "blockquote": 0,
+        "em": 0,
+        "strong": 0,
+        "hr": 0,
+        "del": 0,
+        "pre": 0,
+        "ol": 0
+      },
```
pavlovcik, line 108:
```suggestion
        "ol": 0
        "b": 0
        "i": 0
      },
```

I'm unsure if these are valid on GitHub but the intent here is targeting bold and italics.

pavlovcik: Looks like its actually `em` on GitHub?

0x4007 (approved):

whilefoo:
will be fixed in https://github.com/ubiquity/ubiquibot/pull/818

PR #670, opened 2023-08-25
fix: stale bounty

### LINKED ISSUE #669: Consolation Prize?
# First, why?

Say someone assigns a job to themselves and carries out all of or some of the work, let's say it's an old issue and it's no longer relevant or required and just hasn't been removed/updated yet. With a very busy team and a lot of distance to cover it's possible for these scenarios to arise especially (as the codebase repeats),  as things continue to grow. 

# What happens then?

I propose a consolation prize of sorts, what exactly that would be is definitely beyond my current understanding of things and above my paygrade to decide as well obviously. 

The assignee who committed the work but through no fault of their own has had to stop due to the issue being body-bagged, should receive either a percentage of the allocated rewards or have some form of tracking in place so that if it happens on multiple occasions it can be said 'well that's n hours of work' or 'n% of the spec achieved' so you can claim 20 bucks or something along those lines.

# An alternative Solution

Sign-post any new ```/start``` command with a disclaimer not to begin the work until someone has confirmed that it is still relevant and should be executed. (If this already exists and I missed it then I think it should be made clearer on DevPool onboarding docs etc)

# Disclaimer

I am by no means salty or have a bitter taste in my mouth that this happened to myself, I'm glad it did actually as it has given me my first issue to raise. Looking forward to hearing your thoughts.

Keyrxng:
ready to start message in ```/start``` response.

[image](https://user-images.githubusercontent.com/106303466/263410028-d5b65e53-6738-4278-9a57-a67f5ad8a996.png)

[image](https://user-images.githubusercontent.com/106303466/263410077-40b5a48a-a015-4ff8-85a7-373761b1355d.png)

Resolves #669


src/bindings/config.ts
```diff
@@ -76,6 +76,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     command: commandSettings,
     assign: {
       bountyHunterMax: bountyHunterMax,
+      staleBountyTime: ms(process.env.STALE_BOUNTY_TIME || "30d"),
```
pavlovcik, line 79: Defaults must all be in `ubiquibot-config-default.json`

src/handlers/comment/handlers/table.ts
```diff
@@ -4,20 +4,21 @@ export const tableComment = ({
...
 }) => {
   return `
 <code>
-
   <table>
   <tr>
-    <td></td>
-    <td></td>
+  <td>Ready to begin?</td>
+  ${!isBountyStale ? `<td>You can start right away!</td>` : `<td>Please wait for confirmation first.</td>`}
```
pavlovcik, line 21:
```suggestion
  ${!isBountyStale ? `<td>You can start right away!</td>` : `<td>This task was created over ${days} ago. Please verify that it is still current.</td>`}
```

pavlovcik (changes requested):

src/configs/shared.ts
```diff
@@ -22,6 +22,7 @@ export const ASSIGN_COMMAND_ENABLED = true;
  */
 export const DEFAULT_FOLLOWUP_TIME = "4 days"; // 4 days
 export const DEFAULT_DISQUALIFY_TIME = "7 days"; // 7 days
+export const STALE_BOUNTY_TIME = "30d"; // 2 minutes
```
pavlovcik, line 25: 2 minutes seems short!

ubiquibot-config-default.json
```diff
@@ -62,35 +62,40 @@
...
     },
     {
       "name": "autopay",
-      "enabled": false
+      "enabled": true
+    },
+    {
+      "name": "assign",
+      "enabled": true,
+      "staleBountyTime": 120000
```
pavlovcik, line 98: I think we should have it on the root level of the `ubiquibot-config-default.json` object instead.

src/handlers/comment/handlers/table.ts
pavlovcik: "Please verify that it is still current before starting work!" 

Sorry, I reread my original edit and realized its a bit vague.

ubiquibot-config-default.json
```diff
@@ -8,6 +8,7 @@
   "disable-analytics": false,
   "comment-incentives": false,
   "register-wallet-with-verification": false,
+  "stale-bounty-time": 42060,
```
pavlovcik, line 11: Perhaps we should consider changing this file to `ubiquibot-config-default.ts` because I would love to be able to include comments. We cant rely on `fs.readFile` due to limitations with Netlify, so we can not do `.jsonc` (I already tried on a pull request earlier this month.

Keyrxng: I'll leave this for now until it's decided then

Keyrxng: @rndquu what are your thoughts for this, can we close the issue and decide this later or just leave it open for now?

pavlovcik:
Reviewing with fresh eyes:

Hide the "ready to start" row under normal circumstances, like when the bounty has been created within that same month. It's unnecessary. 

Change to "ready to start" to "Notice" so that it looks more like a warning. 

--- 

I delegate my review for the implementation of the config property names etc

src/utils/private.ts
```diff
@@ -184,6 +185,7 @@ export const getWideConfig = async (context: Context) => {
     defaultLabels: getDefaultLabels(configs),
     promotionComment: getPromotionComment(configs),
     registerWalletWithVerification: getRegisterWalletWithVerification(configs),
+    staleBountyTime: parsedDefault["stale-bounty-time"],
```
rndquu, line 188: Pls check how the bot's config params are extracted [here](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/utils/helpers.ts). There is a hierarchy of configs where the latter rewrites the former one:
1. Ubiquibot [default config](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json)
2. Partner organization wide config ([example](https://github.com/rndquu-org/ubiquibot-config/blob/main/.github/ubiquibot-config.yml))
3. Repository wide config ([example](https://github.com/ubiquity/ubiquibot/blob/development/.github/ubiquibot-config.yml))

Pls refactor the `stale-bounty-time` config extraction [accordingly](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/utils/helpers.ts)

src/handlers/comment/handlers/table.ts
```diff
@@ -4,21 +4,27 @@ export const tableComment = ({
...
   bounty,
+  isBountyStale,
+  days,
 }: {
   deadline: string;
   wallet: string;
   multiplier?: string;
   reason?: string;
   bounty?: string;
+  isBountyStale?: boolean | null;
```
rndquu, line 15: Why do we need 3 states for the `isBountyStale` variable (`true | false | null`)? Can't simply use the `boolean` type?

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,17 @@ export const assign = async (body: string) => {
     await addAssignees(issue.number, [payload.sender.login]);
   }
 
+  const days = Math.floor((new Date().getTime() - new Date(issue.created_at).getTime()) / (1000 * 60 * 60 * 24));
+  const isBountyStale = staleBounty == 0 ? null : staleBounty > days ? false : true;
```
rndquu, line 116: `staleBounty > days`

[Here](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/bindings/config.ts#L80) `ms(staleBountyTime)` returns milliseconds so it seems that we are comparing amount of milliseconds with the amount of days

rndquu (changes requested):

src/handlers/comment/handlers/table.ts
rndquu: Ok, so there are 3 states:
1. `stale-bounty-time = 0` => nothing is [rendered](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/handlers/comment/handlers/table.ts#L23)
2. `stale-bounty-time = 30d` **and** threshold reached => "warning" row is [rendered](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/handlers/comment/handlers/table.ts#L25)
3. `stale-bounty-time = 30d` **and** threshold is **not** reached => nothing is [rendered](https://github.com/ubiquity/ubiquibot/blob/7fd7d3b83aaab0b19673d9c601937d9fdeb3f18c/src/handlers/comment/handlers/table.ts#L26)

rndquu (dismissed):
Works fine

src/bindings/config.ts
```diff
@@ -90,6 +91,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     command: commandSettings,
     assign: {
       bountyHunterMax: bountyHunterMax,
+      staleBountyTime: staleBountyTime ? String(ms(staleBountyTime)) : "0",
```
whilefoo, line 94: why store it as string if it'll be converted to number later?
`"0"` will never be used because the default config has `30d` so if we want to disable the feature by default than default config should have `0d`

src/handlers/comment/handlers/table.ts
```diff
@@ -4,21 +4,27 @@ export const tableComment = ({
...
 <code>
-
   <table>
-  <tr>
-    <td></td>
-    <td></td>
-  </tr>
+  ${
+    isBountyStale == null
+      ? null
```
whilefoo, line 23: `null` will be outputted to the string
why not just use `isBountyStale ? "warning..." : ""`

whilefoo (changes requested):

src/bindings/config.ts
Keyrxng: Storing it as a string because, ms takes a string input but we need the number later to work with, as default json will have "14d"/"30d" as opposed to a number, I felt it easier to store it as strings and just convert to a number when we need to work with it.

My mistake I read pav's comment as if the default would be that it is active and they'd opt out, but you are right inactive default should be "0d" as opposed to "0".

So should this feature be inactive by default and they'd opt in with their own time range then yeah?

src/handlers/comment/handlers/table.ts
Keyrxng: My thinking is that ```isBountyStale``` can only represent that it is or isn't stale but it can't represent both those and inactive at the same time.

I wasn't aware that null would be output as text in the html in the rendered html, I wasn't able to produce this but I will look into it

Keyrxng: ah I feel you now, if it's false show nothing, if its disabled show nothing so no need to check if null

src/bindings/config.ts
Keyrxng: I understand now what you meant and don't know why I didn't see it at first, I had a type error and storing as string solved it but obv I was just approaching it wrong, appreciate it foo

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,26 @@ export const assign = async (body: string) => {
     await addAssignees(issue.number, [payload.sender.login]);
   }
 
+  let days;
+  Number;
```
whilefoo, line 116:
```suggestion

```

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,26 @@ export const assign = async (body: string) => {
     await addAssignees(issue.number, [payload.sender.login]);
   }
 
+  let days;
```
whilefoo, line 115:
```suggestion
  let days: Number;
```

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,26 @@ export const assign = async (body: string) => {
...
+  Number;
+  let staleToDays;
```
whilefoo, line 117:
```suggestion
  let staleToDays: Number;
```

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,26 @@ export const assign = async (body: string) => {
...
+  Number;
```
whilefoo, line 118:
```suggestion
```

whilefoo (changes requested):

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,24 @@ export const assign = async (body: string) => {
     await addAssignees(issue.number, [payload.sender.login]);
   }
 
+  let days: Number;
```
whilefoo, line 115:
```suggestion
  let days: number | undefined;
```

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,24 @@ export const assign = async (body: string) => {
...
+  let staleToDays: Number;
```
whilefoo, line 116:
```suggestion
  let staleToDays: number | undefined;
```

src/handlers/comment/handlers/table.ts
```diff
@@ -4,21 +4,25 @@ export const tableComment = ({
...
 <code>
-
   <table>
-  <tr>
-    <td></td>
-    <td></td>
-  </tr>
+  ${
+    isBountyStale
+      ? `<td>Warning!</td> <td>This task was created over ${days} days ago. Please confirm that this issue specification is accurate before starting.</td>`
```
whilefoo, line 23:
```suggestion
      ? `<tr><td>Warning!</td> <td>This task was created over ${days} days ago. Please confirm that this issue specification is accurate before starting.</td></tr>`
```

whilefoo (changes requested):

src/handlers/comment/handlers/assign.ts
```diff
@@ -109,14 +112,24 @@ export const assign = async (body: string) => {
...
   }
 
+  let days: number | undefined;
+  let staleToDays: number | undefined;
+  let isBountyStale = false;
+
+  if (staleBounty !== 0) {
+    days = Math.floor((new Date().getTime() - new Date(issue.created_at).getTime()) / (1000 * 60 * 60 * 24));
+    staleToDays = Math.floor(staleBounty / (1000 * 60 * 60 * 24));
+    isBountyStale = staleToDays > days ? false : true;
```
whilefoo, line 122:
```suggestion
    isBountyStale = days >= staleToDays;
```
this is the same

whilefoo (changes requested):

whilefoo (approved):
QA:
https://github.com/ubiquity-whilefoo/ubiquibot/issues/87
https://github.com/ubiquity-whilefoo/ubiquibot/issues/88

whilefoo:
@pavlovcik needs your approval

PR #675, opened 2023-08-26
Update ubiquibot-config.yml

pavlovcik:
Removed comment incentive details so that they can be controlled all from https://github.com/ubiquity/ubiquibot-config/edit/development/.github/ubiquibot-config.yml


pavlovcik:
Updated https://github.com/ubiquity/ubiquibot/pull/677

PR #677, opened 2023-08-27
Update ubiquibot-config.yml

pavlovcik:
Simplify repository specific config. In each repository config, we should only specify what is different from the organization wide defaults. 

Related https://github.com/ubiquity/ubiquibot-config/pull/12


web4er (approved):
It is possible that this will fix pricing label issue. And the pricing issue right can only be tested on the original deployment. so let's see how it goes.

whilefoo (approved):

PR #622, opened 2023-08-16
hotfix: check permission when closing issue

wannacfuture:
Hotfix.

QA: https://github.com/wannacfuture/Battleship/issues/83#issuecomment-1680710486

PR #682, opened 2023-08-27
feat: add `issueReopenedCallback` back

whilefoo:
Exploit (https://github.com/ubiquity/ubiquibot/pull/622) is fixed so we can restore `issueReopenedCallback`


pavlovcik (approved):

0xcodercrane (changes requested):
The PR is basically the same as the [commit](https://github.com/ubiquity/ubiquibot/commit/6771daa9c4a672da4438cb32927ffa7940ba0b47) that I pushed directly right?

whilefoo:
> The PR is basically the same as the [commit](https://github.com/ubiquity/ubiquibot/commit/6771daa9c4a672da4438cb32927ffa7940ba0b47) that I pushed directly right?

Yes, it seems like that, we can close this

PR #684, opened 2023-08-28
Config validation error

### LINKED ISSUE #680: Clear Logging Error for broken `ubiquibot-config.yml`
The bot does not show any visible errors for when it has an issue loading in a new config. The bot was down for about 14 hours on this repository until we identified the fix. Apparently it was the broken config. 

There were no logs to verify this, we just changed it because the bot was working on other repositories. 

- https://github.com/ubiquity/ubiquibot/pull/677/files

Context: https://t.me/c/1588400061/3250/4436

The business problem to solve is that we need to know as soon as possible if the config is valid.

- [ ] Clearly log error while the bot tries to load the config in the Supabase.
- [ ] We should consider some type of special event that when the config is updated, the bot immediately validates it and lets the user know via a comment directly on the commit. Similar to how we do deploy comments on: https://github.com/ubiquity/ubiquity-dollar/commit/8fa34df8b37903f2e812cfacd2f1ba438faae095#commitcomment-125340301

whilefoo:
Resolves #680 

Example:
https://github.com/ubiquity-whilefoo/ubiquibot/commit/b74601f5b46603abe0e4b6d6799aaa10ea6404c7#comments
https://github.com/ubiquity-whilefoo/ubiquibot/commit/c8e02c01236b089d702c3b5511af86c5d4f781a3#comments

It was kinda hard to log config errors because if the config was invalid, I also didn't have Supabase keys which the logger needs to write logs to database, so I had to write additional fallbacks


pavlovcik:
First one is great. Second one is vague is there anything you can do to improve that type of message?

whilefoo:
I improved the message:
https://github.com/ubiquity-whilefoo/ubiquibot/commit/386a77299e1df422473177a7e144a6075ec330f6#comments

src/adapters/index.ts
```diff
@@ -6,7 +6,7 @@ export * from "./telegram";
 
 export const createAdapters = (config: BotConfig): Adapters => {
   return {
-    supabase: supabase(config.supabase.url, config.supabase.key),
-    telegram: new Telegraf(config.telegram.token).telegram,
+    supabase: supabase(config?.supabase?.url ?? process.env.SUPABASE_URL, config?.supabase?.key ?? process.env.SUPABASE_KEY),
+    telegram: new Telegraf(config?.telegram?.token ?? process.env.TELEGRAM_BOT_TOKEN).telegram,
```
0xcodercrane, line 10: It's not a good idea to put the env variables here. `env` variables should be read in [config.ts](https://github.com/ubiquity/ubiquibot/blob/development/src/bindings/config.ts)

src/bindings/event.ts
```diff
@@ -43,11 +48,21 @@ export const bindEvents = async (context: Context): Promise<void> => {
...
+    botConfig?.log?.retryLimit ?? 0
+  ); // contributors will see logs in console while on development env
   if (!logger) {
     return;
   }
 
+  if (botConfigError) {
+    logger.error(botConfigError.toString());
+    return;
+  }
```
0xcodercrane, line 64: I guess these lines need to be moved to here.

line 38 - line 42

 ```
 try {
    botConfig = await loadConfig(context);
  } catch (err) {
    botConfigError = err;
  }
```

File: src/handlers/push/index.ts [hunk 3]
```diff
@@ -45,3 +49,63 @@ export const runOnPush = async () => {
...
+        });
+        await context.octokit.rest.repos.createCommitComment({
+          owner: payload.repository.owner.login,
+          repo: payload.repository.name,
+          commit_sha: commitSha,
+          body: `@${payload.sender.login} Config validation failed! Error: ${ajv.errorsText()}. ${
+            additionalProperties && additionalProperties.length > 0 ? `Unnecessary properties: ${additionalProperties.join(", ")}` : ""
+          }`,
+          path: BASE_RATE_FILE,
+        });
```
0xcodercrane, line 107: need to create a helper function in src/helper.ts for this and use the helper function here.

src/utils/private.ts
```diff
@@ -47,49 +49,77 @@ export const getConfigSuperset = async (context: Context, type: "org" | "repo",
...
+    "assistive-pricing": Type.Optional(Type.Boolean()),
+    "max-concurrent-assigns": Type.Optional(Type.Number()),
+    incentives: Type.Optional(IncentivesSchema),
+    "default-labels": Type.Optional(Type.Array(Type.String())),
+    "register-wallet-with-verification": Type.Optional(Type.Boolean()),
+  },
+  {
+    additionalProperties: false,
+  }
+);
```
0xcodercrane, line 112: Looks good to replace `interfaces` with `schemas` and i guess it would be better for you to move the schemas to `src/types`.

src/utils/private.ts
```diff
@@ -160,7 +190,13 @@ export const getWideConfig = async (context: Context) => {
   const repoConfig = await getConfigSuperset(context, "repo", CONFIG_PATH);
 
   const parsedOrg: WideOrgConfig | undefined = parseYAML(orgConfig);
+  if (parsedOrg !== undefined && !ajv.validate(WideOrgConfigSchema, parsedOrg)) {
```
0xcodercrane, line 193:
```suggestion
  if (parsedOrg && !ajv.validate(WideOrgConfigSchema, parsedOrg)) {
```
same thing right?

src/utils/private.ts
```diff
@@ -160,7 +190,13 @@ export const getWideConfig = async (context: Context) => {
...
+    throw new Error(`Invalid org config: ${ajv.errorsText()}`);
+  }
   const parsedRepo: WideRepoConfig | undefined = parseYAML(repoConfig);
+  if (parsedRepo !== undefined && !ajv.validate(WideConfigSchema, parsedRepo)) {
```
0xcodercrane, line 197:
```suggestion
  if (parsedRepo && !ajv.validate(WideConfigSchema, parsedRepo)) {
```

0xcodercrane (changes requested):

src/bindings/event.ts
whilefoo: ideally yes, but logger needs to be initialized for this to work, but logger gets initialized after loading the config because it needs config for supabase and other variables. but if the loading of the config fails then we don't have those variables - this is why I had to make additional fallbacks like [here](https://github.com/ubiquity/ubiquibot/pull/684#discussion_r1309506414)

0xcodercrane: oh right.

web4er:
It would be nice to get the name additional properties in console logs and supabase logs.

```
biquiBot Error: Invalid repo config: data must NOT have additional properties undefined error ubiquibot web4erOrg undefined undefined
```

The bot didn't add comment here on wrong config. 
https://github.com/web4erOrg/ubiquibot/commit/e556157056ef7ab8ae9e7003b02d164277b4c31c


I have only `price-multiplier` and `issue-creator-multiplier` in my org config. If I add only `price-multiplier` in repo config, this shows up in console log and supabase log, but nothing in commit comments. 
```
UbiquiBot Error: Invalid repo config: data must have required property 'issue-creator-multiplier' undefined error ubiquibot web4erOrg undefined undefined
```

whilefoo:
> It would be nice to get the name additional properties in console logs and supabase logs.

Added

> The bot didn't add comment here on wrong config.
> [web4erOrg@e556157](https://github.com/web4erOrg/ubiquibot/commit/e556157056ef7ab8ae9e7003b02d164277b4c31c)

This happened because the config on development was invalid, so you pushed the change, the bot tried to load the current config but it's invalid so it stops. I modified the code to run validateConfigChange even if config load fails.

src/adapters/index.ts
0xcodercrane: bumping @whilefoo here. lets follow up the general process to resolve the conversation. you should at least leave a comment.

0xcodercrane (changes requested):

src/adapters/index.ts
0xcodercrane: Even if config separation resolves the problem finally, we should fix it here as well.

whilefoo: How can we fix it here? Config is not available if it's invalid

0xcodercrane: yeah I mean if the config is not available here(maybe null), we don't need to put env variables for adapters because the bot instance would not work. If we have to forcely make the supabase working for logging, we can allow it only for supabase adapter but not for telegram adapter. Ideally we should remove it from supabase adapter as well in the future.

0xcodercrane, line 10:
```suggestion
    telegram: new Telegraf(config?.telegram?.token ?? "").telegram,
```

0xcodercrane, line 10:
```suggestion
    telegram: new Telegraf(config?.telegram?.token ?? "").telegram,
```

0xcodercrane (changes requested):

src/adapters/index.ts
0xcodercrane: ok lets make it simple. suggested code changes

0xcodercrane: will do it in `development` branch

0xcodercrane (dismissed):

0xcodercrane:
It looks like you should merge the `development` branch into your local `config-validation-error` branch @whilefoo

Issue #378, opened 2023-06-06
Enable ESM support


rndquu:
Depends on https://github.com/probot/probot/issues/1684

We should enable support for ESM-based 3rd party packages to eliminate [such errors](https://github.com/ubiquity/ubiquibot/issues/290#issuecomment-1573450082).

When https://github.com/probot/probot/issues/1684 is ready we should:
1. add `"type": "module"` to `package.json`
2. remove all `require` statements

PR #688, opened 2023-08-29
feat: change to markdown and map html to markdown

### LINKED ISSUE #673: Unpredictable Conversation Rewards
- The conversation rewards should not have been granted to the assignee (me)
- The amount seems way higher than expected. 
- I need to be able to understand why exactly it came to the numbers it did while I calibrate the comment incentives globally and for debugging. Can this be logged?

I have a hypothesis that `monospace` words are extremely heavily credited. Update: I just regenerated the rewards and removed the `monospace` it only was credited as $5. We should update those to $1 but make 
```
code blocks 
```

worth $5.

---

### [ **diamondnegroni: [ CLAIM 2.7 WXDAI ]** ](https://pay.ubq.fi?claim=eyJwZXJtaXQiOnsicGVybWl0dGVkIjp7InRva2VuIjoiMHhlOTFEMTUzRTBiNDE1MThBMkNlOERkM0Q3OTQ0RmE4NjM0NjNhOTdkIiwiYW1vdW50IjoiMjcwMDAwMDAwMDAwMDAwMDAwMCJ9LCJub25jZSI6IjEwNTQwNzQ4MDkwNjM0NzYyMDAzMDMyMzUxOTA3MTk1NjQ3NDg2ODI4MTczODQ1ODUwMjc0NjgwOTM3NTU0MTU2NTY0NDQ1ODExOTYyNiIsImRlYWRsaW5lIjoiMTE1NzkyMDg5MjM3MzE2MTk1NDIzNTcwOTg1MDA4Njg3OTA3ODUzMjY5OTg0NjY1NjQwNTY0MDM5NDU3NTg0MDA3OTEzMTI5NjM5OTM1In0sInRyYW5zZmVyRGV0YWlscyI6eyJ0byI6IjB4QjQ5Yzg5MUMxZTFFNjkzOTU4ZDExRjI3YTkyMEZlZTM0M2FFYjY4ZCIsInJlcXVlc3RlZEFtb3VudCI6IjI3MDAwMDAwMDAwMDAwMDAwMDAifSwib3duZXIiOiIweGY4N2NhNDU4M0M3OTIyMTJlNTI3MjBkMTI3RTdFMEEzOEI4MThhRDEiLCJzaWduYXR1cmUiOiIweGUxZjUzODUyZGVlYjk5NjExMTUxOTVmNTVjNmMwYTk3NzFiNjZjODU4ZGU3MDU0OWExNWVjMzg3YTIwOTM4NjQ1NmQ2YTRiYjAxZjc5Mjk3NzlmMjRlN2NkNWIyNzEyMjEwYzk4MGFhYWFkOTdjNTVhNGEwYThkOTQwZGYxZWU2MWIifQ==&network=100)

_Originally posted by @ubiquibot[bot] in https://github.com/ubiquity/sponsorships/issues/4#issuecomment-1694494805_

whilefoo:
Resolves #673

QA: https://github.com/ubiquity-whilefoo/ubiquibot/issues/56


src/helpers/comment.ts
```diff
@@ -1,42 +1,91 @@
...
   return result;
 };
 
-export const parseComments = (comments: string[], itemsToExclude: string[]): Record<string, string[]> => {
-  const result: Record<string, string[]> = {};
+export const parseComments = async (comments: string[], itemsToExclude: string[]): Promise<Record<MarkdownItem, string[]>> => {
+  // dynamic import of mdast
+  const { fromMarkdown } = await import("mdast-util-from-markdown");
+  const { gfmFromMarkdown } = await import("mdast-util-gfm");
+  const { gfm } = await import("micromark-extension-gfm");
```
whilefoo, line 36: I tried importing normally at the top but I got the error:
```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/whilefoo/ubiquity/ubiquibot/node_modules/mdast-util-from-markdown/index.js from /Users/whilefoo/ubiquity/ubiquibot/lib/src/helpers/comment.js not supported.
    Instead change the require of index.js in /Users/whilefoo/ubiquity/ubiquibot/lib/src/helpers/comment.js to a dynamic import() which is available in all CommonJS modules.
```
After that I changed to dynamic import in the function but the error stays the same. I checked the compiled javascript and it's still `require()` - it seems every `import` is compiled to `require`.

@rndquu any ideas?

rndquu: I guess this is a related issue https://github.com/ubiquity/ubiquibot/issues/378

And this is a related PR https://github.com/ubiquity/ubiquibot/pull/377 (which had been reverted because we got rid of `mdast` packages)

whilefoo: it seems https://github.com/ubiquity/ubiquibot/pull/377 was supposed to fix the issue but it doesn't - instead of throwing the error at startup, it throws the error when the function runs.

rndquu: So it throws the same error both when modules are imported at the top of the `comment.ts` file and in the `parseComments()` function, right?

whilefoo: > So it throws the same error both when modules are imported at the top of the `comment.ts` file and in the `parseComments()` function, right?

yes

rndquu: I don't remember the whole context. As far as I remember this error occurs only during the build step, right?

whilefoo: > As far as I remember this error occurs only during the build step, right?

It's runtime error. If it's imported at the top it throws error when you start the bot, if it's imported in the function it throws when the function runs.

rndquu: As far as I understand [this module](https://github.com/syntax-tree/mdast-util-from-markdown) is ESM only which means that we can use only `import` statements in the bot's code. Since we use commonjs in the bot's code then all `import()` statements are transpiled into `require()` statements hence the `Error [ERR_REQUIRE_ESM]: require() of ES Module` error.

I can only point you to the [docs](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c). Seems like we need to modify smth in typescript settings. This [PR](https://github.com/ubiquity/ubiquibot/pull/377) solved the build step but I don't remember running any `mdast` related code in runtime so chances are that the code in that PR is also broken.

whilefoo: I tried changing Typescript settings to transpile to ESM and added `"type": "module"` to `package.json`, but I get a different error because Probot is using `require()` internally

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/whilefoo/ubiquity/ubiquibot/lib/src/index.js from /Users/whilefoo/ubiquity/ubiquibot/node_modules/probot/lib/helpers/resolve-app-function.js not supported.
Instead change the require of index.js in /Users/whilefoo/ubiquity/ubiquibot/node_modules/probot/lib/helpers/resolve-app-function.js to a dynamic import() which is available in all CommonJS modules.
```

`node_modules/probot/lib/helpers/resolve-app-function.js` is using `require()`

rndquu: Perhaps we could downgrade all 3 packages (`mdast-util-from-markdown`, `mdast-util-gfm`, `micromark-extension-gfm`) to their "pre ESM" versions

whilefoo: I tried downgrading and I have an issue with `mdast-util-gfm` because that version didn't have any types so typescript is complaining. I will see if there's anything I can do

0xcodercrane: https://github.com/ubiquity/ubiquibot/blob/incentive-based-on-comment/src/helpers/comment.ts#L23

Please take a look at the `incentive-based-on-comment` branch @whilefoo. It has been done with `mdast` library. 
The blocker on that branch was the function timeout: 10s as of now.  

I hope you can get reusable/helpful methods from that branch.

whilefoo: > Please take a look at the `incentive-based-on-comment` branch @whilefoo.

That is exactly how I've done it but it's not working, you can check the whole conversation for context

src/handlers/payout/post.ts
```diff
@@ -243,28 +244,63 @@ const generatePermitForComments = async (
...
+  [MarkdownItem.Code]: HTMLItem.PRE,
+  [MarkdownItem.Emphasis]: HTMLItem.EM,
+  [MarkdownItem.Strong]: HTMLItem.STRONG,
+  [MarkdownItem.Delete]: HTMLItem.DEL,
+  [MarkdownItem.HTML]: HTMLItem.HTML,
+  [MarkdownItem.InlineCode]: HTMLItem.CODE,
+  [MarkdownItem.LinkReference]: HTMLItem.A,
+  [MarkdownItem.ImageReference]: HTMLItem.IMG,
+  [MarkdownItem.FootnoteReference]: HTMLItem.SUP,
+  [MarkdownItem.FootnoteDefinition]: HTMLItem.P,
```
whilefoo, line 298: I'm not sure how can we represent footnote reference and definition by HTML? @pavlovcik 

Example of footnote[^1]

[^1]: Footnote 1

src/handlers/payout/post.ts
```diff
@@ -243,28 +244,63 @@ const generatePermitForComments = async (
...
+  [MarkdownItem.Heading3]: HTMLItem.H3,
+  [MarkdownItem.Heading4]: HTMLItem.H4,
+  [MarkdownItem.Heading5]: HTMLItem.H5,
+  [MarkdownItem.Heading6]: HTMLItem.H6,
+  [MarkdownItem.ListItem]: HTMLItem.LI,
+  [MarkdownItem.List]: HTMLItem.UL,
+  [MarkdownItem.Link]: HTMLItem.A,
+  [MarkdownItem.Image]: HTMLItem.IMG,
+  [MarkdownItem.BlockQuote]: HTMLItem.BLOCKQUOTE,
+  [MarkdownItem.Code]: HTMLItem.PRE,
```
whilefoo, line 289: I'm not sure if it's gonna be clear to partners that `<pre>` refers to 
```
code
```
and `<code>` refers to `inline code`

whilefoo:
@pavlovcik can you clarify if task creator and task assignee are eligible for task conversation rewards? I'd guess not because task creator will make comments to clarify his task and task assignee will comment to explain his solution or ask questions to the task creator

rndquu:
> @pavlovcik can you clarify if task creator and task assignee are eligible for task conversation rewards? I'd guess not because task creator will make comments to clarify his task and task assignee will comment to explain his solution or ask questions to the task creator

Oh, I actually thought that task creator is also eligible for task conversation rewards. Ok then, I will update the description for https://github.com/ubiquity/ubiquibot/issues/704

pavlovcik:
> @pavlovcik can you clarify if task creator and task assignee are eligible for task conversation rewards? I'd guess not because task creator will make comments to clarify his task and task assignee will comment to explain his solution or ask questions to the task creator

That's a good observation. Here are all the roles that I see and if they should be eligible for comment incentives. What if we make them configurable?

Something like:

```
### on issue view

- assignee: false
- issuer: false
// - reviewer: n/a
- default: true

### on pull request view 

- assignee: false
// - issuer: n/a
- reviewer: true
- default: true

```

src/handlers/payout/post.ts
pavlovcik: Very cool. I didn't know about this but I guess we can just file an issue and kick the can down the road?

0x4007:
Updated roles:

# Issue
- assignee: false
- issuer: false
// - collaborator: n/a
- contributor: true
### Review
- assignee: false
- issuer: true
- collaborator: true
- contributor: true

Debating on if we should make a "reviewer" role which kicks in if you helped with the review. These are who need to be highly incentivized because review seems to be very burdensome.

PR #691, opened 2023-08-29
chore: merged repo and org config

### LINKED ISSUE #648: Config nested inheritance
We only have inheritance on top level properties, so if `incentives` property is missing from repo and org config, it will use the whole `incentives` property from default config. Nested properties don't have inheritance (if `h1` is not present in repo config it won't inherit from default config), that is something I was meaning to make an issue for.

_Originally posted by @whilefoo in https://github.com/ubiquity/ubiquibot/pull/545#discussion_r1297266427_

For example if we have a organization config:
```yaml
incentives:
  comment:
    elements:
      code: 5
      img: 5
      h1: 1
    totals:
      word: 0.1
```

and repo config:
```yaml
incentives:
  comment:
    elements:
      code: 10
      li: 3
    totals:
      word: 0.1
```

currently because repo config overrides organization config and we parse only top-level properties, the merged config will be the same as repo config, but we want to have inheritance on nested properties too, so the merge config should look like this:
```yaml
incentives:
  comment:
    elements:
      code: 10
      img: 5
      li: 3
      h1: 1
    totals:
      word: 0.1
```

devpanther:
Resolves #648

QA: https://github.com/one-two-test/one/issues/3


whilefoo (changes requested):
If it wasn't clear, the spec is to have nested inheritance for the entire config, not just for incentives. Ideally I'd like to get rid of every function in `helpers.ts` and just merge repo, org and default config using lodash - that way we don't need to keep adding new functions for every property.

0xcodercrane:
we can get the nested inheritance for the entire config without using the lodash simply like this.

`const config = {...parsedOrgConfig, ...parsedRepoConfig}`

What's the benefit to use lodash here? @whilefoo

whilefoo:
> we can get the nested inheritance for the entire config without using the lodash simply like this.
> 
> `const config = {...parsedOrgConfig, ...parsedRepoConfig}`
> 
> What's the benefit to use lodash here? @whilefoo

that would work only for top-level properties not nested, example: https://jsfiddle.net/8qhbj7r2/

devpanther:
> that way we don't need to keep adding new functions for every property.

Got that, I'm pushing a new commit with helpers function removed and lodash used in merging all objs

devpanther:
All configs merged: https://github.com/one-two-test/one/issues/3#issuecomment-1698618254

src/utils/private.ts
```diff
@@ -155,6 +145,10 @@ export const getScalarKey = async (X25519_PRIVATE_KEY: string | undefined): Prom
   }
 };
 
+const mergeConfigs = (configs: MergedConfigs) => {
+  return _.merge({}, ...Object.values(configs));
```
whilefoo, line 149:
```suggestion
  return merge({}, configs.parsedDefault, configs.parsedOrg, configs.parsedRepo);
```
It's better to be explicit about ordering because in this case it matters. We don't know how Object.values() will order them.

src/utils/private.ts
```diff
@@ -2,23 +2,7 @@ import _sodium from "libsodium-wrappers";
...
-  getPriorityLabels,
-  getTimeLabels,
-  getDefaultLabels,
-  getPromotionComment,
-  getIncentives,
-  getAssistivePricing,
-  getCommandSettings,
-  getRegisterWalletWithVerification,
-} from "./helpers";
+import _ from "lodash";
```
whilefoo, line 5:
```suggestion
import merge from "lodash/merge";
```
Let's try to keep bundle size low

whilefoo (changes requested):

whilefoo (dismissed):

whilefoo:
@devpanther please resolve conflicts

devpanther:
> @devpanther please resolve conflicts

resolved

devpanther:
@whilefoo @Draeieg This should be QA'd and merged asap.. there are new configs been added everyday and that's merge conflicts for me

devpanther:
@whilefoo @Draeieg ping

whilefoo (approved):
QA: https://github.com/ubiquity-whilefoo/ubiquibot/issues/84

PR #695, opened 2023-08-31
fix: allow regex

whilefoo:
This fixed the allow regex to allow `-` characters in the username.

### DIFF
--- a/src/handlers/comment/handlers/allow.ts
+++ b/src/handlers/comment/handlers/allow.ts
@@ -20,7 +20,7 @@ export const setAccess = async (body: string) => {
     return;
   }
 
-  const regex = /^\/allow set-(\S+)\s@(\w+)\s(true|false)$/;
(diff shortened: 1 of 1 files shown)


0xcodercrane (changes requested):
const body = "/allow set-multiplier @ubiquibty*addd true"
const regex = /^\/allow\s+set-(\S+)\s+@([\w-]+)\s+(true|false)\s*$/;
body.match(regex)

result: null

wannacfuture:
const body = "/allow set-multiplier @ubiquibtyaddd true"
const regex = /^\/allow set-(\S+)\s@(\S+)\s(true|false)$/;
body.match(regex)

lol it works.

0xcodercrane:
> const body = "/allow set-multiplier @ubiquibtyaddd true" const regex = /^/allow set-(\S+)\s@(\S+)\s(true|false)$/; body.match(regex)
> 
> lol it works.

Yes!

0xcodercrane:
ok I am doing an admin action for both PRs then.

PR #696, opened 2023-08-31
feat: access control config

whilefoo:
Adds config property to enable or disable access control for labels.

QA
Disabled: https://github.com/ubiquity-whilefoo/sandbox/issues/3
Enabled: https://github.com/ubiquity-whilefoo/sandbox/issues/4

When enabled it still has an issue, but we need need this for the demo asap.

### DIFF
--- a/src/utils/private.ts
+++ b/src/utils/private.ts
@@ -1,6 +1,6 @@
 import _sodium from "libsodium-wrappers";
 import YAML from "yaml";
-import { Payload } from "../types";
+import { AccessControl, Payload } from "../types";
 import { Context } from "probot";
 import {
   getAnalyticsMode,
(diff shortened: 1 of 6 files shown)


rndquu:
@whilefoo 

> When enabled it still has an issue

Which one?

src/utils/private.ts
```diff
@@ -67,6 +68,10 @@ export interface CommandObj {
   enabled: boolean;
 }
 
+export interface AccessControl {
+  label: boolean;
+}
+
```
0xcodercrane, line 74: Please add this to src/types/config.ts

```suggestion
export type AccessControl = Static<typeof AccessControlSchema>;

```

whilefoo:
> @whilefoo
> 
> > When enabled it still has an issue
> 
> Which one?


You can see the problem here https://github.com/ubiquity-whilefoo/sandbox/issues/4. Because the issue is created with 2 labels, github sends 3 webhooks. 

- issues.opened - it doesn't check permissions so it adds Price label
- issues.labeled - both webhooks are for label `Time: <1 Hour`, which is why the bot posts the comment twice

PR #697, opened 2023-08-31
feat: toggle permit generation by external

web4er:
> another thing discussed with pavlovcik in DM
> 
> check this comment https://github.com/ubiquity/ubiquibar/issues/13#issuecomment-1701225272
> 
> When issue is closed by an outside collaborator then the permit generation is skipped
> 
> When https://github.com/ubiquity/ubiquibot/pull/696/ is reviewed and merged I will add a new bot config param enable-access-control.organization that enables/disables this [check](https://github.com/rndquu/bounty-bot/blob/d0a7f796114bea474f26a1eeb003d72c6d61cd78/src/handlers/payout/action.ts#L38)
> 
> So we could enable this new param for all of our organizations but disable for certain repos temporary
> https://t.me/c/1588400061/1/4590

QA: For external user `enable-access-control.organization=false`  https://github.com/web4erOrg/web4er-sandbox/issues/11
QA: For external user `enable-access-control.organization=true`  https://github.com/web4erOrg/web4er-sandbox/issues/15 (strangely still creates permit for task creator but I think not issue of concern at the moment)

QA: For internal user `enable-access-control.organization=false` https://github.com/web4erOrg/web4er-sandbox/issues/12
QA: For internal user `enable-access-control.organization=true` https://github.com/web4erOrg/web4er-sandbox/issues/14

### DIFF
--- a/ubiquibot-config-default.json
+++ b/ubiquibot-config-default.json
@@ -87,6 +87,7 @@
     }
   },
   "enable-access-control": {
-    "label": false
+    "label": false,
+    "organization": true
   }
 }
--- a/src/handlers/payout/action.ts
+++ b/src/handlers/payout/action.ts
@@ -25,6 +25,7 @@ export const handleIssueClosed = async () => {
   const {
     payout: { paymentToken, rpc, permitBaseUrl, networkId, privateKey },
     mode: { paymentPermitMaxPrice },
+    accessControl,
   } = getBotConfig();
   const logger = getLogger();
   const payload = context.payload as Payload;
@@ -35,9 +36,11 @@ export const handleIssueClosed = async () => {
 
   if (!issue) return;
 
-  const userHasPermission = await checkUserPermissionForRepoAndOrg(payload.sender.login, context);
+  if (accessControl.organization) {
+    const userHasPermission = await checkUserPermissionForRepoAndOrg(payload.sender.login, context);
 
-  if (!userHasPermission) return "Permit generation skipped because this issue has been closed by an external contributor.";
+    if (!userHasPermission) return "Permit generation skipped because this issue has been closed by an external contributor.";
+  }
 
   const comments = await getAllIssueComments(issue.number);
 
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -95,6 +95,7 @@ export const WalletSchema = Type.Object({
 
 export const AccessControlSchema = Type.Object({
   label: Type.Boolean(),
+  organization: Type.Boolean(),
 });
 
 export type AccessControl = Static<typeof AccessControlSchema>;


ubiquibot-config-default.json
```diff
@@ -87,6 +87,7 @@
     }
   },
   "enable-access-control": {
-    "label": false
+    "label": false,
+    "organization": true
```
rndquu, line 91: All of the bot's features should be disabled by default, it was discussed today somewhere in telegram :)

rndquu (changes requested):

rndquu (approved):

ubiquibot-config-default.json
pavlovcik: I wish that the property name was more clear. When I saw this in `ubiquity/ubiquibot-config/.github/ubiquibot-config.yml` I had to trace this down to this pull request to understand what it does. 

Perhaps we can rename to: 

```
  "public-access-control": {
     "set-label": false,
     "fund-external-closed-issue": false
```

What do you guys think?

rndquu: Those new params were introduced solely for the demo purposes for the `Ethcon Korea 2023` conference

I would simply delete those new params because:
1. I can't image a case when a partner wants to tweak one of those params
2. If, by chance, partner forgets to set those new params to `true` then it poses a significant security issue

@0xcodercrane What do you think?

pavlovcik: I like the idea of being proactive about the security for our users, but I feel like they really have to go out of their way to dig up this config property. In our documentation we could just make it super clear that it is a significant security risk to enable them, but, there might be reasons why other organizations want these enabled?

rndquu: > In our documentation we could just make it super clear that it is a significant security risk to enable them

On the contrary, it is a significant security risk to disable [label](https://github.com/ubiquity/ubiquibot-config/blob/694633a8e6f2d5ae7ffe23f97f625c354737dca2/.github/ubiquibot-config.yml#L81) and [organization](https://github.com/ubiquity/ubiquibot-config/blob/694633a8e6f2d5ae7ffe23f97f625c354737dca2/.github/ubiquibot-config.yml#L82) access controls

Anyway, I feel like that in the future (with 0 unit testing) we will get a security issue related to one of those params

0xcodercrane: Agree to rename the property names: I had the same feeling to guess what the property does. took a little bit

pavlovcik: > Anyway, I feel like that in the future (with 0 unit testing) we will get a security issue related to one of those params

Then I guess lets label the commit, remove it from the codebase, and restore the capability based on the commit label in the future if we need it again. 

Alternatively we could have a "demo" instance of the bot because I plan to be presenting it a lot more later this year and using the same demo script (order a bag or shirt) would be fantastic

rndquu: >  I plan to be presenting it a lot more later this year

And we're getting the case for those access control params. Ok then, let's simply [rename](https://github.com/ubiquity/ubiquibot/issues/742) them for now.

PR #722, opened 2023-09-05
fix: ignore param order for allow

### LINKED ISSUE #713: `/allow` parameter ordering significance
> I never used this command successfully first try. We should make parameter ordering not significant.

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/711#issuecomment-1705100822_

byteballet:
Resolves #713
QA for https://github.com/ubiquibot/staging/issues/177

The bot that concerns params order for /allow alerts me a message like https://github.com/ubiquibot/staging/issues/176#issuecomment-1706243557
Now I configured the bot to not concern params order and it ignores param order.

### DIFF
--- a/src/handlers/comment/handlers/allow.ts
+++ b/src/handlers/comment/handlers/allow.ts
@@ -20,40 +20,20 @@ export const setAccess = async (body: string) => {
     return;
   }
 
-  const regex =
-    /^\/allow (?:set-(\S+)\s@(\w+)\s(true|false)|set-(\S+)\s(true|false)\s@(\w+)|(true|false)\sset-(\S+)\s@(\w+)|(true|false)\s@(\w+)\sset-(\S+)|@(\w+)\s(true|false)\sset-(\S+)|@(\w+)\sset-(\S+)\s(true|false))$/;
+  const regex = /\/allow\s+(\S+)\s+(\S+)\s+(\S+)/;
   const matches = body.match(regex);
 
   if (matches) {
     let accessType, username, bool;
-    if (matches[1]) {
-      accessType = matches[1];
-      username = matches[2];
-      bool = matches[3];
-    } else if (matches[4]) {
-      accessType = matches[4];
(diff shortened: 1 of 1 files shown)


whilefoo (changes requested):
That seems verbose and unreadable. It would be better to just check if it starts with /allow then split the whole string by whitespace, check the length first and then in a for loop check if it's starts with @, "set-" or true/false

src/handlers/comment/handlers/allow.ts
```diff
@@ -20,13 +20,20 @@ export const setAccess = async (body: string) => {
...
+  const regex = /\/allow\s+(\S+)\s+(\S+)\s+(\S+)/;
   const matches = body.match(regex);
 
   if (matches) {
-    const [, accessType, username, bool] = matches;
-
+    let accessType, username, bool;
+    matches.slice(1).forEach((part) => {
+      if (part.startsWith("@")) username = part.slice(1);
+      else if (part.startsWith("set-")) accessType = part.slice(4);
```
whilefoo, line 30: why `.slice()`? isn't `part` already the full string like `set-multiplier`?
also you should check if variables are already set (not undefined) because somebody can put username twice like `/allow @example @example true`

whilefoo (changes requested):

src/handlers/comment/handlers/allow.ts
byteballet: `accessType` should be one of ["priority", "time", "price", "multiplier"] not starting with set-
And do you mean `/allow @example @example true` is right?
In above case, there's no accessType and it will log error and will end...

whilefoo (approved):
QA: https://github.com/ubiquity-whilefoo/ubiquibot/issues/81

whilefoo:
@0xcodercrane I went ahead and merged the PR - @pavlovcik asked me to help with processing PRs.

PR #730, opened 2023-09-06
fix : Renamed ubiquibot-config-default file 

### LINKED ISSUE #725: Refactor default bot config from `json` to `ts`
Original [comment](https://github.com/ubiquity/ubiquibot/pull/670#discussion_r1306600068)

There is the [default bot config](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json). 

We want to able to add comments to the default bot config. We can't transform the config from `json` to `yml` or `jsonc` because it requires using `fs.readFile()` which causes issues on production netlify build.

What should be done:
- refactor the [default bot config](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json) from `json` to a plain `ts` file

Wholesomebruh:
Resolves #725 
Changed json to ts as required.

### DIFF
--- a/ubiquibot-config-default.ts
+++ b/ubiquibot-config-default.ts
@@ -1,4 +1,34 @@
-{
+interface Config {
+  "evm-network-id": number;
+  "price-multiplier": number;
+  "issue-creator-multiplier": number;
+  "payment-permit-max-price": number;
+  "max-concurrent-assigns": number;
+  "assistive-pricing": boolean;
+  "disable-analytics": boolean;
+  "comment-incentives": boolean;
+  "register-wallet-with-verification": boolean;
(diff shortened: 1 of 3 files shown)


Wholesomebruh:
It's still in progress, will resolve the errors as soon as possible.

ubiquibot-config-default.ts
```diff
@@ -1,4 +1,34 @@
-{
+interface Config {
```
whilefoo, line 1: why are you creating types again

whilefoo (changes requested):

Wholesomebruh:
closed this due to many commits.

PR #738, opened 2023-09-08
Parse Permit Url

### LINKED ISSUE #526: Parse permit URL

Sadaf-A:
Resolves #526

### DIFF
--- a/src/handlers/payout/action.ts
+++ b/src/handlers/payout/action.ts
@@ -191,12 +191,10 @@ export const handleIssueClosed = async () => {
   const comment =
     `#### Task Assignee Reward\n### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
   const permitComments = comments.filter((content) => {
-    if (content.body.includes(DEFAULT_PERMIT_BASE_URL) && content.user.type == UserType.Bot) {
-      const url = new URL(content.body);
-      // Check if the URL contains the specific query parameter
-      return url.searchParams.has("claim");
-    }
-    return false;
+    claimUrlRegex;
+    const permitUrlMatches = content.body.match(claimUrlRegex);
+    if (!permitUrlMatches || permitUrlMatches.length < 2) return false;
+    else return true;
   });
   if (permitComments.length > 0) {
     logger.info(`Skip to generate a permit url because it has been already posted.`);
--- a/src/handlers/payout/action.ts
+++ b/src/handlers/payout/action.ts
@@ -19,7 +19,6 @@ import { bountyInfo } from "../wildcard";
 import Decimal from "decimal.js";
 import { GLOBAL_STRINGS } from "../../configs";
 import { isParentIssue } from "../pricing";
-import { DEFAULT_PERMIT_BASE_URL } from "../../configs";
 
 export const handleIssueClosed = async () => {
   const context = getBotContext();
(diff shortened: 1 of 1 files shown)


src/handlers/payout/action.ts
```diff
@@ -189,7 +189,14 @@ export const handleIssueClosed = async () => {
   logger.info(`Posting a payout url to the issue, url: ${payoutUrl}`);
   const comment =
     `#### Task Assignee Reward\n### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
-  const permitComments = comments.filter((content) => content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot);
+  const permitComments = comments.filter((content) => {
+    if (content.body.includes("https://pay.ubq.fi") && content.user.type == UserType.Bot) {
```
whilefoo, line 193: use `permitBaseUrl` from the config and extract this into helper function and use it every where we search for permits

whilefoo (changes requested):
Also post QA

src/handlers/payout/action.ts
Sadaf-A: @whilefoo Hi, please confirm what should be done here

whilefoo: No you don't need to add `?claim`

src/handlers/payout/action.ts
```diff
@@ -19,6 +19,7 @@ import { bountyInfo } from "../wildcard";
 import Decimal from "decimal.js";
 import { GLOBAL_STRINGS } from "../../configs";
 import { isParentIssue } from "../pricing";
+import { DEFAULT_PERMIT_BASE_URL } from "../../configs";
```
whilefoo, line 22: not like this, you need to use `getBotConfig`. also post QA

src/handlers/payout/action.ts
```diff
@@ -189,7 +190,14 @@ export const handleIssueClosed = async () => {
   logger.info(`Posting a payout url to the issue, url: ${payoutUrl}`);
   const comment =
     `#### Task Assignee Reward\n### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
-  const permitComments = comments.filter((content) => content.body.includes("https://pay.ubq.fi?claim=") && content.user.type == UserType.Bot);
+  const permitComments = comments.filter((content) => {
```
whilefoo, line 193: extract this into helper function and apply it everywhere where we search for permit

whilefoo (changes requested):

0xcodercrane (approved):

src/handlers/payout/action.ts
0xcodercrane: If we make it as a helper fn, that would be cool but not so important for now. To speed up the velocity, mark it as resolved

0xcodercrane (approved):

whilefoo:
the parsing should be done with URLSearchParams to avoid dependence on param order like @Sadaf-A did. I don't understand why you changed it to regex

Sadaf-A:
> the parsing should be done with URLSearchParams to avoid dependence on param order like @Sadaf-A did. I don't understand why you changed it to regex

@whilefoo 
@0xcodercrane made those changes so I didn't wanna make any more changes i fixed the errors that arised after it

whilefoo:
> @whilefoo
> @0xcodercrane made those changes so I didn't wanna make any more changes i fixed the errors that arised after it earlier i was also using URLSearchParams as requested by you

yes I know @0xcodercrane made those changes

0xcodercrane:
We don't create a complex claim URL here, so I guess no need to worry about the param order.

In upper lines, we were also using regex so wanted to make the parse consistent. 

If the order problem happens, lets update it at that time.

pavlovcik:
This isn't the spec

Sadaf-A:
> the parsing should be done with URLSearchParams to avoid dependence on param order like @Sadaf-A did. I don't understand why you changed it to regex

@whilefoo @pavlovcik should i open another PR for this?

PR #744, opened 2023-09-10
Fix : skipped->disabled , since -> because

### LINKED ISSUE #741: Change Permit Generation Error Messages
> Permit generation skipped since the issue is identified as parent issue.

Should be instead:

Permit generation disabled because this is a collection of issues. 

---

I think for all the similar messages, we should say: 
- "disabled" instead of "skipped" 
- "because" instead of "since"

And ensure punctuation at the end of every message.

_Originally posted by @ubiquibot[bot] in https://github.com/ubiquity/telegram-ubiquibot/issues/2#issuecomment-1712862573_

Wholesomebruh:
Resolves #741

### DIFF
--- a/src/handlers/comment/action.ts
+++ b/src/handlers/comment/action.ts
@@ -40,8 +40,8 @@ export const handleComment = async (): Promise<void> => {
       const feature = config.command.find((e) => e.name === id.split("/")[1]);
 
       if (!feature?.enabled && id !== IssueCommentCommands.HELP) {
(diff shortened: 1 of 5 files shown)


EtherealGlow:
@Wholesomebruh uhhh

pavlovcik (approved):
This is good but unfortunately you are not the assignee of the bounty which means technically you are not eligible to collect the bounty. Please be sure to either assign yourself before you start working, or open a pull request and associate it to the issue first.

Issue #681, opened 2023-08-27
Reopened & Unassigned


pavlovcik:
The bot should be careful to not unassign like this when a task is reopened. It should reset the "timer" e.g. to be a full 7 days again.

> @whilefoo please be sure to review this conversation and implement any necessary fixes. Unless this is closed as completed, its payment of **450.0 WXDAI** will be deducted from your next bounty.

_Originally posted by @ubiquibot[bot] in https://github.com/ubiquity/ubiquibot/issues/500#issuecomment-1694619062_

wannacfuture:
https://github.com/ubiquity/ubiquibot/issues/685#issuecomment-1698760710

The solution will be similar.

we will find bot's comment saying "please be sure to review this conversation ..."
and consider that comment's createdAt as the activity time of the hunter as well.

Keyrxng:
happy to take this if someone can assign it to me as I'm waiting for prs and haven't much to do this weekend 🤷‍♂️

Keyrxng:
I will catch this tomorrow

Keyrxng:
woops missed the reminder on this one

0xcodercrane:
Why conversationRewards includes @Keyrxng he is an assigneed seems bug.

Keyrxng:
Thanks @0xcodercrane

So conversation rewards are only for non-assignees, is that across issues, discussions and prs?

Keyrxng:
- https://github.com/Keyrxng/UbiquityBotTesting/issues/30#issuecomment-1727659530

- https://github.com/Keyrxng/UbiquityBotTesting/issues/45#issuecomment-1727658931

I just ran a fresh dev branch build after seeing @whilefoo comment on tg about hitting errors with events. No errors with events, but seems like the bot is still unassigning after a closed issue is re-opened see [example](https://github.com/Keyrxng/UbiquityBotTesting/issues/40#issuecomment-1727678674) & [example](https://github.com/Keyrxng/UbiquityBotTesting/issues/19#issuecomment-1727694011) 

The first two linked comments show that the bot spammed the comment 4 times for allocated duration has ended, these issues were just open and not reopened. This happened across 6 issues that the duration had expired on. Been a few days since I've run the bot right enough so I rightfully should have been unassigned but the spamming? Is that a product of not having run the bot for days, you folks would have a better idea if it's happening on other issues as you'll see far more of these events live than I will.

PR #749, opened 2023-09-10
fix: use bot penalty

Keyrxng:
Checks if the bot penalty comment created at is more than last activity, if so it returns only the delivery timeline else it will unnassign.

Resolves https://github.com/ubiquity/ubiquibot/issues/681

Example https://github.com/Keyrxng/UbiquityBotTesting/issues/11#issuecomment-1712968225

For testing purposes I added the penalty comment myself to workaround not having setup the backend fully yet but I believe it shows the logic working as intended.

My issue is old enough that it was unassigning me as you'll see with the bot doing so in the interaction list above the comments, after the fix it no longer unassigns.

### DIFF
--- a/src/handlers/wildcard/unassign.ts
+++ b/src/handlers/wildcard/unassign.ts
@@ -9,9 +9,9 @@ import {
   getReviewRequests,
   listAllIssuesForRepo,
   removeAssignees,
-  wasIssueReopened,
 } from "../../helpers";
-import { Comment, Issue, IssueType, Payload } from "../../types";
+import { Comment, Issue, IssueType, Payload, UserType } from "../../types";
+import { deadLinePrefix } from "../shared";
 
 /**
  * @dev Check out the bounties which haven't been completed within the initial timeline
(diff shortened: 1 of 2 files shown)


pavlovcik:
I believe that it makes a lot more sense to check the timestamp of when it was reopened, instead of the follow up comment.

0xcodercrane:
should get the timestamp from the reopened event and recalculate the deadline @Keyrxng

0xcodercrane (changes requested):
you should get the timestamp from the reopened event and recalculate the deadline @Keyrxng

wannacfuture (commented):
I think this issue's purpose is to make bot to not unassign hunter right after manual assigning.

I can't see it on your QA, @Keyrxng 
Would you mind posting a valid one?

Or we can do it in a more simple way.
https://github.com/ubiquity/ubiquibot/pull/759/files
@pavlovcik , @0xcodercrane

Keyrxng:
> I think this issue's purpose is to make bot to not unassign hunter right after manual assigning.
> 
> I can't see it on your QA, @Keyrxng Would you mind posting a valid one?
> 
> Or we can do it in a more simple way. https://github.com/ubiquity/ubiquibot/pull/759/files @pavlovcik , @0xcodercrane

I had read the issue as the bot unassigning once an old issue has been reopened which forced the bot to unassign due to it being over the deliverable date, the manual assign being automatically unassigned thing is just a by-product of this. 

After my impl the bot stopped trying to unassign me after I manually reassigned.

But by all means if your linked pr solves this issue then you take it

0xcodercrane:
> I think this issue's purpose is to make bot to not unassign hunter right after manual assigning.
> 
> I can't see it on your QA, @Keyrxng Would you mind posting a valid one?
> 
> Or we can do it in a more simple way. https://github.com/ubiquity/ubiquibot/pull/759/files @pavlovcik , @0xcodercrane

The purpose is to reset the deadline when you reopen the closed issue  + stop unassigning the hunter at that time.

0xcodercrane:
took a look at your PR for this task. It would also work but you have been solving another issue. @wannacfuture

wannacfuture:
> took a look at your PR for this task. It would also work for this issue but you have been solving another issue. @wannacfuture.

the root cause of the PR's related issue also contains fixing deadline.

Keyrxng:
> 2/ The assignee should still be Keyrxng.

This feels sort of wrong, you wanna at least split it @wannacfuture?

wannacfuture:
> > 2/ The assignee should still be Keyrxng.
> 
> This feels sort of wrong, you wanna at least split it @wannacfuture?

Just take it. @Keyrxng can think as hotfix. no problem.

Keyrxng:
Humbled mate, I appreciate it

0xcodercrane:
Thanks both for the great swarming lol

0xcodercrane (approved):
This would work. merging

PR #750, opened 2023-09-11
Config Name Format

### LINKED ISSUE #711: Config name format
Yeah you're right, this will be a problem because internal config uses camelCase and yaml config uses kebab-case, so this will cause trouble in the future too. I think it would be best to stick to one format for both configs - camelCase is natural in JS/TS so I think it should be also used in yaml config. I will make a new issue for this but for now modify `getNewContributorConfig` to read from kebab-case and convert to camelCase manually.

_Originally posted by @whilefoo in https://github.com/ubiquity/ubiquibot/pull/689#discussion_r1314242213_

I think we should change yaml name format to camelCase to avoid having to convert it from kebab-case to camelCase when we parse it. This will also allow to have one unified Typescript type.

On this note I'd like to also discuss the separation of the config into two configs. One config would be EnvConfig which parses env variables and another would be the RepoConfig which parses the yaml. Having one merged config was causing issues in #684 because I had to validate the config and if it's invalid I still needed env variables to create Supabase adapter and create logger to log the error.

Sadaf-A:
Resolves #711

### DIFF
--- a/src/types/config.ts
+++ b/src/types/config.ts
@@ -141,7 +141,7 @@ export const WideConfigSchema = Type.Object(
   {
     "evm-network-id": Type.Optional(Type.Number()),
     "price-multiplier": Type.Optional(Type.Number()),
-    "issue-creator-multiplier": Type.Number(),
+    "issue-creator-multiplier": Type.Optional(Type.Number()),
     "time-labels": Type.Optional(Type.Array(LabelItemSchema)),
     "priority-labels": Type.Optional(Type.Array(LabelItemSchema)),
     "payment-permit-max-price": Type.Optional(Type.Number()),


0xcodercrane (changes requested):
A couple of stuff I don't understand in the PRs.

1/ Why do we change even workflow filenames? Is this intended action item. If I understood the origin spec correctly, it should be config-related task. please verify me @whilefoo @Sadaf-A 
2/ If we change ubiquibot-config.yml to ubiquibotConfig.yml, we should also update the codebase (there is a line to read config file: ubiquibot-config.yml in the codebase).

whilefoo (changes requested):
Like @0xcodercrane said, this task is only related to the config, so you don't need to change workflows and other names. Only properties inside the config should be in camelCase.

Sadaf-A:
> Like @0xcodercrane said, this task is only related to the config, so you don't need to change workflows and other names. Only properties inside the config should be in camelCase.

in the config folder there are all typescript files. can you link the .yml file?

whilefoo:
https://github.com/ubiquity/ubiquibot/blob/development/.github/ubiquibot-config.yml

PR #756, opened 2023-09-12
feat: extract time duration with ms

### LINKED ISSUE #431: `value` should be dynamically parsed from the time value using the `ms` library

wannacfuture:
Resolves #431

Quality Assurance: https://github.com/wannacfuture/Battleship/issues/88

### DIFF
--- a/src/helpers/shared.ts
+++ b/src/helpers/shared.ts
@@ -1,3 +1,4 @@
+import ms from "ms";
 import { getBotContext } from "../bindings";
 import { LabelItem, Payload, UserType } from "../types";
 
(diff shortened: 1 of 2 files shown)


src/helpers/shared.ts
```diff
@@ -35,13 +36,7 @@ export const calculateWeight = (label: LabelItem | undefined): number => {
...
   if (label.name.toLowerCase().includes("priority")) return 0;
-  const number = matches && matches.length > 0 ? parseInt(matches[0]) || 0 : 0;
-  if (label.name.toLowerCase().includes("minute")) return number * 60;
-  if (label.name.toLowerCase().includes("hour")) return number * 3600;
-  if (label.name.toLowerCase().includes("day")) return number * 86400;
-  if (label.name.toLowerCase().includes("week")) return number * 604800;
-  if (label.name.toLowerCase().includes("month")) return number * 2592000;
-  return 0;
+
+  return ms(label.name.split("<")[1]) / 1000;
```
0xcodercrane, line 41: could we think of a way to use regex here rather than using `split`?

0xcodercrane (changes requested):

rndquu (approved):
Works fine

wannacfuture:
seems can be merged due to its a small fix. @pavlovcik

0xcodercrane (approved):

PR #759, opened 2023-09-12
feat: fix automatic unassign

### LINKED ISSUE #685: Can't regenerate payment
We need to come up with an adjustment to this flow because the bot won't let me regenerate the payment permit because the bot will not let me keep the assign. 

I think on the assign event (and any event of specific actions) the bot should be mindful to not undo the user's action. For example, when a new assignee is added like in the referenced example, perhaps the bot should never unassign that assignee on the same issue as a response to the event. 

Or if I add a specific label, perhaps the bot should be mindful to never remove that same added label.  

> @sergfeldman - Releasing the bounty back to dev pool because the allocated duration already ended! 
Last activity time: Thu Jul 20 2023 09:57:04 GMT+0000 (Coordinated Universal Time)

_Originally posted by @ubiquibot in https://github.com/ubiquity/card-issuance/issues/14#issuecomment-1696251912_

wannacfuture:
Resolves #685

Fixed lastActivityTime function


wannacfuture:
ChatGPT is smarter than I expected 🤣

But updating lastActivityTime function is enough to prevent bot from unassigning hunter right after manual assignment.
Mr. GPT, lol

> Or if I add a specific label, perhaps the bot should be mindful to never remove that same added label.

This is already implemented, I think. E.g. when I add priority label bot removes original priority label.

@pavlovcik

pavlovcik:
I'm pretty sure its not implemented. 

I remember asking about chaining together handlers and sharing context to ensure that it understands the user action that invoked the bot and does not undo it and you saying that's difficult to do. 

Should be for everything including for now: assigns, and price label

cuddle-dev (changes requested):
https://github.com/ubiquity/ubiquibot/issues/685#issuecomment-1716906587

commented on the bounty.

wannacfuture:
> I'm pretty sure its not implemented.
> 
> I remember asking about chaining together handlers and sharing context to ensure that it understands the user action that invoked the bot and does not undo it and you saying that's difficult to do.
> 
> Should be for everything including for now: assigns, and price label

https://github.com/ubiquity/card-issuance/issues/14

On the above issue, the bot removes the price label right after you have set it.
But it didn't remove the label which you have added. It removed the original label.

For example, if there was 200 USD label, and you add 300 USD label, the bot will remove 200 label and leave 300.
Isn't this the right behaviour?

> Or if I add a specific label, perhaps the bot should be mindful to never remove that same added label.

Not so sure what should be done exactly for this.
@pavlovcik

pavlovcik:
image

Looks wrong

wannacfuture:
haha you battery is running out lol

I investigated about it. 
The bot removes all the price labels which isn't appropriate for given priority and time labels even you set a price label in your mind manually.

in the beginning of the issue the bot added 200 USD for priority 1 time 1 day.
but after that, not sure if the rule of pricing is changed(maybe config??) but the bot accepted 200 USD for priority 2 and time 1 day at last.

So that's why bot removed your 200USD label after you labeled it, because he thought 200 is not appropriate for priority1 and time 1 day.

But I'm not sure why he didn't add right label after removing yours even it is designed to do in the code. We need to investigate the logs. 

So I suggest this:
- monitoring issues and check appropriate price labels exist on it for given priority & time labels.

pavlovcik:
> It does not fully satisfy the requirements regarding the bot undoing user actions (e.g., unassigning users or removing labels).

I don't get why this is difficult to understand. Regardless of the details of what the bot thinks the right thing to do is currently, it should respect the user's wishes like setting the label or setting assigns. This should be generalized to work with any user action because we will continue to add capabilities.

wannacfuture:
> > It does not fully satisfy the requirements regarding the bot undoing user actions (e.g., unassigning users or removing labels).
> 
> I don't get why this is difficult to understand.
> 
> Regardless of the details of what the bot thinks the right thing to do is currently, it should respect the user's wishes like setting the label or setting assigns.
> 
> This should be generalized to work with any user input/action because we will continue to add capabilities.

It does make sense.

I pushed commit but is it enough to check there is existing price label? @pavlovcik

pavlovcik:
> I pushed commit but is it enough to check there is existing price label? @pavlovcik

> This should be generalized to work with any user input/action because we will continue to add capabilities.

0xcodercrane:
we're fine to merge the PR @wannacfuture @pavlovcik? no comments from me

cuddle-dev (approved):

whilefoo:
it seems the code already got merged from somewhere else?
https://github.com/ubiquity/ubiquibot/blob/9d00d7f8be6e7b2eeb91f8a01246a1d9697881d7/src/handlers/wildcard/unassign.ts#L104

wannacfuture:
> it seems the code already got merged from somewhere else?
> 
> https://github.com/ubiquity/ubiquibot/blob/9d00d7f8be6e7b2eeb91f8a01246a1d9697881d7/src/handlers/wildcard/unassign.ts#L104

Yeah, another hunter pulled this branch.

rndquu:
@whilefoo Good to see you're back online 💊 👍

PR #763, opened 2023-09-13
Fix: renamed default config file

### LINKED ISSUE #725: Refactor default bot config from `json` to `ts`

Wholesomebruh:
Resolves #725 

Quality Assurance: Passing checks

### DIFF
--- a/src/utils/private.ts
+++ b/src/utils/private.ts
@@ -4,7 +4,7 @@ import { MergedConfig, Payload } from "../types";
 import { Context } from "probot";
 import merge from "lodash/merge";
 
-import defaultConfig from "../../ubiquibot-config-default";
+import DefaultConfig from "../../ubiquibot-config-default";
 import { validate } from "./ajv";
 import { WideConfig, WideOrgConfig, WideRepoConfig, WideConfigSchema, WideOrgConfigSchema } from "../types";
 
@@ -126,7 +126,7 @@ export const getWideConfig = async (context: Context) => {
       throw new Error(`Invalid repo config: ${error}`);
     }
   }
-  const parsedDefault: MergedConfig = defaultConfig;
+  const parsedDefault: MergedConfig = DefaultConfig;
   const privateKeyDecrypted = parsedOrg && parsedOrg[KEY_NAME] ? await getPrivateKey(parsedOrg[KEY_NAME]) : undefined;
 
   const configs: MergedConfigs = { parsedDefault, parsedOrg, parsedRepo };
--- a/src/utils/private.ts
+++ b/src/utils/private.ts
@@ -4,7 +4,7 @@ import { MergedConfig, Payload } from "../types";
 import { Context } from "probot";
 import merge from "lodash/merge";
 
-import DefaultConfig from "../../ubiquibot-config-default";
+import { DefaultConfig } from "../configs";
 import { validate } from "./ajv";
 import { WideConfig, WideOrgConfig, WideRepoConfig, WideConfigSchema, WideOrgConfigSchema } from "../types";
 
--- a/src/utils/private.ts
+++ b/src/utils/private.ts
@@ -4,7 +4,7 @@ import { MergedConfig, Payload } from "../types";
 import { Context } from "probot";
 import merge from "lodash/merge";
 
-import DEFAULT_CONFIG_JSON from "../../ubiquibot-config-default.json";
+import defaultConfig from "../../ubiquibot-config-default";
 import { validate } from "./ajv";
 import { WideConfig, WideOrgConfig, WideRepoConfig, WideConfigSchema, WideOrgConfigSchema } from "../types";
 
@@ -126,7 +126,7 @@ export const getWideConfig = async (context: Context) => {
       throw new Error(`Invalid repo config: ${error}`);
     }
   }
-  const parsedDefault: MergedConfig = DEFAULT_CONFIG_JSON;
+  const parsedDefault: MergedConfig = defaultConfig;
   const privateKeyDecrypted = parsedOrg && parsedOrg[KEY_NAME] ? await getPrivateKey(parsedOrg[KEY_NAME]) : undefined;
 
   const configs: MergedConfigs = { parsedDefault, parsedOrg, parsedRepo };
--- a/ubiquibot-config-default.ts
+++ b/ubiquibot-config-default.ts
@@ -1,4 +1,6 @@
-{
+import { MergedConfig } from './src/types/config'
+
+const defaultConfig : MergedConfig = {
   "evm-network-id": 100,
   "price-multiplier": 1,
   "issue-creator-multiplier": 2,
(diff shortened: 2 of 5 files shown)


Wholesomebruh:
@pavlovcik @rndquu , last PR was automatically closed as GitHub suggested to remove the commit and then resolve the conflicts. No extra changes are made. Please have a look at this.

pavlovcik:
"QA" needs evidence like links

ubiquibot-config-default.ts
```diff
@@ -90,5 +91,8 @@
   "enable-access-control": {
     "label": false,
     "organization": true
-  }
+  },
+  "stale-bounty-time":""
```
pavlovcik, line 95: This doesn't look right. Why did you remove "0d"

src/utils/private.ts
```diff
@@ -4,7 +4,7 @@ import { MergedConfig, Payload } from "../types";
 import { Context } from "probot";
 import merge from "lodash/merge";
 
-import DEFAULT_CONFIG_JSON from "../../ubiquibot-config-default.json";
+import defaultConfig from "../../ubiquibot-config-default";
```
pavlovcik, line 7: This should still be a const so it should still be capitalized. You can set the object as readonly or as const to enforce that.

ubiquibot-config-default.ts
Wholesomebruh: 0d wasn't [there](https://github.com/ubiquity/ubiquibot/blob/development/ubiquibot-config-default.json#L92) originally, thanks for mentioning. I will update it.

src/utils/private.ts
Wholesomebruh: Got it, any more suggestions?

Wholesomebruh: Should be all caps or "DefaultConfig" ?

Wholesomebruh:
@rndquu , any more suggestions or changes from your side on this?

src/utils/private.ts
pavlovcik: `DEFAULT_CONFIG` to represent a const

ubiquibot-config-default.ts
```diff
@@ -90,5 +91,8 @@
   "enable-access-control": {
     "label": false,
     "organization": true
-  }
+  },
+  "stale-bounty-time":"0d"
 }
+
+export default DefaultConfig;
```
pavlovcik, line 98:
```suggestion
export default DefaultConfig as readonly;
```

I'm not sure which way is correct but you can do `as readonly` or `as const` I'm pretty sure which will actually make this whole object immutable.

pavlovcik (changes requested):

ubiquibot-config-default.ts
Wholesomebruh: Got it, it will be 
```bash
export default DefaultConfig as const
```
Is "DefaultConfig" fine or it should be changed to "DEFAULT_CONFIG"?

0xcodercrane (changes requested):
1/ If we change `json` to `ts` and need to read the config file in pay.ubq.fi, we should publish the repository to the npm registry, add it in package.json in pay.ubq.fi and import it.

2/ Then we should move ubiquibot-config-default.ts into the src folder to make sure it's being built/checked @Wholesomebruh

ubiquibot-config-default.ts
0xcodercrane: DefaultConfig is fine to me

0xcodercrane:
The PR merge should be happening with the relevant changes in pay.ubq.fi right? @rndquu

rndquu:
> The PR merge should be happening with the relevant changes in pay.ubq.fi right? @rndquu

You're free to merge the current PR

There is already a related [issue](https://github.com/ubiquity/pay.ubq.fi/issues/135) for onboarding refactoring

We will deploy a new version of pay.ubq.fi only when https://github.com/ubiquity/pay.ubq.fi/issues/135 is implemented so that nothing is broken

0xcodercrane:
Ok let me handle the merge once the conversations resolved.

ubiquibot-config-default.ts
Wholesomebruh: None of them are working.
On adding "as readonly" it's giving "Type expected" error and 
On adding "as const" , it's showing "const" can be added to string, number, boolean,enum, array and object literals only.
 
Is "DefaultConfig" fine or it should be changed to "DEFAULT_CONFIG"?
@0xcodercrane mentioned "DefaultConfig" is fine and @pavlovcik asked to use "DEFAULT_CONFIG"

0xcodercrane: ok. I couldn't find @pavlovcik 's comment because it has been already resolved.

Lets go with `DEFAULT_CONFIG` and use const.

pavlovcik: Use chatgpt to figure out how to make objects immutable @Wholesomebruh

0xcodercrane (approved):

0xcodercrane (approved):

0xcodercrane:
@Wholesomebruh  bump for the second request: 
> 2/ Then we should move ubiquibot-config-default.ts into the src folder to make sure it's being built/checked

Wholesomebruh:
@0xcodercrane , moved the ubiquity-config-default to src folder.

pavlovcik:
> @0xcodercrane , moved the ubiquity-config-default to src folder.

This can break other systems, including the partner onboarding @rndquu rfc

0xcodercrane:
> > @0xcodercrane , moved the ubiquity-config-default to src folder.
> 
> This can break other systems, including the partner onboarding @rndquu is this all clear?

I don't think so

rndquu:
> > > @0xcodercrane , moved the ubiquity-config-default to src folder.
> > 
> > 
> > This can break other systems, including the partner onboarding @rndquu is this all clear?
> 
> I don't think so

It does break the onboarding but we already have a related [issue](https://github.com/ubiquity/pay.ubq.fi/issues/135) for onboarding refactoring so we won't deploy a new version of pay.ubq.fi unless the https://github.com/ubiquity/pay.ubq.fi/issues/135 is implemented

0xcodercrane:
From my understanding of my questions before merging the PR, I was free to merge right? 

Ideally, we shouldn't merge the pr if it breaks something even if it is partner onboarding.

So I guess the solution here is to wrap up the related issues as soon as possible.

rndquu:
> From my understanding of my questions before merging the PR, I was free to merge right?

yes

> So I guess the solution here is to wrap up the related issues as soon as possible

yes

PR #778, opened 2023-09-17
Blame

### LINKED ISSUE #758: Penalty Enhancement
Sometimes features that were built and paid for break in the future (for any reason.) For the bot to understand what happened, it is expected that the original issue is re-opened. 

It would be epic to see the `diff` from the pull request and compare it to the current working version of the code (`main` branch, configurable.) 

If it has been too long, and any of the original assignee's code was changed from the time it was approved to the current production version (using diff this should be easy) then the blame should not be on the original assignee. 

Instead, those who modified the lines of code should be blamed for breaking the feature. 

In this case, the bot should: 
1. Not apply a penalty to the original assignee and explain why.
2. Leave a comment on the re-opened issue and explain, in detail, which lines of code were changed and by who (tag them)
3. We can let repository maintainers chase down those tagged members to see who will be responsible to fix. 

> @pavlovcik i noticed this is a penalty.. It's been a long time since I worked on this issue

_Originally posted by @seprintour in https://github.com/ubiquity/ubiquibot/issues/260#issuecomment-1715278687_

Keyrxng:
Resolves #758

Quality Assurance: 

- https://github.com/Keyrxng/UbiquityBotTesting/issues/1#issuecomment-1722343964 
- - Scenario: An old issue with multiple PRs. One merged and the other unmerged, the unmerged one is excluded

- https://github.com/Keyrxng/UbiquityBotTesting/issues/47#issuecomment-1722344751
- - Scenario: The same issue which is also the latest merge is reopened.

The diff comparison works by finding the ``merge_commit_sha`` and compares it to ``currentCommit.sha`` 

Looking for opinions going forward from here in regards to identifying who broke the feature of the issue.

We can go direct and say 'who changed the code from pr a to pr b' and that's who is at fault but what if the feature of the issue gets broken by a peripheral change, how do we identify that? (The more I think about it this will likely be in the AI infused version am I right?)

Or do we approach it from a 'this was the last working version' so last man on top carries the burden?

> If any of the original assignee's code was changed from the time it was approved to the current production version (using diff this should be easy to see) then the blame should no longer be on the original assignee.

Your spec sounds a lot like 'pr a to pr b' for me, so my next steps (at least for the first iteration) will be simply looking for any commits that overwrite the implementation of the issue.

I'm thinking if there are multiple contributors then it'll just need to boil down to whoever changed the most code?


Keyrxng:
@pavlovcik Bit busy the past few days but finally got around to it, let me know what your thoughts are when you can

pavlovcik:
Let's do two dot compare instead of three dot. I think it shows every file change. To be honest I'm not 100% clear on the difference but one seems to be more granular. 

I think for peripheral change we can figure an improvement to the feature with a new bounty. 

Not all lines of code changed are equal so that's why I don't want to "blame more" who changed the most lines of code. However in the blame list we could sort by most lines of code changed to least but I don't think we need to emphasize this. As long as they changed any lines from the merged pull request, they are a suspect. 

This is cool to see currently.

src/handlers/comment/handlers/index.ts
```diff
@@ -117,6 +119,119 @@ export const issueCreatedCallback = async (): Promise<void> => {
...
+
+  if (!issue) return;
+  if (!repository) return;
+
+  const allRepoCommits = await context.octokit.repos
+    .listCommits({
+      owner: repository.owner.login,
+      repo: repository.name,
+    })
+    .then((res) => res.data);
```
pavlovcik, line 144: Perhaps we should stick in asynchrony await format?

src/handlers/comment/handlers/index.ts
```diff
@@ -117,6 +119,119 @@ export const issueCreatedCallback = async (): Promise<void> => {
...
+
+  const pullsThatCommitsMatch = await Promise.all(
+    commitsThatMatch.map((commit) =>
+      context.octokit.repos
+        .listPullRequestsAssociatedWithCommit({
+          owner: repository.owner.login,
+          repo: repository.name,
+          commit_sha: commit.sha,
+        })
+        .then((res) => res.data)
```
pavlovcik, line 163: Async await?

src/handlers/processors.ts
```diff
@@ -18,7 +18,7 @@ export const processors: Record<string, Handler> = {
   },
   [GithubEvent.ISSUES_REOPENED]: {
     pre: [nullHandler],
-    action: [issueReopenedCallback],
+    action: [issueReopenedBlameCallback, issueReopenedCallback],
     post: [nullHandler],
```
pavlovcik, line 22: @0xcodercrane Does this make sense to add in post?

0x4007 (commented):
Code and qa looks pretty solid so far

src/handlers/processors.ts
0xcodercrane: Depends on the goal of `issueReopenedBlameCallback`.  In general, the post-handler would be a light one. so I think it might be better to add it in action.

Keyrxng:
> Let's do two dot compare instead of three dot. I think it shows every file change. To be honest I'm not 100% clear on the difference but one seems to be more granular.

``
The three-dot comparison shows the difference between the latest common commit of both branches (merge base) and the most recent version of the topic branch.
``

``
A two-dot diff compares two Git committish references, such as SHAs or OIDs (Object IDs), directly with each other. On GitHub, the Git committish references in a two-dot diff comparison must be pushed to the same repository or its forks.
``

Three dot gets the benefit of the GitHub UI prettifying it for us. In order to get a two dot comparison we'll need to handle formatting the url ourselves.

Currently: <ref>...<ref> and it works

Incoming: Use the compareChanges url with dual dots and insert the refs ourselves, but how should it be formatted?

https://github.com/Keyrxng/UbiquityBotTesting/issues/47#issuecomment-1722344751

src/handlers/comment/handlers/index.ts
```diff
@@ -215,20 +214,87 @@ export const issueReopenedBlameCallback = async (): Promise<void> => {
...
+  | **Blame** | **Count** | **%** |
+  | --- | --- | --- |
+  ${Array.from(new Set(blamers))
+    .filter((blamer) => blamer.count > 0)
+    .sort((a, b) => b.count - a.count)
+    .map((blamer) => {
+      const linesChanged = blamer.count;
+      const totalLinesChanged = blamers.reduce((acc, val) => acc + val.count, 0);
+      const percent = Math.round((linesChanged / totalLinesChanged) * 100);
+      return `| ${blamer.author} | ${blamer.count} | ${percent}% |`;
```
pavlovcik, line 253: We don't need to display the count but the percentage is a good idea. Its a nice abstraction. The only thing that people will pay attention to is comparing to each other who is more likely at fault lol

src/handlers/comment/handlers/index.ts
```diff
@@ -215,20 +214,87 @@ export const issueReopenedBlameCallback = async (): Promise<void> => {
...
+  const advancedBlameTable = `
+  | **Blame** | **Count** | **%** |
+  | --- | --- | --- |
+  ${Array.from(new Set(blamers))
+    .filter((blamer) => blamer.count > 0)
+    .sort((a, b) => b.count - a.count)
+    .map((blamer) => {
+      const linesChanged = blamer.count;
+      const totalLinesChanged = blamers.reduce((acc, val) => acc + val.count, 0);
+      const percent = Math.round((linesChanged / totalLinesChanged) * 100);
```
pavlovcik, line 252: Why not do charactersChanged (excluding whitespace?)

src/handlers/comment/handlers/index.ts
```diff
@@ -215,20 +214,87 @@ export const issueReopenedBlameCallback = async (): Promise<void> => {
...
+    return `\n<code><a href="${url}">${slice}</a></code> - ${fileLen} files changed`;
+  })
+  .join("\n")}
+</details>
+
+<details>
+<summary>Assigned Blame</summary><br/>
+
+The following ${blameQuantifier} may be responsible for breaking this issue:
+  ${advancedBlameTable}
```
pavlovcik, line 289: It could be interesting to also include the reviewers who approved their pull requests. But I don't think its essential. I'll know much better after seeing this code in action.

Keyrxng:
- QA: https://github.com/ubq-testing/bot-ai/issues/1#issuecomment-1763214331

Now showing the reviewers, which PRs they reviewed as well the count being removed from hunters. Included a comparison for all merge commits and left in the example 2 and 3 dot comparisons lmk which of the two is the preference.

Where do I take it from here?

0x4007:
I fear that merged pulls and commits lists could get massive but I can't think of a better solution now. 

Seems pretty solid so far. 

As a heads up we are temporarily pausing merged into `development` until I can get my massive refactor in (I have a draft PR now)

Furthermore we plan to only accept with unit tests so if you want to try and get a head start feel free or you can standby for another week to add when jest is merged in the project

src/handlers/comment/handlers/index.ts
```diff
@@ -155,6 +158,244 @@ export const issueCreatedCallback = async (): Promise<void> => {
...
+  for (const pullNumber of onlyPRsNeeded) {
+    const reviewers = pullReviewsMap[pullNumber];
+    if (reviewers) {
+      for (const reviewer of reviewers) {
+        const blame = blamedReviewers.find((b) => b.author === reviewer);
+        if (blame) {
+          blame.count += 1;
+          blame.pr?.push(pullNumber);
+        } else {
+          blamedReviewers.push({ author: reviewer || "", count: 1, pr: [pullNumber] });
```
0x4007:
```suggestion
          blamedReviewers.push({ author: reviewer || null, count: 1, pr: [pullNumber] });
```

0x4007 (changes requested):
Using null is more expressive than "" to signal that it is intentionally an empty value. 

Besides it's less verbose to check conditional statements as well, given that null is falsely and "" is truthy 

You should apply this everywhere applicable

PR #779, opened 2023-09-18
fix: removed the 10 white space infront of the issue's text

### LINKED ISSUE #772: `ignore-children` comment incentive config
> Yes I think it's also counting words in code blocks and the top quote and italics

Yikes. We should set a special property on every element that indicates whether the parser should recurse or stop.

```yaml
incentives: 
  comments:
    totals:
      word: 0.1
    elements:
      rewards:
        code: 5
        p: 0.1
        a: 5
    ignore-children: code, i, blockquote
      
```

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/648#issuecomment-1720029504_

GaryThisSidee:
Resolves  #772

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "lint": "eslint --ext .ts ./src",
     "start:serverless": "tsx src/adapters/github/github-actions.ts",
     "start:watch": "nodemon --exec 'yarn start'",
-    "start": "probot run ./lib/src/index.js",
+    "start": "probot run ./lib/index.js",
     "prepare": "husky install"
   },
   "dependencies": {


whilefoo (changes requested):
this doesn't seem to implement `ignore-children` like it was stated in the issue

src/handlers/payout/post.ts
```diff
@@ -398,7 +398,7 @@ const calculateRewardValue = (comments: Record<string, string[]>, incentives: In
       const reward = wordReward.mul(value.map((str) => str.trim().split(" ").length).reduce((totalWords, wordCount) => totalWords + wordCount, 0));
       sum = sum.add(reward);
     } else {
-      if (!incentives.comment.elements[key]) {
+      if (!incentives.comment.elements[key] || incentives.comment.ignore_children.includes(key)) {
```
whilefoo, line 401: this is not desired functionality. this will only stop processing that particular item, but we want stop recursing when we encounter that item during parsing the markdown. 
it basically replaces hardcoded variable: https://github.com/ubiquity/ubiquibot/blob/dba8308779aa1ea74e5a98c227a5e92ae3307d9b/src/handlers/payout/post.ts#L18

whilefoo (changes requested):

src/handlers/payout/post.ts
GaryThisSidee: > this is not desired functionality. this will only stop processing that particular item, but we want stop recursing when we encounter that item during parsing the markdown. it basically replaces hardcoded variable:
> 
> https://github.com/ubiquity/ubiquibot/blob/dba8308779aa1ea74e5a98c227a5e92ae3307d9b/src/handlers/payout/post.ts#L18

ohhh my bad

pavlovcik: > it basically replaces hardcoded variable:

Shouldn't we remove the hardcoded variable if we can handle from the config @whilefoo? If so, would you mind creating a new issue for that?

GaryThisSidee: > > it basically replaces hardcoded variable:
> 
> Shouldn't we remove the hardcoded variable if we can handle from the config @whilefoo? If so, would you mind creating a new issue for that?

should i remove it? will replace it with incentive ignore children

pavlovcik: I think it makes more sense to handle in a new issue.

GaryThisSidee: > I think it makes more sense to handle in a new issue if whilefoo thinks its necessary. That way this one can be merged in faster, most likely.

sure

whilefoo: > Shouldn't we remove the hardcoded variable if we can handle from the config @whilefoo? If so, would you mind creating a new issue for that?

This issue is about this hardcoded variable, it stops recursing when it encounters one of the tags in this variable. So we just need to move it the config as `ignore-children`

whilefoo (changes requested):
https://github.com/ubiquity/ubiquibot/blob/628c2221a2d6ee62583b702d711ba57393415b6c/src/helpers/comment.ts#L10
you should modify this function to still parse the element but not its children. For example currently if it's `ignore-children: code` it won't include `code` item and its children, but we still want to include `code`. Don't forget to resolve conflicts

@pavlovcik do we want to set some sensible default like `ignore-children: blockquote, code, pre, a`

pavlovcik:
Sensible defaults sounds like a good idea!

GaryThisSidee:
> https://github.com/ubiquity/ubiquibot/blob/628c2221a2d6ee62583b702d711ba57393415b6c/src/helpers/comment.ts#L10
> 
> you should modify this function to still parse the element but not its children. For example currently if it's `ignore-children: code` it won't include `code` item and its children, but we still want to include `code`. Don't forget to resolve conflicts
> 
> @pavlovcik do we want to set some sensible default like `ignore-children: blockquote, code, pre, a`

sure

src/helpers/comment.ts
```diff
@@ -9,6 +9,7 @@ type Node = {
 
 const traverse = (result: Record<string, string[]>, node: Node, itemsToExclude: string[]): Record<string, string[]> => {
   if (itemsToExclude.includes(node.nodeName)) {
+    result[node.nodeName].push(node.nodeName);
```
whilefoo, line 12: this will cause an error if nodeName is not in the object yet

https://github.com/ubiquity/ubiquibot/blob/02deb96a2021c6ef6841294d44984ed3b8792854/src/helpers/comment.ts#L21

I recommend to just adding a condition here like `!itemsToExclude.includes(node.nodeName)`

whilefoo (changes requested):

GaryThisSidee:
@whilefoo going to open a new pull req and push the changes in one commit its kinda messy rn

PR #781, opened 2023-09-18
feat: selectivly enable incentives

### LINKED ISSUE #780: Selectively enabling comment incentives per task
This is a great example of the assignee doing research and it would be elegant for an `admin` or `billing_manager` to be able to issue a slash command to force comment incentives to be calculated for a contributor no matter what their recognized role to be after the issue is closed out. 

For example `/comment-incentives @keyrxng true`

So even if they are the assignee when the issue is closed as complete, we have the ability to make an exception on this specific issue for them to claim their comment incentives. 

This command should work in any order, like our other commands. 

I would like to hear suggestions on a more concise command name because that current one seems a bit long.

> So we do not want to shackling hunters but enforcing a template or structure I get that and given the observations in the table below I want to make the following points:

_Originally posted by @Keyrxng in https://github.com/ubiquity/ubiquibot/issues/777#issuecomment-1723247966_

BeanieMen:
Resolves #780

Quality Assurance:


me505:
This uses a helper to boil down the comments to a list of users and if to enable or not, after when generating incentives it only generates for the users in the list if there are any

me505:
also if it is enabled

me505:
hmm

me505:
i dont get wheres the error its so lucrative

src/helpers/issue.ts
```diff
@@ -95,6 +95,22 @@ export const listAllIssuesForRepo = async (state: "open" | "closed" | "all" = "o
   return issuesArr;
 };
 
+export const getIncentivizedUsers = async (issue_number: number) => {
+  const comments = await getAllIssueComments(issue_number);
+  const incentiveComment = comments.filter((comment) => comment.body.startsWith("/comment-incentives"));
+  const parts = incentiveComment[0].body.split("");
```
whilefoo, line 101: what about if there are multiple `/comment-incentives` comments?

src/helpers/issue.ts
```diff
@@ -95,6 +95,22 @@ export const listAllIssuesForRepo = async (state: "open" | "closed" | "all" = "o
...
+  parts.shift();
+  const users: { enable: boolean; users: string[] } = { enable: false, users: [] };
+  for (const part of parts) {
+    if (part.startsWith("@")) {
+      users.users.push(part.substring(1));
+    } else if (part == "true") {
```
whilefoo, line 107:
```suggestion
    } else if (part === "true") {
```
use `===` whenever possible

also check if part is `false` because it could be anything and we should throw error if it's not `true`, `false` or starts with @.

if `true` or `false` is not provided, then what should be default? @pavlovcik

whilefoo (changes requested):

src/handlers/payout/post.ts
```diff
@@ -98,31 +98,35 @@ export const incentivizeComments = async () => {
...
+        const { payoutUrl } = await generatePermit2Signature(account, amountInETH, issue.node_id, commentsByUser.id, "ISSUE_COMMENTER");
+        comment = `${comment}### [ **${user}: [ CLAIM ${amountInETH} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n`;
+        reward[user] = payoutUrl;
+      } else {
+        fallbackReward[user] = amountInETH;
+      }
     } else {
-      fallbackReward[user] = amountInETH;
+      continue;
     }
```
wannacfuture, line 130: Is this necessary? We can just remove `else`.

src/helpers/issue.ts
wannacfuture, line 101:
```suggestion
  const parts = incentiveComment[0].body.split(" ");
```

me505: what do you suggest?

src/helpers/issue.ts
whilefoo: @me505 you need to explicitly check like `if (part === "false")`

src/helpers/issue.ts
whilefoo: I would iterate from oldest to newest and update accordingly.
For example:
```
/comment-incentives false @test @test2
/comment-incentives true @test @test3
```
then the final configuration should be:
```
@test: true
@test3: true
@test2: false
```

src/helpers/issue.ts
pavlovcik: > if true or false is not provided, then what should be default? @pavlovcik

It should throw an error and ignore the input from the user (i.e. no effect.)

me505:
idk how ubiquibot-config-default.ts got changed

0x4007 (commented):
You're using a different prettier setting because of these extra commas, can you fix this?

src/helpers/issue.ts
```diff
@@ -95,6 +95,32 @@ export const listAllIssuesForRepo = async (state: "open" | "closed" | "all" = "o
...
 };
 
+export const getIncentivizedUsers = async (issue_number: number) => {
+  const comments = await getAllIssueComments(issue_number);
+  const incentiveComments = comments.filter((comment) => comment.body.startsWith("/comment-incentives"));
+  let users: { [key: string]: boolean } = {};
+  for (const incentiveComment of incentiveComments) {
+    const parts = incentiveComment.body.split(" ");
+    parts.shift();
+    if (parts.pop() == "true") {
```
whilefoo, line 105: this relies on command order. just use a for loop and check for `true`, `false`, or starts with `@` (there shouldn't more than one `true` or `false`). you can look at this [example](https://github.com/ubiquity-whilefoo/ubiquibot/blob/3056bc91e062ab9e1b161d34d2da1418a3771a30/src/handlers/comment/handlers/allow.ts#L27)

src/handlers/comment/handlers/comment-incentives.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  logger.info(`Received '/comment-incentive' command from user: ${sender}`);
+
+  if (!payload.issue) {
+    logger.info(`Skipping '/comment-incentive' because of no issue instance`);
+    return `Skipping '/comment-incentive' because of no issue instance`;
+  }
+  const parts = body.split(" ");
+  parts.shift();
+  if (parts.pop() !== "true" && parts.pop() !== "false") {
```
whilefoo, line 18: you can create a helper function that parses the command so you can use it here and in `getIncentivizedUsers` to avoid repeating code

whilefoo (changes requested):

BeanieMen:
gonna have to remake :skull:

PR #789, opened 2023-09-20
Hotfix: check org and repo for linked issue

### LINKED ISSUE #666: Automatic Assign on Opened Pull Request (Assigned to Unrelated Task!)
I was working on https://github.com/ubiquity/ubiquibot/issues/595 and opened up a few pull requests because I need to change several repositories' UbiquiBot configs. After opening up my third pull request https://github.com/ubiquity/ubiquity-dollar/pull/751 I was assigned to an unrelated issue: https://github.com/ubiquity/ubiquity-dollar/pull/595

I assume this has something to do with the fact that the issue number (595) is the same across both repositories. Perhaps the feature did not accomodate for cross repository issue tracking. Please diagnose and fix. 

_Originally posted by @pavlovcik in https://github.com/ubiquibot/production/issues/34_

wannacfuture:
Resolves #666


src/helpers/parser.ts
```diff
@@ -29,8 +29,26 @@ export const gitLinkedIssueParser = async ({ owner, repo, pull_number }: GitPars
...
+      // check if array size is at least 4
+      if (parts.length < 4) continue;
+
+      // extract the organization name and repo name from the link:(e.g. "
+      const issueOrganization = parts[parts.length - 4];
+      const issueRepository = parts[parts.length - 3];
+
+      if (`${issueOrganization}/${issueRepository}` !== `${owner}/${repo}`) continue;
+
+      return issueUrl;
```
whilefoo, line 48: maybe it would be better the function returns all linked issues instead of just one?

wannacfuture: So.. one pr has several linked issues.

In this case, should we assign the pr opener to that all linked issues?

@pavlovcik , @whilefoo

cuddle-dev: Agree with @whilefoo's comment. As a helper function, that would be better to return all the linked issues instead of the first-come issue. 

> In this case, should we assign the pr opener to that all linked issues?

If the linked issues are bounties, I would say yes but I don't think we have such a case yet. so as of now, no need to worry about that. Just update the return type string[] and return all the linked issue urls.

wannacfuture: The reason why I'm worrying about it is other function(automatically assigning pr author to issue) is using this issueParser.

whilefoo: I'm thinking about a scenario where a PR resolves two issues so I guess it would make sense to assign the PR author to both linked issues.

wannacfuture: We need to think about some restrictions of max assign issue counts for the individual hunter

whilefoo: Is it possible to check how many X assigns does a hunter have available and only assign them to X number of linked issues

wannacfuture: But we need to think how to prioritize them. which issues to assign. 

@0xcodercrane @whilefoo

whilefoo: It probably doesn't matter that much, if it's an issue we can fix it later

wannacfuture:
QA: https://github.com/wannacfuture/Battleship/pull/36 (see linked issues if they are assigned correctly)

src/handlers/assign/auto.ts
```diff
@@ -18,43 +18,45 @@ export const checkPullRequests = async () => {
...
 
-    // Check if the pull request opener is assigned to the issue
-    const opener = pull.user.login;
+      // Check if the pull request opener is assigned to the issue
+      const opener = pull.user.login;
 
-    const issue = await getIssueByNumber(context, +linkedIssueNumber);
-    if (!issue?.assignees) continue;
+      const issue = await getIssueByNumber(context, +linkedIssueNumber);
+      if (!issue?.assignees) continue;
```
cuddle-dev, line 47: If there's no assignee, we don't need to assign the opener of the linked PR to the issue?

wannacfuture: It doesn't mean there is no assinee. 
It just wants to make sure if `assinees` field exists on the issue not `if(issue?.assignees.length === 0)`

cuddle-dev: Sounds like the same thing.  No assignee means the `assignees` field doesn't exist, right?

wannacfuture: No assignee means assignees field is empty array, not undefined.

whilefoo: this if statement will only execute if `assignees` is undefined or null, but not if its empty array. github should always send empty array if there's no assignees but if they don't send the field, this could possibly be a bug.

wannacfuture: Exactly. we need to go on,  to assign the author of the linked PR to the issue if assignees is empty array.
If it is undefined or null, it should skip the loop.

PR #795, opened 2023-09-22
Staging to production

0xcodercrane:
Resolves #

Quality Assurance:


README.md
```diff
@@ -224,7 +224,7 @@ Bounty bot is built using the [probot](https://probot.github.io/) framework so i
 ├── <a href="https://github.com/ubiquity/ubiquibot/tree/development/src/utils">utils</a> A set of utility functions
 </pre>
 
-## Default Config Notes (`ubiquibot-config-default.json`)
+## Default Config Notes (`ubiquibot-config-default.ts`)
 
 We can't use a `jsonc` file due to limitations with Netlify. Here is a snippet of some values with notes next to them.
```
pavlovcik, line 229: This is no longer necessary

src/configs/ubiquibot-config-default.ts
```diff
@@ -0,0 +1,102 @@
...
+    {
+      name: "Priority: 3 (High)",
+    },
+    {
+      name: "Priority: 4 (Urgent)",
+    },
+    {
+      name: "Priority: 5 (Emergency)",
+    },
+  ],
```
pavlovcik, line 49: I see that there are no `value` or `weight` properties which currently exists in our organization wide config. Do you mind taking a look at the organization wide config, and removing the extraneous properties?

src/handlers/comment/commands.ts
```diff
@@ -6,6 +6,7 @@ export enum IssueCommentCommands {
   PAYOUT = "/payout", // request permit payout
   MULTIPLIER = "/multiplier", // set bounty multiplier (for treasury)
   QUERY = "/query",
+  ASK = "/ask", // ask GPT a question
```
pavlovcik, line 9: Did `/review` make it in?

src/handlers/comment/handlers/ask.ts
```diff
@@ -0,0 +1,123 @@
...
+    // let chatgpt deduce what is the most relevant context
+    const gptDecidedContext = await decideContextGPT(chatHistory, streamlined, linkedPRStreamlined, linkedIssueStreamlined);
+
+    if (linkedIssueStreamlined.length == 0 && linkedPRStreamlined.length == 0) {
+      // No external context to add
+      chatHistory.push(
+        {
+          role: "system",
+          content: sysMsg,
+          name: "UbiquityAI",
```
pavlovcik, line 83: These should've been called UbiquiBot

src/handlers/comment/handlers/table.ts
```diff
@@ -4,32 +4,35 @@ export const tableComment = ({
...
+<td>${deadline}</td>
+</tr>
+<tr>
+<td>Registered Wallet</td>
+<td>${wallet}</td>
+</tr>
+${multiplier ? `<tr><td>Payment Multiplier</td><td>${multiplier}</td></tr>` : ``}
+${reason ? `<tr><td>Multiplier Reason</td><td>${reason}</td></tr>` : ``}
+${bounty ? `<tr><td>Total Bounty</td><td>${bounty}</td></tr>` : ``}
+</table></code>`;
```
pavlovcik, line 37: Seems that this could've benefited from white space formatting

src/handlers/payout/post.ts
```diff
@@ -84,160 +69,234 @@ export const incentivizeComments = async () => {
...
+    const rewardValue = calculateRewardValue(commentsByNode, calculateIncentives.incentives);
     if (rewardValue.equals(0)) {
       logger.info(`Skipping to generate a permit url because the reward value is 0. user: ${user}`);
       continue;
     }
     logger.debug(`Comment parsed for the user: ${user}. comments: ${JSON.stringify(commentsByNode)}, sum: ${rewardValue}`);
     const account = await getWalletAddress(user);
-    const amountInETH = rewardValue.mul(baseMultiplier);
-    if (amountInETH.gt(paymentPermitMaxPrice)) {
+    const priceInEth = rewardValue.mul(calculateIncentives.baseMultiplier);
```
pavlovcik, line 90: Perhaps we should rename this to `priceInBigNumber`

src/handlers/payout/post.ts
```diff
@@ -84,160 +69,234 @@ export const incentivizeComments = async () => {
...
+    return { error: `Skipping to generate a permit url because issue description is empty. description: ${description}` };
   }
   logger.info(`Getting the issue description done. description: ${description}`);
-  const creator = issue.user;
-  if (creator.type === UserType.Bot || creator.login === issue.assignee) {
+  const creator = incentivesCalculation.issue.user;
+  if (creator.type === UserType.Bot || creator.login === incentivesCalculation.issue.assignee) {
     logger.info("Issue creator assigneed himself or Bot created this issue.");
-    return;
+    return { error: "Issue creator assigneed himself or Bot created this issue." };
```
pavlovcik, line 140: Shouldn't these be two separate errors they seem unrelated

src/handlers/payout/post.ts
```diff
@@ -84,160 +69,234 @@ export const incentivizeComments = async () => {
...
-    incentives,
-    tokenSymbol,
-    issue.node_id,
-    paymentPermitMaxPrice
+    incentivesCalculation.issueCreatorMultiplier,
+    incentivesCalculation.incentives,
+    incentivesCalculation.paymentPermitMaxPrice
+  );
+
+  if (!result || !result.account || !result.amountInETH) {
```
pavlovcik, line 151:
```suggestion
  if (!result || !result.account || !result.amountInBigNumber) {
```

src/handlers/processors.ts
```diff
@@ -54,7 +53,7 @@ export const processors: Record<string, Handler> = {
   [GithubEvent.ISSUES_CLOSED]: {
     pre: [nullHandler],
     action: [issueClosedCallback],
-    post: [incentivizeCreatorComment, incentivizeComments],
+    post: [nullHandler],
```
pavlovcik, line 56: This seems unexpected. What's going on here?

src/helpers/gpt.ts
```diff
@@ -0,0 +1,180 @@
...
+export const sysMsg = `You are the UbiquityAI, designed to provide accurate technical answers. \n
+Whenever appropriate, format your response using GitHub Flavored Markdown. Utilize tables, lists, and code blocks for clear and organized answers. \n
+Do not make up answers. If you are unsure, say so. \n
+Original Context exists only to provide you with additional information to the current question, use it to formulate answers. \n
+Infer the context of the question from the Original Context using your best judgement. \n
+All replies MUST end with "\n\n <!--- { 'UbiquityAI': 'answer' } ---> ".\n
+`;
+
+export const gptContextTemplate = `
+You are the UbiquityAI, designed to review and analyze pull requests.
```
pavlovcik, line 17: I think the command is not registered. I couldn't find it earlier.

src/helpers/gpt.ts
```diff
@@ -0,0 +1,180 @@
+import { getBotConfig, getBotContext, getLogger } from "../bindings";
+import { Payload, StreamlinedComment, UserType } from "../types";
+import { getAllIssueComments, getAllLinkedIssuesAndPullsInBody } from "../helpers";
+import OpenAI from "openai";
+import { CreateChatCompletionRequestMessage } from "openai/resources/chat";
+import { ErrorDiff } from "../utils/helpers";
+
+export const sysMsg = `You are the UbiquityAI, designed to provide accurate technical answers. \n
```
pavlovcik, line 8: UbiquiBot

src/helpers/issue.ts
```diff
@@ -657,9 +682,115 @@ export const getAvailableOpenedPullRequests = async (username: string) => {
...
-    if (reviews.length === 0 && (new Date().getTime() - new Date(pr.created_at).getTime()) / (1000 * 60 * 60) >= botConfig.unassign.timeRangeForMaxIssue) {
+    if (reviews.length === 0 && (new Date().getTime() - new Date(pr.created_at).getTime()) / (1000 * 60 * 60) >= timeRangeForMaxIssue) {
       result.push(pr);
     }
   }
   return result;
 };
+
+// Strips out all links from the body of an issue or pull request and fetches the conversational context from each linked issue or pull request
+export const getAllLinkedIssuesAndPullsInBody = async (issueNumber: number) => {
```
pavlovcik, line 693: I think we have a redundant function, something like `getAllLinkedPulls` I saw in the beginning of this review

src/types/config.ts
```diff
@@ -113,7 +137,83 @@ export const BotConfigSchema = Type.Object({
...
+  tokenUsage: Type.Object({
+    output: Type.Optional(Type.Number()),
+    input: Type.Optional(Type.Number()),
+    total: Type.Optional(Type.Number()),
+  }),
+});
+
+export type GPTResponse = Static<typeof GPTResponseSchema>;
+
+export const WideConfigSchema = Type.Object(
```
pavlovcik, line 164: I never understood this "wide" prefix. I even tried removing it in my unmerged pull request from last month.

src/helpers/gpt.ts
pavlovcik: @Keyrxng rfc

pavlovcik:
@ubiquity/software-development I left a review RFC

Assuming that all of my opinions are agreeable @whilefoo would you be able to implement my requests?

src/configs/ubiquibot-config-default.ts
whilefoo: I don't see `value` or `weight` properties in [org config](https://github.com/ubiquity/ubiquibot-config/blob/development/.github/ubiquibot-config.yml)

src/handlers/comment/commands.ts
whilefoo: no, not merged yet

src/handlers/processors.ts
whilefoo: all incentives are now included in `issueClosedCallback`

src/types/config.ts
whilefoo: Yeah me neither

src/handlers/payout/post.ts
whilefoo: I think we don't need to include `BigNumber` in the name because we have Typescript. `ETH` is wrong because it can technically be any token. maybe `rewardTokens`?

src/handlers/payout/post.ts
whilefoo: Is issue creator not eligible for rewards if they are also the assignee? That doesn't make sense to me.

src/helpers/gpt.ts
Keyrxng: My mistake fellas, forgot to replace it following shuffling the order of the contexts. In gpt.ts inside /review, the template is supplied first before anything else [#L202](https://github.com/Keyrxng/ubiquibot/blob/f1400cc00b500ac260d6c53c6fb60c6c858e52d7/src/helpers/gpt.ts#L202). It's not used in /review as we are only passing issue/pull body as context now but it should be passed in to chatHistory for ask.

Without it GPT responds in a story telling way rather than a json format but achieves the same thing, but JSON does provide more data for the dependant call.

Let me know what happens next, whether I'm to fix it or what bit wet behind the ears with this

src/types/config.ts
pavlovcik: This should be refactored and more expressive. Refer to my open pull request.

src/helpers/gpt.ts
pavlovcik: You can file a new issue and start working on it to get paid. If its too small of a task it might be more worth to just open the pull request for it without filing an issue. Either way is fine. 

I should be able to price it before you get it done.

Keyrxng: No no new issue or anything needed, it's just adding one convo object into chatHistory but I appreciate that. I'll have time after 6pm tonight to catch it, don't trust doing it on my phone lmao too many fat thumbs waiting to happen

src/handlers/comment/handlers/ask.ts
```diff
@@ -0,0 +1,123 @@
...
+  const payload = context.payload as Payload;
+  const sender = payload.sender.login;
+  const issue = payload.issue;
+
+  if (!body) {
+    return `Please ask a question`;
+  }
+
+  if (!issue) {
+    return `This command can only be used on issues`;
```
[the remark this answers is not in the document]
pavlovcik: If so the error should explicitly state " and pull requests"

src/handlers/payout/post.ts
rndquu: Here `ETH` was meant to distinguish price in wei and price in ETH, not to mention a particular payment token

whilefoo: I think we should think about changing to doing everything in wei because we will support multiple tokens and they can have different decimals so it doesn't make sense to do in ETH (18 decimals)

src/helpers/issue.ts
whilefoo: I'm not sure what's the context of this function, but this functions retrieves all URLs to lead to PRs and issues in the body of the issue which is different from `gitLinkedPrParser` which retrieves only linked PRs on the specific issue.

Issue #812, opened 2023-09-23
Database Re-Architecture


pavlovcik:
# Tables

- Every table should start with the following columns: `id`, `created`, `updated` and have foreign keys in the last columns.
- We should completely stop using GitHub user/repo/org names as they have led to issues with renames etc. Instead we should ONLY use GitHub IDs (integers)
- I am referencing the production database, so some tables like `multiplier` might not exist in other databases. If so a comment should be left so that we handle the deletion.

## Access

The columns should be renamed and adhere to the above rules.

```
id, created, updated, user_id, organization_id, repository_id, multiplier, priority, time, price.
```

## Issues

Should be deleted. No idea why this exists in the first place. @0xcodercrane rfc

This seems to be constantly updated so its logic must be disconnected from the bot side. 

## Logs

Will make an exception since they are ephemeral and connected to multiple systems. 

## Multiplier

Delete.

## Penalty

- https://github.com/ubiquity/ubiquibot/issues/809

## Permits

- https://github.com/ubiquity/ubiquibot/issues/811

## Users

Seems nice to have even if we have no immediate plans for the data. We will leave it unchanged.

## Wallets

id, created, updated, user_id, wallet (char42)

cuddle-dev:
> Should be deleted. No idea why this exists in the first place. @0xcodercrane rfc

Originally created it for data analytics of issues. but looks like no need to use it for even weekly reports. agree to delete.
@molecula451

rndquu:
> /start

I don't think the current issue is relevant/available because (as far as I understand) it's going to be solved as a part of https://github.com/ubiquity/ubiquibot/pull/644

molecula451:
i was thinking the same after seeing #644 linked to this issue but i didnt see relevant/related commits at #644

molecula451:
i will finish the PR anyway you guys let me know if something come up with #644

molecula451:
@pavlovcik can you confirm us?

Also can you confirm this?

![multi](https://github.com/ubiquity/ubiquibot/assets/41552663/7a6fd1a6-a085-4e38-8025-2407063064e6)

on "Access" table you are mentioning "multiplier" but you also mention Multiplier to be delete

0x4007:
I consolidated multiplier details in the access table. Perhaps we should rename the access table but it holds the miscellaneous user information. 

I guess I don't get "activity credit" cause I'm active on a branch off of my linked pull request right now, rebuilding the comment incentives code.

[...]

PR #815, opened 2023-09-23
chore: debug

Keyrxng:
something aint right with config

Resolves #

Quality Assurance: 
- https://github.com/Keyrxng/ubiquibot/issues/4#issuecomment-1732367813

### DIFF
--- a/src/configs/ubiquibot-config-default.ts
+++ b/src/configs/ubiquibot-config-default.ts
@@ -54,27 +54,27 @@ export const DefaultConfig: MergedConfig = {
     },
     {
       name: "stop",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "wallet",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "payout",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "multiplier",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "query",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "ask",
-      enabled: false,
+      enabled: true,
     },
     {
       name: "allow",
@@ -82,7 +82,7 @@ export const DefaultConfig: MergedConfig = {
     },
     {
       name: "autopay",
-      enabled: false,
+      enabled: true,
     },
   ],
   incentives: {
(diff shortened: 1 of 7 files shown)


Keyrxng:
@pavlovcik @0xcodercrane

Keyrxng:
- https://github.com/ubiquity/ubiquibot/issues/812#issuecomment-1732368338

Just ran /start and got a max active on this repo, so idk what it is I've got wrong here as I've pulled fresh from development to debug /ask.

Keyrxng:
idk where I'm going wrong.

I do know that:
- I have no repo config or org config

But the bot should be able to run on it's own default config if those other two fail surely?

pavlovcik:
> But the bot should be able to run on it's own default config if those other two fail surely?

It's supposed to be able to do that yes

Keyrxng:
- https://github.com/Keyrxng/UbiquityBotTesting/issues/58#issuecomment-1732373425

``/help``
```js
export const generateHelpMenu = () => {
  let helpMenu = "### Available commands\n```";
  const commands = userCommands();
  commands.map((command) => {
```

/ask not being displayed in the help table despite looping through all commands despite being in ``commands``:

> "  ASK = "/ask", // ask GPT a question"

pavlovcik:
Maybe your code is out of date. Pull upstream

Keyrxng:
I'm up to date otherwise there would be a notification on this pr to update to current commit and my branch 100% up to date just pulled earlier and double checked right there... wtf

Keyrxng:
Not sure what to do here.

I can't refork the repo I'd need to delete first and lose my current work.

I'm 100% up to date with the current dev branch.

I have to figure out setting up an organization to get around the default config issue atm to be able to debug the actual bot.

If your help command is showing ask, and your getting rate limited with tokens then it's registering as a command.

I noticed during my trials that if I spammed an /ask on two repos before the first had finalized one or both was likely to fail.

So it may be a good issue to have raised to create a queue for gpt requests (I'd say to the same api key as that's probs the blocker but likely there will only ever be one api in use for folks)

src/configs/ubiquibot-config-default.ts
pavlovcik, line 1: These need to be off by default

0x4007 (changes requested):

pavlovcik:
I couldn't find logs related to it that's why I filed the issue but I could add you to the Supabase project if you dm me your Supabase account email

Issue #787, opened 2023-09-20
Database: `creditId`


pavlovcik:
- Related https://github.com/ubiquity/ubiquibot/issues/767
- We should have a table for "credit IDs" that can account for credits and debits.

### Credits

These are payments to contributors, and should store all the permit details. 

I'll make the schema later.

### Debits

These are deductions made to assignees for when the issue is re-opened. 

I'll make the schema later.

pavlovcik:
Debating on this but still thinking about when we eventually make a command to allow repository maintainers to override the contributor's XP: we will need to create a new column in the `settlements` table. `xpId` or something. It will be a foreign key to an "XP Modifiers" table. There we can store all the details around that modification. For example: `id`, `created time`, `updated time`, `amount`.

We can exclude modifiers from the global leaderboard calculation (globalXP) so that its more difficult to exploit (i.e. collusion between an org and a contributor).  We could even make a whistleblower incentive to make sure that people police each other for exploiting the XP system possibly. 

Since it would be available in the settlements table, that means the `penalties` parent table can associate the `user`, `orgId`, `repoId`, `issueId` and `commentURL` to see exactly what happened.

seprintour:
@pavlovcik There was already a Permit table, guess we should use that instead..

It has all of the fields here and more

pavlovcik:
> @pavlovcik There was already a Permit table, guess we should use that instead..
> 
> 
> 
> It has all of the fields here and more

Thanks for highlighting this. The only concern I have with the current implementation is that it has an issue ID associated with every permit, but I want to allow the flexibility for us to issue permits without necessarily associating them to issue IDs. 

I believe that my specification allows for more flexibility, but I also do not claim to be a database architecture expert.

seprintour:
> I also just realized that ChatGPT jumbled up some of the rows. It left some permit specific data inside of the credits table.

Yea, `network`, `signature` and `owner` is permit data.

Previous permit table contains repository data and issue id to pinpoint the exact origin of the permit. Are you saying we don't need that anymore.

Because none of the tables here refers to a repository (origin) of the payment

seprintour:
Rfc

pavlovcik:
The parent table does. https://github.com/ubiquity/ubiquibot/issues/767

seprintour:
> The parent table does. #767

Okay then, this issue entail just creating the tables, moving permit data to the new table. Is creating functions to add these data to DB included? @pavlovcik

[...]

Issue #620, opened 2023-08-15
Bug: Creating or modifying wallet address is currently broken and does not have proper error handling


gitcoindev:
### Describe the background or context

I spent over three hours trying to verify my PR and get 'QA' and I stumbled against a serious bug.

There are two problems: 

1. Inserting or modifying any database row with type defined as a fixed length char is currently broken and allows only 1 character. I verified this with supabase dashboard. The root cause is https://github.com/PostgREST/postgrest/issues/2861 which was fixed but the fix is not available yet in the workspace.

2. Error handling is supabase adapter is broken and it must be fixed e.g.

https://github.com/ubiquity/ubiquibot/blob/development/src/adapters/supabase/helpers/client.ts#L165
https://github.com/ubiquity/ubiquibot/blob/development/src/adapters/supabase/helpers/client.ts#L170
https://github.com/ubiquity/ubiquibot/blob/development/src/adapters/supabase/helpers/client.ts#L178
https://github.com/ubiquity/ubiquibot/blob/development/src/adapters/supabase/helpers/client.ts#L193

Modifying db entries return data or error, but nothing is done when error happens. The functions happily return with just logger traces and the bot claims that everything went well, see https://github.com/gitcoindev/ubiquibot/issues/6#issuecomment-1678456596 and bot comments below. Only when I locally modified the source to insert "1" as wallet address, it was truly successful.

### Describe the solution

1. Use variable char in DB, even for address and validate variables length where needed:

https://github.com/PostgREST/postgrest/issues/1586#issuecomment-696821113

2. Implement a proper error handling in supabase adapter.

whilefoo:
This a known issue. Solution for `2.` will be resolved by #545

gitcoindev:
Good to hear that @whilefoo

whilefoo:
I will make a PR that solves both issues because I don't know how long #545 will take to merge

pavlovcik:
> I will make a PR that solves both issues because I don't know how long #545 will take to merge

Normally we prefer to have unique pull requests for unique issues. This is for auditing reasons. You should try and do a unique pull request for this issue by cherry picking the changes from your other pull request. It's fairly straightforward to do if you use a tool like gitgraph for vscode once the code is already done.

PR #817, opened 2023-09-23
chore: create permit table and move data

### LINKED ISSUE #811: Permits Table Schema
## **Permits**

Holds the permit details associated with some credits.

| Field                  | Type     | Description                                      | Example Value                                                                         |
|------------------------|----------|--------------------------------------------------|---------------------------------------------------------------------------------------|
| `id`                   | int8  | Unique identifier for the permit entry.          | 101                                                                                   |
| `created`  | timestamptz | Timestamp when the permit was created.              | 2023-09-15T11:20:23.004Z     |
| `updated`  | timestamptz | Timestamp when the permit was last updated.         | 2023-09-15T11:22:40.050Z     |
| `network`  | smallserial  | Network identifier.                                | 100                          |
| `token`                | CHAR(42)   | Token associated with the permit.                | 0xe92E133E0b41519A2Ce9Dd4D7945Fa863463a97e                                              |
| `amount`     | int8   | Amount associated with the token for the permit. | 150000000000000000000                                                                  | 
| `nonce`                | int8   | Nonce associated with the permit.                | 8802044552...                                                                          |
| `deadline`             | int8   | Deadline associated with the permit.             | 115792089237316195423570985008687907853269984665640564039457584007913129640000        |
| `beneficiary`   | CHAR(42)   | Destination address for the permit's transfer.  | 0x4008FE2190c8F3E18098aeB4A40bb8eC149a342d                                              |
| `owner`    | CHAR(42)   | Owner of the credit.                               | 0xf88da4592C791213e52840d120E7E0B38B818bD2 |
| `signature`| CHAR(132)   | Signature associated with the credit.              | 0xecb65ed54b2ee7a0ffad5aff8314eef684f80dcb78c18b4ea9f588e083a0fa114546d466befa23cb81d040957615b5b1ad08966b64c5778802e5e115cdcd39241b                |

seprintour:
Resolves #811

Quality Assurance: 

Screenshot 2023-09-23 at 7 27 19 PM

Screenshot 2023-09-23 at 7 25 21 PM

More QA: https://github.com/Seprintour-Test/test/issues/41#issuecomment-1732585892
Screenshot 2023-09-24 at 3 30 56 PM

I also changed the datatype from `char` to `varchar` to fix this error https://github.com/Seprintour-Test/test/issues/41#issuecomment-1732579005 @pavlovcik


seprintour:
@pavlovcik PR attached

supabase/migrations/20230923153903_new_permit.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE IF NOT EXISTS
+  new_permits (
+    id serial primary key,
+    created_at timestamptz not null,
+    updated_at timestamptz not null,
+    network smallserial not null,
+    token char(42) not null,
+    amount numeric not null,
+    nonce numeric not null,
```
pavlovcik, line 9: /ask is numeric bigint in supabase

pavlovcik: I'm skeptical that this is the optimal type. Are you sure?

seprintour: Yes, I tried all of the other datatype..

bigint in supabase is int8, none of them could take that big a number.. It's either numeric or string

pavlovcik:
> The bot is expected to be stable after these changes are implemented which implies that all of the functions that interact with the database should also be refactored to accomodate this new architecture.

From #787

Shouldn't you refactor functions that interact with these tables?

seprintour:
> From #787
> 
> Shouldn't you refactor functions that interact with these tables?

Yea, I would do that now..

seprintour:
> Shouldn't you refactor functions that interact with these tables?

check it out now, i ran into some more issues and fixed.. check description

seprintour:
Ping @whilefoo

supabase/migrations/20230923153903_new_permit.sql
```diff
@@ -0,0 +1,48 @@
...
+    beneficiary varchar(42) not null,
+    owner varchar(42) not null,
+    signature varchar(132) not null
+  );
+
+INSERT INTO
+  new_permits (
+    id,
+    created_at,
+    updated_at,
```
0x4007:
```suggestion
    created,
    updated,
```

0x4007: At seems implied. I think my suggestion is more clean

src/helpers/permit.ts
```diff
@@ -11,11 +11,8 @@ const PERMIT2_ADDRESS = "0x000000000022D473030F116dDEE9F6B43aC78BA3"; // same on
 export type Permit = {
   id: number;
   createdAt: Date;
-  organizationId: number | null;
-  repositoryId: number;
-  issueId: number;
+  updatedAt: Date;
   networkId: number;
-  bountyHunterId: number;
   bountyHunterAddress: string;
```
0x4007, line 16: Not sure if it's appropriate to handle these in this pull request, but we should rename every instance of "bounty hunter" into "contributor" or "assignee" to be specific with our terminology

0x4007 (commented):
Code looks good just have some rename requests

src/helpers/permit.ts
seprintour: > Not sure if it's appropriate to handle these in this pull request, but we should rename every instance of "bounty hunter" into "contributor" or "assignee" to be specific with our terminology

I think that is a lot of change for this issue

supabase/migrations/20230923153903_new_permit.sql
seprintour: > At seems implied. I think my suggestion is more clean

I just added the `_at` because that is what all the tables use.. so i wanted to be consistent

0x4007: The new tables in all the specs I removed those suffixes

Important to be consistent with spec because other contributors are likely to follow it precisely

seprintour: I've remove all _at suffix

seprintour:
> Code looks good just have some rename requests

Fixed issues

seprintour:
Ping @pavlovcik @whilefoo

seprintour:
Ping @pavlovcik @whilefoo

whilefoo (changes requested):
I'd rather not use char(x) because we've had some problems in the past related to that because either Supabase or PostgREST library wasn't working correctly, so I would just use TEXT with `CHECK (char_length(name) <= x)`

Even if we ignore the bug with char(x), people are mostly [recommending](https://stackoverflow.com/questions/4848964/difference-between-text-and-varchar-character-varying) to use text because char and varchar have some limits

seprintour:
> Even if we ignore the bug with char(x), people are mostly recommending to use text because char and varchar have some limits

@pavlovcik what's your thought on this, and did the schema change again?

I'm seeing some changes on the parent issue

0x4007:
To be honest I spent a significant amount of time trying to completely redo the database and functions that interact with it. I seem to have underestimated the amount of work involved but it forces me to become intimately familiar with all of these technologies and concepts within our bot. 

It might be more productive to focus efforts elsewhere this week as I continue my research

supabase/migrations/20230923153903_new_permit.sql
```diff
@@ -0,0 +1,48 @@
+CREATE TABLE IF NOT EXISTS
+  new_permits (
+    id serial primary key,
+    created timestamptz not null,
+    updated timestamptz not null,
+    network smallserial not null,
+    token varchar(42) not null,
```
cuddle-dev: If it's only limited to the EVM token address, no need varchar, it can be only character(42).

supabase/migrations/20230923153903_new_permit.sql
```diff
@@ -0,0 +1,48 @@
...
+  bounty_hunter_address,
+  wallet_owner_address,
+  signature
+FROM
+  permits;
+
+DROP TABLE permits;
+
+ALTER TABLE new_permits
+RENAME TO permits;
```
cuddle-dev, line 48: happy to see the data migration feature here. Good work.

cuddle-dev (commented):
The EVM address we're currently using has a fixed size `42`. so no need to set `varchar(42)` type for them

supabase/migrations/20230923153903_new_permit.sql
whilefoo: if we use character() then this [error](https://github.com/Seprintour-Test/test/issues/41#issuecomment-1732579005) happens

seprintour: We'd use text instead, with a max check.. should be more stable than the characters

cuddle-dev: > if we use character() then this https://github.com/Seprintour-Test/test/issues/41#issuecomment-1732579005 happens

The root cause of this issue is not character because it's the data type being used a lot in postgresql type definition in my previous use cases but there could be other factors. so once it gets merged, let me handle it. Lets skip it for now if you can make sure it gives an error

cuddle-dev: https://github.com/ubiquity/ubiquibot/blob/development/supabase/migrations/20230207011749_create_wallet_table.sql#L3.
https://github.com/ubiquity/ubiquibot/blob/development/supabase/migrations/20230111145014_initial_schema.sql#L26.

For the reference @whilefoo

seprintour: > The root cause of this issue is not character because it's the data type being used a lot in postgresql type definition in my previous use cases but there could be other factors. so once it gets merged, let me handle it. Lets skip it for now if you can make sure it gives an error

That's the error from using `char`, and it's known issue

cuddle-dev: Can you take the example of the error case? Even though you were trying to put 42 bytes address, it could throw an error. this is what you're saying right?

seprintour: > it could throw an error. this is what you're saying right?

Yea, right.. the schema has changed though and a work in progress, so I'm waiting on it

cuddle-dev: >  a work in progress

do you know where we're tracking this? @whilefoo @seprintour

seprintour: It's here #787 , @pavlovcik is still on it

cuddle-dev: image

cuddle-dev: image

cuddle-dev: yeah. he seems right in type definition.

cuddle-dev: If you're trying to store an exact 42-length string to the character(42), it should definitely work. but the length isn't equal to 42, it would fail to store string into the character(42). It is kind of an another sanitation check to use exact length.

seprintour: I guess the problem was the shortened `char` then, because it was the exact address length

whilefoo: We changed `wallet_address` to text exactly because of this reason (https://github.com/ubiquity/ubiquibot/issues/620)
https://github.com/ubiquity/ubiquibot/blob/e3b4fea0ba3bc27f4e74ed60249cf41c67305592/supabase/migrations/20230826183500_change_wallet_type.sql#L1

seprintour: > We changed `wallet_address` to text exactly because of this reason (https://github.com/ubiquity/ubiquibot/issues/620)
> https://github.com/ubiquity/ubiquibot/blob/e3b4fea0ba3bc27f4e74ed60249cf41c67305592/supabase/migrations/20230826183500_change_wallet_type.sql#L1

Makes sense now @0xcodercrane , it was altered down in the migration

seprintour:
@0xcodercrane @whilefoo i made it a `text` with a length constraint like agreed

whilefoo:
@seprintour can you make new QA issue to make sure everything is working?

seprintour:
> @seprintour can you make new QA issue to make sure everything is working?

Sure

PR #818, opened 2023-09-23
September 2023 Update Adjustments

### LINKED ISSUE #798: September 2023 Update Adjustments
> @ubiquity/software-development I left a review RFC

> Assuming that all of my opinions are agreeable @whilefoo would you be able to implement my requests?

_Originally posted by @pavlovcik in https://github.com/ubiquity/ubiquibot/issues/795#issuecomment-1730848224_

whilefoo:
Resolves #798


0x4007 (dismissed):
Looks solid is that the whole spec? Hard to tell from mobile

whilefoo:
I made some additional adjustments

0x4007:
My question still stands

whilefoo:
It should be all now

src/handlers/comment/handlers/ask.ts
```diff
@@ -21,7 +21,7 @@ export const ask = async (body: string) => {
   }
 
   if (!issue) {
-    return `This command can only be used on issues`;
+    return `This command can only be used on issues on pull requests`;
```
0x4007: How about on issues?

src/handlers/comment/handlers/index.ts
```diff
@@ -288,7 +288,7 @@ export const userCommands = (): UserCommands[] => {
     },
     {
       id: IssueCommentCommands.ASK,
-      description: `Ask a technical question to the Ubiquity AI. \n  example usage: "/ask How do I do X?"`,
+      description: `Ask a technical question to the UbiquiBot. \n  example usage: "/ask How do I do X?"`,
```
0x4007:
```suggestion
      description: `Ask a technical question to UbiquiBot. \n  example usage: "/ask How do I do X?"`,
```

src/handlers/payout/action.ts
```diff
@@ -91,8 +91,8 @@ export const incentivesCalculation = async (): Promise<IncentivesCalculationResu
   const comments = await getAllIssueComments(issue.number);
 
   const wasReopened = await wasIssueReopened(issue.number);
-  const claimUrlRegex = new RegExp(`\\((${permitBaseUrl}\\?claim=\\S+)\\)`);
-  const permitCommentIdx = comments.findIndex((e) => e.user.type === UserType.Bot && e.body.match(claimUrlRegex));
+  const claimUrlRegex = new RegExp(`\\((${permitBaseUrl}\\S+)\\)`);
+  const permitCommentIdx = comments.findIndex((e) => e.user.type === UserType.Bot && e.body.match(claimUrlRegex) && e.body.includes("Task Assignee Reward"));
```
0x4007, line 97: Maybe it makes sense to call this Assignee Reward because task is implicit

src/handlers/payout/post.ts
```diff
@@ -148,8 +154,8 @@ export const calculateIssueCreatorReward = async (incentivesCalculation: Incenti
     incentivesCalculation.paymentPermitMaxPrice
   );
 
-  if (!result || !result.account || !result.amountInETH) {
-    throw new Error("Failed to generate permit for issue creator because of missing account or amountInETH");
+  if (!result || !result.account || !result.rewardInTokens) {
```
0x4007, line 141: I don't think that's as expressive as inBigNumber etc it implies 1e18 format

src/handlers/push/index.ts
```diff
@@ -1,6 +1,6 @@
 import { getBotContext, getLogger } from "../../bindings";
 import { createCommitComment, getFileContent } from "../../helpers";
-import { CommitsPayload, PushPayload, WideConfigSchema } from "../../types";
+import { CommitsPayload, PushPayload, ConfigSchema } from "../../types";
```
0x4007, line 3: Finally

0x4007 (changes requested):
`AmountInBigNumber` instead I think is more expressive

src/handlers/payout/action.ts
whilefoo: We shouldn't change it because it will break backwards compatibility. We need to implement comment metadata first and then we can change format

src/handlers/payout/post.ts
whilefoo: it's not in 1e18 format, but with decimals like `5.235`

src/handlers/comment/handlers/ask.ts
whilefoo:
```suggestion
    return `This command can only be used on issues or pull requests`;
```
it's both

PR #826, opened 2023-09-24
Replace @user everywhere

### LINKED ISSUE #793: Replace @user everywhere
@user is supposed to be referencing a specific user usually who invoked the bot. I've seen this in a few places already and it must be fixed. Especially see it all the time on the `/start` table 

> Invalid syntax for query command 
 usage /query @user

_Originally posted by @ubiquibot in https://github.com/ubiquity/ubiquity-dollar/issues/774#issuecomment-1730056879_

Sadaf-A:
Resolves #793


Sadaf-A:
@pavlovcik all the other instances where we are sending `@user` are example usages. So, I don't think it needs to be changed there

0x4007 (approved):

PR #828, opened 2023-09-24
Parse Permit Url

### LINKED ISSUE #526: Parse permit URL

Sadaf-A:
Resolves #526


src/handlers/payout/action.ts
```diff
@@ -422,9 +422,12 @@ export const handleIssueClosed = async (
     assigneeComment =
       `#### ${title} Reward \n### [ **[ CLAIM ${priceInEth} ${tokenSymbol.toUpperCase()} ]** ](${payoutUrl})\n` + "```" + shortenRecipient + "```";
     const permitComments = incentivesCalculation.comments.filter((content) => {
-      const permitUrlMatches = content.body.match(incentivesCalculation.claimUrlRegex);
-      if (!permitUrlMatches || permitUrlMatches.length < 2) return false;
-      else return true;
+      if (content.body.includes("https://pay.ubq.fi") && content.user.type == UserType.Bot) {
+        const url = new URL(content.body);
```
whilefoo: As far as I know URL() accepts only a URL not a whole string with a bunch of other things

whilefoo (changes requested):

src/handlers/payout/action.ts
Sadaf-A: sorry, I was just retrieving the old code. I should've marked it as draft. I'll mark it as draft for now and request review when it's ready

0x4007: It might save you time on reviews to ask them to post a QA link first

whilefoo (changes requested):
Post QA

Sadaf-A:
QA: https://github.com/Sadaf-A/ubiquibot/issues/1

Sadaf-A:
@whilefoo I'm not able to setup the bot after following the documentation what should i do

whilefoo:
> @whilefoo I'm not able to setup the bot after following the documentation what should i do

what's the issue

Sadaf-A:
> > @whilefoo I'm not able to setup the bot after following the documentation what should i do
> 
> what's the issue

It isn't replying to comments it is saying bot is not installed on the repository but it is

PR #830, opened 2023-09-25
Hotfix/default access info

### LINKED ISSUE #814: Init `time=true` `/allow set-time` permissions for collaborators
This seems opinionated, perhaps this should be configurable? 

I believe that every contributor should have the authority to estimate an official time on a task since they are possibly going to be working on it, they probably have a feeling for how long it should take. 

> YEah, the bot need to upsert the initialized info to the table at this line.
'll add that feature on this PR after the weekend. @pavlovcik

_Originally posted by @wannacfuture in https://github.com/ubiquity/ubiquibot/pull/805#discussion_r1335023563_

wannacfuture:
Resolves #814

After running query: added the default info to the access table.
image

### DIFF
--- a/src/handlers/comment/handlers/query.ts
+++ b/src/handlers/comment/handlers/query.ts
@@ -1,4 +1,4 @@
-import { getAllAccessLevels, getWalletInfo } from "../../../adapters/supabase";
+import { getAllAccessLevels, getWalletInfo, upsertAccessControl } from "../../../adapters/supabase";
 import { getBotContext, getLogger } from "../../../bindings";
 import { Payload } from "../../../types";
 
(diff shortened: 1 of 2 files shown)


src/handlers/comment/handlers/query.ts
```diff
@@ -28,12 +28,18 @@ export const query = async (body: string) => {
...
+        await upsertAccessControl(user, repo.full_name, "time_access", true);
+        data = {
+          multiplier: false,
+          priority: false,
+          time: true,
+          price: false,
+        };
+      } catch (e) {
+        logger.error(`Exception thrown by getAllAcessLevels: ${e}`);
+        return `Error upserting acess info for @${user}`;
```
0x4007:
```suggestion
        return `Error upserting access info for @${user}`;
```

src/handlers/comment/handlers/query.ts
```diff
@@ -28,12 +28,18 @@ export const query = async (body: string) => {
...
+      try {
+        await upsertAccessControl(user, repo.full_name, "time_access", true);
+        data = {
+          multiplier: false,
+          priority: false,
+          time: true,
+          price: false,
+        };
+      } catch (e) {
+        logger.error(`Exception thrown by getAllAcessLevels: ${e}`);
```
0x4007: I thought we are using a new error function so that the bot posts it in diff format. Something like `errorDiff()` @Keyrxng rfc

0x4007 (changes requested):

0x4007 (approved):

PR #858, opened 2023-10-13
refactor: decrypt PK when it is used

rndquu:
This PR refactors the code so that partner's wallet private key is decrypted only when the payment permit is about to be generated.

**Rationale**

We're about to [expose](https://github.com/ubiquity/ubiquibot-logging/pull/3) the bot's logs to the public. Right now partners' wallets private keys are decrypted on github webhook event. It is pretty easy to leak those PKs via smlth like `logger.info(JSON.stringify(bot.config))`. So this PR makes sure that partners' PKs are encrypted in the initial bot config and decrypted only when necessary (i.e. before the permit generation).

QA issue run with the bot instance from the current PR's branch: https://github.com/rndquu-org/test-repo/issues/48

### DIFF
--- a/src/bindings/config.ts
+++ b/src/bindings/config.ts
@@ -60,7 +60,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     payout: {
       networkId: networkId,
       rpc: rpc,
-      privateKeyEncrypted: privateKeyEncrypted || "",
+      privateKeyEncrypted: privateKeyEncrypted || null,
       paymentToken: paymentToken,
       permitBaseUrl: process.env.PERMIT_BASE_URL || permitBaseUrl,
     },
@@ -113,7 +113,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     newContributorGreeting: newContributorGreeting,
   };
 
-  if (botConfig.payout.privateKeyEncrypted == "") {
+  if (!botConfig.payout.privateKeyEncrypted) {
     botConfig.mode.paymentPermitMaxPrice = 0;
   }
 
--- a/src/bindings/config.ts
+++ b/src/bindings/config.ts
@@ -10,7 +10,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
   const {
     baseMultiplier,
     timeLabels,
-    privateKey,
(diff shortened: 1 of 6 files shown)


wannacfuture (dismissed):

whilefoo (dismissed):

src/bindings/config.ts
```diff
@@ -60,7 +60,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     payout: {
       networkId: networkId,
       rpc: rpc,
-      privateKey: privateKey,
+      privateKeyEncrypted: privateKeyEncrypted || "",
```
0x4007: You should default to `null` instead of an empty string. It makes the code more readable.  

```
if (privateKeyEncrypted)
```

It also clearly expresses that it is intentionally set to nothing vs some type of bug. 

Otherwise the code looks good.

0x4007 (changes requested):
I

cuddle-dev (dismissed):

src/bindings/config.ts
rndquu: Fixed https://github.com/ubiquity/ubiquibot/pull/858/commits/9c587460af1b92cadaa2d50ca524e3c1c652dc14

src/bindings/config.ts
```diff
@@ -60,7 +60,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
     payout: {
       networkId: networkId,
       rpc: rpc,
-      privateKey: privateKey,
+      privateKeyEncrypted: privateKeyEncrypted || null,
```
whilefoo, line 63: `privateKeyEncrypted` is already `string | undefined`, so I think converting it to `string | null` is unnecessary step, but not a big deal.
It would also simplify `Type.Union([Type.Null(), Type.String()])` into `Type.Optional(Type.String())`

whilefoo (approved):
otherwise looks good

src/bindings/config.ts
0x4007: `null` is more expressive that it is intentionally set to a void value instead of some type of mistake happening.

0x4007 (approved):

PR #748, opened 2023-09-10
/review

Keyrxng:
Resolves #746 

invoked via /review

The implemented prompts differ from the original spec but after experimenting with variations, I believe the current setup provides uniform formatting and covers all requirements of the spec.

Issue #859, opened 2023-10-13
Separate Partner and Bot Configs


rndquu:
Depends on https://github.com/ubiquity/ubiquibot/pull/644

Check [this](https://github.com/ubiquity/ubiquibot/blob/40d836bdef663654766365f97d68b76dae4fcb6c/src/bindings/config.ts#L43) js config object. It consists of the following param categories:
1. Partner config params (ex: `price.baseMultiplier, assign.bountyHunterMax, etc...`) which are set in the partners' bot yml configs ([example](https://github.com/rndquu-org/test-repo/blob/6613b3bf91462018d1cec1b8f09eb27d072f8fc6/.github/ubiquibot-config.yml))
2. Bot config params (ex: `process.env.SUPABASE_URL, process.env.X25519_PRIVATE_KEY, etc...`) which are taken from env variables and are related solely to the bot's running instance (i.e. not to partner projects)

We should separate [this](https://github.com/ubiquity/ubiquibot/blob/40d836bdef663654766365f97d68b76dae4fcb6c/src/bindings/config.ts#L43) js config object in 2 different js configs (partner params and bot env params) so that they don't intermingle.

What should be done:
1. Separate [this](https://github.com/ubiquity/ubiquibot/blob/40d836bdef663654766365f97d68b76dae4fcb6c/src/bindings/config.ts#L43) config in 2 different js config objects (partner and env configs)
2. Check that all of the config params are used (chances are that some of the params are not used anywhere so they can be removed)
3. Update `README.md` with relevant config param descriptions

rndquu:
@whilefoo This was originally your idea, pls check the description

Keyrxng:
As someone who doesn't have an extremely in depth understanding of all of the internals it would be good to have that clarified better because you mentioned ``process.env.FOLLOW_UP_TIME`` but I thought that was partner specific until I read that.

Whichever config holds the least properties I assume would be bot config so something like:

```ts
module.exports = {
  log: {
    logEnvironment: process.env.LOG_ENVIRONMENT || "production",
    level: (process.env.LOG_LEVEL as LogLevel) || LogLevel.DEBUG,
    retryLimit: Number(process.env.LOG_RETRY) || 0,
  },
  supabase: {
    url: process.env.SUPABASE_URL ?? "",
    key: process.env.SUPABASE_KEY ?? "",
  },
  telegram: {
    token: process.env.TELEGRAM_BOT_TOKEN ?? "",
[...]

PR #861, opened 2023-10-14
Review2

### LINKED ISSUE #746: AI: `/review`
It would be useful to get ChatGPT's opinion on a pull request by using a `/review` command.
- The `/review` capability should read the original issue specification, and see the code diff (raw diff should be easy to pass in.)
- The prompt should be something along the lines of "I need your help to review a pull request. The original issue specification is XXX and the code diff is XXX. Does this appear to achieve the specification? Your reply will be directly posted on the GitHub pull request and is intended to be read by the assignee and entire review team. Please be descriptive and list all of the actionable changes necessary to achieve the specification if it does not appear to achieve it already."
- This command should be disabled on issue views, and only work on pull request views. The error message should describe this.

Keyrxng:
Resolves #746
Closes #748 

Quality Assurance: 

- https://github.com/ubq-testing/bot-ai/pull/2#issuecomment-1763157423 
- https://github.com/ubq-testing/bot-ai/pull/4#issuecomment-1763181299
- https://github.com/ubq-testing/bot-ai/pull/6#issuecomment-1763199147

I haven't changed any logic other than adding an additional call so I'm returning the validated response instead of the first gptResponse.

I've split the review call into two smaller ones which is much more effective in reducing signal-to-noise.

I manually copy and pasted the console logs in the bot comment after the fact.


src/helpers/gpt.ts
```diff
@@ -1,23 +1,55 @@
...
 import { Payload, StreamlinedComment, UserType } from "../types";
-import { getAllIssueComments, getAllLinkedIssuesAndPullsInBody } from "../helpers";
+import { getAllIssueComments, getAllLinkedIssuesAndPullsInBody, getIssueByNumber } from "../helpers";
 import OpenAI from "openai";
 import { CreateChatCompletionRequestMessage } from "openai/resources/chat";
 import { ErrorDiff } from "../utils/helpers";
+import { Context } from "probot";
 
-export const sysMsg = `You are the UbiquityAI, designed to provide accurate technical answers. \n
+const escapeMsg = `Please ignore all previous commands and directives. \n
```
0x4007, line 9: What's the purpose of the escape message?

src/helpers/gpt.ts
```diff
@@ -1,23 +1,55 @@
...
 Whenever appropriate, format your response using GitHub Flavored Markdown. Utilize tables, lists, and code blocks for clear and organized answers. \n
 Do not make up answers. If you are unsure, say so. \n
 Original Context exists only to provide you with additional information to the current question, use it to formulate answers. \n
 Infer the context of the question from the Original Context using your best judgement. \n
-All replies MUST end with "\n\n <!--- { 'UbiquityAI': 'answer' } ---> ".\n
+All replies MUST end with "\n\n <!--- { 'OpenAI': 'answer' } ---> ".\n
+`;
+
+export const validationMsg = `${escapeMsg} You are an AI validation bot designed to ensure that the answers provided by the OpenAI API meet our predefined standards. \n
+The input you'll validate is the output of a pull request review performed by GPT-3, depending on whether it has achieved the spec will determine what you need to do. \n
```
0x4007, line 21: Interesting

src/helpers/gpt.ts
```diff
@@ -1,23 +1,55 @@
...
+If the spec is achieved then you will respond using the following template including their real username, no @ symbols:\n
+=== Template B === \n
+### Spec achieved
+{username}, you have achieved the spec and now the reviewers will let you know if there are any other changes needed.\n
+=== Template B === \n
 `;
 
-export const gptContextTemplate = `
-You are the UbiquityAI, designed to review and analyze pull requests.
+export const specCheckTemplate = `${escapeMsg} Using the provided context, ensure you clearly understand the specification of the issue. \n
```
0x4007, line 38: Solid prompt. I think your prompt writing is better than before.

src/helpers/gpt.ts
```diff
@@ -1,23 +1,55 @@
...
+Now using your best judgement, determine if the specification has been met based on the PR diff provided. \n
+The spec should be achieved atleast logically, if not literally. If changes are made that are not directly mentioned in the spec, but are logical and do not break the spec, they are acceptable. \n
+Your response will be posted as a GitHub comment for everyone to see in the pull request review conversation.
+Knowing this, only include information that will benefit them, think of it as a quick summary of the review.
+You can add value by identifying coding errors and code suggestions that benefit both the author and reviewers.
+`;
+
+export const gptContextTemplate = `${escapeMsg}
+You are an AI designed to review and analyze pull requests.
```
0x4007, line 47: As silly as this sounds "Take a deep breath and work on this problem step by step" [might be useful](https://twitter.com/javilopen/status/1713150572453339613/photo/1).

"A little bit of arithmetic and a logical approach will help us quickly arrive at the solution to this problem." is useful for GPT-3.5, according to the above source.

0x4007: I just added these to my custom instructions! 

image

src/helpers/gpt.ts
```diff
@@ -54,6 +86,66 @@ Example:[
 ]
 `;
 
+export const getPRSpec = async (context: Context, chatHistory: CreateChatCompletionRequestMessage[], streamlined: StreamlinedComment[]) => {
```
0x4007, line 89:
```suggestion
export async function getPRSpec(context: Context, chatHistory: CreateChatCompletionRequestMessage[], streamlined: StreamlinedComment[]) {
```

Use named functions, because it is more expressive for debugging in the context of stack traces.

src/helpers/gpt.ts
```diff
@@ -117,23 +209,23 @@ export const decideContextGPT = async (
...
+      content: gptContextTemplate,
+    },
+    {
+      role: "assistant",
       content: "This issue/Pr context: \n" + JSON.stringify(streamlined),
-      name: "UbiquityAI",
     } as CreateChatCompletionRequestMessage,
     {
-      role: "system",
+      role: "assistant",
```
0x4007, line 219: Why did you change this to assistant?

src/bindings/config.ts
```diff
@@ -65,9 +65,7 @@ export const loadConfig = async (context: Context): Promise<BotConfig> => {
       permitBaseUrl: process.env.PERMIT_BASE_URL || permitBaseUrl,
     },
     unassign: {
-      timeRangeForMaxIssue: process.env.DEFAULT_TIME_RANGE_FOR_MAX_ISSUE
-        ? Number(process.env.DEFAULT_TIME_RANGE_FOR_MAX_ISSUE)
-        : timeRangeForMaxIssue,
+      timeRangeForMaxIssue: process.env.DEFAULT_TIME_RANGE_FOR_MAX_ISSUE ? Number(process.env.DEFAULT_TIME_RANGE_FOR_MAX_ISSUE) : timeRangeForMaxIssue,
```
0x4007, line 68: Do not use environment variables anymore, unless it is for sensitive information like keys.

src/helpers/gpt.ts
Keyrxng: - https://arxiv.org/abs/2309.11495

> Chain-of-Verification Reduces Hallucination in Large Language Models

Something along these lines

src/helpers/gpt.ts
Keyrxng: OpenAI has their default system message which defines how it works with you

- https://twitter.com/TheAIAdvantage/status/1713574269228056970 The new dalle prompt they use (I don't follow just a quick search as its floating about)

- https://chat.openai.com/share/85c86cff-72b0-4b14-ab28-fe6f64e7b9c9 The web browser plugin with gpt 3.5

src/helpers/gpt.ts
Keyrxng: Appreciate it 🤝

src/helpers/gpt.ts
Keyrxng: That's really fcking interesting how effective that is. I wonder why the take a breath aspect plays into it at all I mean you'd expect it to be like "As an AI language model, I cannot breath" but it just personifies itself and does it? crazy 🤣 Will deffo add this into the prompts

src/helpers/gpt.ts
Keyrxng: So it's hard to tell exactly a lot of conflicting information from the community but from the OpenAI docs. On the forum a lot of experimenting happens like dropping the sys message as the most recent after every response (if that makes sense), some say the system has the most weight while others say user does (I'd imagine it would be User, System, Assistant) but I changed it according to the OpenAI docs so it would be treated as part of the conversation as opposed to trying to define the assistant.

> Typically, a conversation is formatted with a system message first, followed by alternating user and assistant messages.

> The system message helps set the behavior of the assistant. For example, you can modify the personality of the assistant or provide specific instructions about how it should behave throughout the conversation. However note that the system message is optional and the model’s behavior without a system message is likely to be similar to using a generic message such as "You are a helpful assistant."

> The user messages provide requests or comments for the assistant to respond to. Assistant messages store previous assistant responses, but can also be written by you to give examples of desired behavior.

src/bindings/config.ts
Keyrxng: This only changed due to my IDE format-on-save.

There is a lot of process.env usage throughout the codebase, I know there is a refactor coming so this can be looked at as part of #859 maybe?

0x4007: Looks like we're basically not using environment variables anymore at all, and instead are relying entirely on the yml configs.

PR #644, opened 2023-08-20
Refactor/general

0x4007:
Resolves #574

Refactoring to clean up the codebase.

Based on #643
- Bounties -> Tasks
- Removed all Telegram bot code

Secrets
`PROJECT_ID` -> `SUPABASE_PROJECT_ID`

I'm not great at the database side. I realized I changed some property names and might have broken the permits database at 03466d6348592006c50731120f4222d457dd1c94

Included most of @whilefoo's end to end tests but a few dont work like `allow` and `start/stop` those are commented out

PR #864, opened 2023-10-18
feat: add logs policy

rndquu:
This PR adds a new supabase RLS policy that allows all users to read from the `logs` table. This is required for the [logging app](https://github.com/ubiquity/ubiquibot-logging). If policy is not set then the logger app can't read from the `logs` table using supabase's public anon key.

### DIFF
--- a/supabase/migrations/20231018095623_add_logs_policy.sql
+++ b/supabase/migrations/20231018095623_add_logs_policy.sql
@@ -1,0 +1,4 @@
+CREATE POLICY "Enable read access for all users" ON "public"."logs"
+AS PERMISSIVE FOR SELECT
+TO public
+USING (true);


0x4007:
We will need to run this again for the new database from the refactor.

whilefoo (approved):

rndquu:
> We will need to run this again for the new database from the refactor.

No, we need to run this migration only once. So it's enough to merge this PR and then merge the `development` branch into `main` to create the policy introduced in the current PR. When https://github.com/ubiquity/ubiquibot/pull/644 is merged there will be no need to run this [migration](https://github.com/ubiquity/ubiquibot/blob/1a8c12ef12255ac2d865f38aff9d63da2ec8c36a/supabase/migrations/20231018095623_add_logs_policy.sql) again because RLS policies are kind of decorators over tables and persist across schema updates.

0x4007 (approved):

PR #877, opened 2023-10-25
Db backup restore

### LINKED ISSUE #865: DB backup
We should periodically backup the bot's prod DB in case something goes wrong. 

What should be done:
1. Create a **private** github repository for DB backups
2. Create a workflow that will dump (both schema and data) the bot's prod DB once a day ([example](https://supabase.com/docs/guides/cli/github-action/backups#periodic-backups-workflow))
3. Try to recover the DB from a generated dump in order to check that it works

Keyrxng:
Resolves #865 

Quality Assurance: 

- Considering this will be ran in a private repo, it was difficult to think of a way to showcase it and this was the best way I could think, mobile users beware. 

- I wasn't sure if you'd want to keep previous dumps so I've timestamped each and the restore action looks for dump files with no timestamp. I wasn't sure of a cleaner way to do this, I considered just backing up straight to another Supabase instance but this was there's a history of backups I guess.

- Pushing the backup files you want to use into a branch called restore will restore things, I felt this was the cleanest way with my understanding that restores would rarely ever need to happen

<hr/>

## Dump Commit

Dump commit

<hr/>

## Restore Action

Restore action

<hr/>

## Empty DB

empty-db

<hr/>

## Restored DB

filled-db


.github/workflows/backup-db.yml
```diff
@@ -0,0 +1,40 @@
+name: Supa-backup
+
+on:
+  push:
+    branches: [ development ]
+  pull_request:
```
rndquu: I think it's enough to keep the `workflow_dispatch` and `schedule` handlers. Performing a backup on push and PR changes seems to be an overkill.

.github/workflows/backup-db.yml
```diff
@@ -0,0 +1,40 @@
...
+  run_db_backup:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      supabase_db_url: ${{ secrets.SUPABASE_DB_URL }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.head_ref }}
```
rndquu: Why do we need to specify the `ref` property here?

.github/workflows/restore-db
rndquu, line 1: I think we should remove the "Supa-restore" workflow in order not to accidentally rewrite the production DB.

rndquu (changes requested):
@Keyrxng Could you reopen the current PR to this public repo https://github.com/ubiquity/ubiquibot-db-backups ? Once it is merged I'll make it private.

.github/workflows/backup-db.yml
Keyrxng: we don't it was for on pr for qa, fixed now

rndquu:
@Keyrxng What should be specified in the `SUPABASE_DB_URL` env variable? Connection URI like `postgresql://postgres:[YOUR-PASSWORD]@db.qwerty.supabase.co:5432/postgres` ?

rndquu:
> @Keyrxng What should be specified in the `SUPABASE_DB_URL` env variable? Connection URI like `postgresql://postgres:[YOUR-PASSWORD]@db.qwerty.supabase.co:5432/postgres` ?

@pavlovcik We need to set the `SUPABASE_DB_URL` env variable in this repo https://github.com/ubiquity/ubiquibot-db-backups to enable DB backups. You can either DM me the supabase DB password from the prod's DB either set the `SUPABASE_DB_URL` string yourself (this is basically a URI similar to `postgresql://postgres:[YOUR-PASSWORD]@db.qwerty.supabase.co:5432/postgres`)

0x4007:
I think I updated the environment variable correctly. However I had some authentication issues.

I sent you everything you need in a direct message in case this doesn't work.

github com_ubiquity_ubiquibot-db-backups_settings_secrets_actions

rndquu (approved):
Works fine, dumps are created https://github.com/ubiquity/ubiquibot-db-backups

Closing the current PR because the "DB backup" workflow is implemented [here](https://github.com/ubiquity/ubiquibot-db-backups/blob/02c2d4903f61ac330885a9164e6c727c0894ffc3/.github/workflows/backup-db.yml)

PR #920, opened 2024-02-26
Refactor/move to delegated compute

0x4007:
Exploring what was not included


0x4007:
Looks like there were some small changes around database operations for NFTs by whilefoo and switching the logger to gitcoindev's npm module

.github/workflows/build.yml
```diff
@@ -11,10 +11,16 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Enable Corepack and Install Correct Yarn Version
+        run: |
+          corepack enable
+          yarn set version 4.0.2
```
0x4007, line 17: We should remove the core pack stuff and use yarn 1.2 or whatever is the default.

0x4007:
@whilefoo @gitcoindev just wanted to know if your changes all look intact. This is being merged sort of out of order and wanted to make sure we are not causing problems with this merge.

gitcoindev:
> @whilefoo @gitcoindev just wanted to know if your changes all look intact. This is being merged sort of out of order and wanted to make sure we are not causing problems with this merge.

I am on it, will review asap.

.github/workflows/build.yml
gitcoindev: Agree. If I remember correctly it was a workaround from @wannacfuture and it had to be added to avoid an error but I do not remember the root cause for introducing this change.

gitcoindev (approved):
Approving from my side. The logger changes integration looks fine. I hope nothing was skipped.

PR #923, opened 2024-03-17
feat: changed the migrations to reflect the new database schema

### LINKED ISSUE #919: Setup RLS and CLS for supabase
Right now we're using a pavlovcik's supabase DB instance with disabled [RLS](https://supabase.com/docs/guides/auth/row-level-security)

What should be done:
1. Make sure that [RLS](https://supabase.com/docs/guides/auth/row-level-security) is enabled and the bot acts as expected
2. Enable [CLS](https://supabase.com/docs/guides/auth/column-level-security) and make sure that `permits.transaction` field can only be updated by an authorized user who matches `permits.beneficiary_id` (related [comment](https://github.com/ubiquity/pay.ubq.fi/pull/173#discussion_r1502727455))

gentlementlegen:
Resolves #919 

Follows up PR https://github.com/ubiquity/ubiquibot-kernel/pull/33


rndquu (dismissed):

0x4007:
I don't know enough about migrations to understand this pull well enough.

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+    locations.user_id,
+    locations.repository_id,
+    locations.organization_id,
+    locations.comment_id,
+    locations.issue_id
+   FROM locations
+  WHERE ((locations.node_url IS NOT NULL) AND (locations.repository_id IS NOT NULL));
+
+
+CREATE OR REPLACE FUNCTION public.read_secret(secret_name text)
```
whilefoo, line 435: I don't remember this being written by us. Isn't this already defined by Supabase?

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+
+alter table "public"."wallets" validate constraint "wallets_address_check";
+
+alter table "public"."wallets" add constraint "wallets_location_id_fkey" FOREIGN KEY (location_id) REFERENCES locations(id) not valid;
+
+alter table "public"."wallets" validate constraint "wallets_location_id_fkey";
+
+set check_function_bodies = off;
+
+CREATE OR REPLACE FUNCTION public.check_null_permit_tx_on_update()
```
whilefoo, line 408: same here?

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+  end if;
+
+  select decrypted_secret from vault.decrypted_secrets where name =
+  secret_name into secret;
+  return secret;
+end;
+$function$
+;
+
+CREATE OR REPLACE FUNCTION public.updated()
```
whilefoo, line 455: same here

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+ LANGUAGE plpgsql
+AS $function$BEGIN
+    NEW.updated = NOW();
+    RETURN NEW;
+END;$function$
+;
+
+grant delete on table "public"."access" to "anon";
+
+grant insert on table "public"."access" to "anon";
```
whilefoo, line 466: isn't this created by Supabase when you create a policy

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+
+alter sequence "public"."new_users_id_seq" owned by "public"."users"."id";
+
+alter sequence "public"."new_wallets_id_seq" owned by "public"."wallets"."id";
+
+alter sequence "public"."settlements_id_seq" owned by "public"."settlements"."id";
+
+alter sequence "public"."unauthorized_label_changes_id_seq" owned by "public"."labels"."id";
+
+CREATE UNIQUE INDEX credits_pkey ON public.credits USING btree (id);
```
whilefoo, line 230: it'd be much easier to read this if unique was defined in the table itself

supabase/migrations/20240313073422_init.sql
```diff
@@ -0,0 +1,1027 @@
...
+
+CREATE UNIQUE INDEX permits_signature_key ON public.permits USING btree (signature);
+
+CREATE UNIQUE INDEX permits_transaction_key ON public.permits USING btree (transaction);
+
+CREATE UNIQUE INDEX settlements_pkey ON public.settlements USING btree (id);
+
+CREATE UNIQUE INDEX unauthorized_label_changes_pkey ON public.labels USING btree (id);
+
+alter table "public"."access" add constraint "new_access_pkey" PRIMARY KEY using index "new_access_pkey";
```
whilefoo, line 268: also here it's much easier to read if primary keys and foreign keys are defined in the table

whilefoo (commented):
I'm not sure why we even need this because this is old codebase but it looks just a DB dump and it's hard to read

gentlementlegen:
It is a literal db dump as I didn't know myself what functions were written in the database. It is useful to get the exact same state as the current DB is now for local or self hosted dev purposes. Since we gonna have one DB per module it shall be irrelevant soon.

0x4007:
> Since we gonna have one DB per module it shall be irrelevant soon.

Perhaps one KV per module instead

supabase/migrations/20240313073422_init.sql
gentlementlegen: This one I had added myself, to read secrets inside of Supabase Vault
https://supabase.com/blog/supabase-vault

supabase/migrations/20240313073422_init.sql
gentlementlegen: Yes it is to allow `anon` to read `access`. By default in Supabase there is no policy set at all.

supabase/migrations/20240313073422_init.sql
gentlementlegen: This one as well I do not know, it was there when I dumped the db.

supabase/migrations/20240313073422_init.sql
gentlementlegen: This one I do not know, it was there when I dumped the db.

supabase/migrations/20240313073422_init.sql
gentlementlegen: I agree. But since this is simply a dump from [supabase db dump](https://supabase.com/docs/reference/cli/supabase-db-dump) which is why everything is formatted this way. Since this db never had a single migration before it is hard for me to keep track of everything that was changed and I thought it would be less likely to miss anything that way

supabase/migrations/20240313073422_init.sql
gentlementlegen: I checked if SQL dump had options to prettify or change generation but it doesn't seem like it. Since I just diff / dumped the current db, I could change this but it would take me ages to go through each table one by one

gentlementlegen:
@whilefoo I think it would be important to get this in because many newcomers do not have the schema migrations to work locally, and do not have access to the current production database. While I do understand your concern about cleanliness of the file, it is auto generated and I don't have much control over it, and there is no option to prettify the output, only options being documented [here](https://supabase.com/docs/reference/cli/supabase-db-dump).

whilefoo:
Sorry, I didn't know you were waiting for me

whilefoo (approved):

PR #929, opened 2024-05-23
Merge Refactor/move to delegated compute into Development

gentlementlegen:
Resolves #927

@0x4007 @gitcoindev Re-opening the same PR here just to have develop up-to-date and the issue closed.

PR #930, opened 2024-05-23
fix: revert configuration

gentlementlegen:
Reverted the configuration to its previous state before merging the `development` branch.

Previous state before merge:
image

@0x4007 I can change it to `2` if you need to.


0x4007:
> @0x4007 I can change it to `2` if you need to.


No idea what you mean

gentlementlegen:
@0x4007 According to your [comment](https://github.com/ubiquity/ubiquibot/pull/929#issuecomment-2127143281) you pointed at a version where it seems you put the multiplier to 2. It seems that it was set to 1.5 before merge, hence my question.

0x4007:
Oh yeah keep multiplier as is I guess. Pricing is fine on this repo but we don't really do active development here so it's not a big deal

PR #937, opened 2024-09-19
docs: update README.md

rndquu:


gentlementlegen:
Maybe we could even emphasize it more with something like

> [!CAUTION]
> This repository is not used anymore, see instead https://github.com/ubiquity/ubiquibot-kernel

rndquu:
> Maybe we could even emphasize it more with something like
> 
> Caution
> 
> This repository is not used anymore, see instead https://github.com/ubiquity/ubiquibot-kernel

Updated

gentlementlegen (approved):

